PR #6, opened 2019-07-04
test travis integration

nimrod-becker:
WIP


Makefile
```diff
@@ -3,6 +3,8 @@ BIN = build/_output/bin
 REPO = github.com/noobaa/noobaa-operator
 IMAGE = noobaa/noobaa-operator:master
 
+GO_FLAGS ?= GOOS=linux GOARCH=amd64 CGO_ENABLED=0 GO111MODULE=on
```
guymguym: I see the current CI failure is because we are building "all" which also includes "cli" and for the cli `go build` call we didn't set the `GO111MODULE=on` env.

If you just want to see it pass this step then add just `GO111MODULE=on` directly to the cli go build call because we don't want to restrict building the cli for linux only.

What I want to do next in the makefile is to build the cli to all supported platforms and make a process to add these to a github release.

nimrod-becker: Changes for now

PR #13, opened 2019-07-16
adds noobaa-mixins as a dependency

umangachapagain:
adding `jsonnet/` to pull noobaa-mixins for build and integration into noobaa-operator

`make mixins` will fetch the updated resources from noobaa-mixins and create a `noobaa-prometheus-rules.yaml` which can be used for deploying alerts. 

`kubectl create -f noobaa-prometheus-rules.yaml` or,
`oc create -f noobaa-prometheus-rules.yaml`

Signed-off-by: Umanga Chapagain <chapagainumanga@gmail.com>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ VENV = $(OUTPUT)/venv
 
 # Default tasks:
 
-all: build
+all: mixins build
 	@echo "@@@ All Done."
 .PHONY: all
 
@@ -93,5 +93,32 @@ install-sdk:
 	curl https://github.com/operator-framework/operator-sdk/releases/download/${SDK_VERSION}/operator-sdk-${SDK_VERSION}-x86_64-linux-gnu -sLo ${GOPATH}/bin/operator-sdk
 	chmod +x ${GOPATH}/bin/operator-sdk
 .PHONY: install-sdk
+
+mixins: dependencies
+	cd jsonnet && \
+	jb update && \
+	./build-jsonnet.sh make.jsonnet
+.PHONY: mixins
+
+dependencies: jb jsonnet gojsontoyaml jq
+.PHONY: dependencies
+
+jb:
+	go get -u github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb
+.PHONY: jb
+
+jsonnet:
+	go get -u github.com/google/go-jsonnet/cmd/jsonnet
+.PHONY: jsonnet
+
+gojsontoyaml:
+	go get -u github.com/brancz/gojsontoyaml
+.PHONY: gojsontoyaml
+
+jq:
+	wget https://github.com/stedolan/jq/releases/download/jq-1.6/jq-linux64 -O jq && \
+	chmod +x jq && \
+	mv jq $(GOPATH)/bin
+.PHONY: jq
 
 #TODO scorecard 
--- a/jsonnet/build-jsonnet.sh
+++ b/jsonnet/build-jsonnet.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+set -e
+set -x
+# only exit with zero if all commands of the pipeline exit successfully
+set -o pipefail
+
+prefix="manifests"
+rm -rf $prefix
+mkdir $prefix
+
+rm -rf tmp
+mkdir tmp
+
+jsonnet -J vendor main.jsonnet > tmp/main.json
+
+mapfile -t files < <(jq -r 'keys[]' tmp/main.json)
+
+for file in "${files[@]}"
+do
+	dir=$(dirname "${file}")
+	path="${prefix}/${dir}"
+	mkdir -p ${path}
+    # convert file name from camelCase to snake-case
+    fullfile=$(echo "${file}" | awk '{
+
+  while ( match($0, /(.*)([a-z0-9])([A-Z])(.*)/, cap))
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -94,4 +94,8 @@ install-sdk:
 	chmod +x ${GOPATH}/bin/operator-sdk
 .PHONY: install-sdk
 
+mixins:
+	cd jsonnet && \
+	jb update && \
+	./build-jsonnet.sh make.jsonnet
```
guymguym:
```suggestion
	./build-jsonnet.sh make.jsonnet
.PHONY: mixins
```

jsonnet/manifests/noobaa-prometheus-rules.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: monitoring.coreos.com/v1
```
nimrod-becker: @umangachapagain for my understanding, why do we need the manifest committed? Can't we bring it from the dependent repo?

jsonnet/build-jsonnet.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+set -e
+set -x
+# only exit with zero if all commands of the pipeline exit successfully
+set -o pipefail
+
+prefix="manifests"
```
nimrod-becker: it seems like there are a few dependencies for this to run (jb, jq, jsonnet, mapfile, gojsontoyaml ) we should add them to the makefile.

Makefile
```diff
@@ -94,4 +94,8 @@ install-sdk:
 	chmod +x ${GOPATH}/bin/operator-sdk
 .PHONY: install-sdk
 
+mixins:
```
nimrod-becker: let's add mixins as a dependency to the "all" target

jsonnet/manifests/noobaa-prometheus-rules.yaml
umangachapagain: Ideally, noobaa-mixins should not have this manifest committed. It's scope should be limited to generating the  [alerting rule](https://prometheus.io/docs/prometheus/latest/configuration/alerting_rules/#defining-alerting-rules) file. (which is why I was thinking of a cleanup PR there)

Since it is the use-case of the noobaa-operator that this alerting file needs to be deployed in a K8s/Openshift environment, it should be responsible for handling creation of this manifest. (with all the customized metadata)

That being said, we could choose to generate this file at build time and copy it into the container (without a commit it here). But, from what I have seen, all other projects commit this manifest when something changes. (makes it easier to track what went into the container and user can always directly use this for manual deploy)

guymguym: Hi @umangachapagain I agree - it makes sense, just like when generating the CRD's using operator-sdk generate...

jsonnet/build-jsonnet.sh
umangachapagain: added dependencies like `jb`, `jsonnet` and `gojsontoyaml`. But, `mapfile` is an in-built bash command and `jq` [installation](https://stedolan.github.io/jq/download/) is a little bit tricky. How do you want to handle this?

jsonnet/manifests/noobaa-prometheus-rules.yaml
umangachapagain: That is a good example. Never thought of it :P

jsonnet/build-jsonnet.sh
nimrod-becker: I think we can make a target for install-deps. We will mainly call it from our CI in which we control the platform

umangachapagain: So i'll go ahead and add target to install 64-bit jq binary?

nimrod-becker: Yes, we will call it under travis-ci.

jsonnet/manifests/noobaa-prometheus-rules.yaml
jeniawhite: There is a bug in this CRD. It won't apply.
I've got a PR in NooBaa Mixins that I'm waiting to merge which fixes that.
It would be great if you could update the package.lock with the newer master and generate this CRD once again with the new commit.
I will update here once I merge the PR.

umangachapagain: Can you provide a link to that PR?

jeniawhite: https://github.com/noobaa/noobaa-mixins/pull/11
I've merged the PR.
Please update the package.lock and generate the CRD with the new changes.
Thank you.

umangachapagain: done.

guymguym:
Hi @umangachapagain 

Many thanks for submitting the PR!

We reviewed the process again and although the automation of pulling mixins sources and then building in the operator can work, we think perhaps this was a bit of an overkill.

I think you also suggested it before that we can build the mixins output yaml and then just make a PR to the operator to update the rules yaml, and it seems like a simple a straightforward solution to our needs.

@jeniawhite already updated the makefile of mixins to build the rules yaml, so that the process will be easy.

If this makes sense I suggest we close this PR and only take what is needed from it to the mixins build process.
Thanks!

PR #26, opened 2019-08-08
WIP:  Add the standard custom-resource-status conditions.

sacharya:
This PR replaces the current SystemCondition in favor of [custom-resource-status ](https://github.com/openshift/custom-resource-status) conditions, which provides some level of standardization and best practices to managing the status of custom resources. It specifically gives you standard conditionType like Available, Progressing, Degraded and Upgradeable that meta-operator like ocs-operator can then reply on.


guymguym:
Thanks @sacharya

guymguym:
@sacharya  Note that the build failed with the following error:
```
Build failed: API has been changed but the deep copy functions aren't up to date. Run 'make gen-api' and update your PR.
Makefile:136: recipe for target 'gen-api-fail-if-dirty' failed
```
But anyway I can handle it so no worries.
Thanks again for the PR work!

PR #46, opened 2019-09-10
fixing most gosec issues

jackyalbo:
1. fixing error handling issues, if by panic or by ignore
2. fixing G304: Potential file inclusion via variable by using filepath.Clean in bundle.go
3. NOSEC in cli-test file so gosec will disregard it

still fails on two types of errors:
1. not verifying certificate: G402: TLS InsecureSkipVerify set true - need to be fixed with a valid certificate
2. G101: Potential hardcoded credentials -  in the build output - seems like we need gosec to skip the build/_output directory when running

### DIFF
--- a/pkg/operator/manager.go
+++ b/pkg/operator/manager.go
@@ -63,11 +63,11 @@ func RunOperator(cmd *cobra.Command, args []string) {
 		log.Fatalf("Failed AddToManager: %s", err)
 	}
 
-	mgr.Add(manager.RunnableFunc(func(stopChan <-chan struct{}) error {
+	util.Panic(mgr.Add(manager.RunnableFunc(func(stopChan <-chan struct{}) error {
 		system.RunOperatorCreate(cmd, args)
 		<-stopChan
 		return nil
-	}))
+	})))
 
 	// // Create Service object to expose the metrics port.
 	// _, err = metrics.CreateMetricsService(util.Context(), config, metricsPort)
--- a/pkg/system/system.go
+++ b/pkg/system/system.go
@@ -299,7 +299,7 @@ func RunList(cmd *cobra.Command, args []string) {
 func RunYaml(cmd *cobra.Command, args []string) {
 	sys := LoadSystemDefaults()
 	p := printers.YAMLPrinter{}
-	p.PrintObj(sys, os.Stdout)
+	util.Panic(p.PrintObj(sys, os.Stdout))
 }
 
 // RunStatus runs a CLI command
@@ -389,7 +389,7 @@ func RunStatus(cmd *cobra.Command, args []string) {
 		log.Println("")
 	} else {
 		log.Printf("❌ System Phase is %q\n", r.NooBaa.Status.Phase)
-		CheckWaitingFor(r.NooBaa)
+		util.IgnoreError(CheckWaitingFor(r.NooBaa))
 	}
 }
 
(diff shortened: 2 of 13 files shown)


pkg/system/system.go
```diff
@@ -433,7 +436,7 @@ func WaitReady() bool {
 			return false, fmt.Errorf("SystemPhaseRejected")
 		}
 		if sys.Status.Phase != nbv1.SystemPhaseReady {
-			return CheckWaitingFor(sys)
+			return false, CheckWaitingFor(sys)
```
guymguym: @jackyalbo Was this a security review comment?

pkg/system/system.go
```diff
@@ -385,7 +385,10 @@ func RunStatus(cmd *cobra.Command, args []string) {
 		log.Println("")
 	} else {
 		log.Printf("❌ System Phase is %q\n", r.NooBaa.Status.Phase)
-		CheckWaitingFor(r.NooBaa)
+		err := CheckWaitingFor(r.NooBaa)
+		if err != nil {
+			util.Logger().Fatalf("❌ %s", err)
```
guymguym: This print is not redundant because CheckWaitingFor is already printing in all of its cases - that's the reason we wrote it - to print.
So I would just use `util.IgnoreError(CheckWaitingFor(r.NooBaa))`.

pkg/operator/manager.go
```diff
@@ -63,11 +63,13 @@ func RunOperator(cmd *cobra.Command, args []string) {
 		log.Fatalf("Failed AddToManager: %s", err)
 	}
 
-	mgr.Add(manager.RunnableFunc(func(stopChan <-chan struct{}) error {
+	if err := mgr.Add(manager.RunnableFunc(func(stopChan <-chan struct{}) error {
```
guymguym: I would use `util.Panic()` as an error is unexpected here.

pkg/util/fastmapper.go
```diff
@@ -55,7 +55,7 @@ func (m *FastRESTMapper) Discover() error {
 			VersionedResources: make(map[string][]metav1.APIResource),
 		}
 		grs = append(grs, gr)
-		wg.Start(func() { m.DiscoverGroup(gr) })
+		wg.Start(func() { Panic(m.DiscoverGroup(gr)) })
```
guymguym: I don't think Panic is suitable here. Discover can fail due to disconnection.
Lets create a new func `util.LogError()` that will log if there was an error but continue.

pkg/util/fastmapper.go
```diff
@@ -80,7 +80,7 @@ func (m *FastRESTMapper) DiscoverGroup(gr *restmapper.APIGroupResources) error {
 // DiscoverOnError check if the error is NoMatchError and calls discover
 func (m *FastRESTMapper) DiscoverOnError(err error) bool {
 	if meta.IsNoMatchError(err) {
-		m.Discover()
+		Panic(m.Discover())
```
guymguym: Same as previous comment - use a new `util.LogError()`

pkg/system/system.go
jackyalbo: @guymguym Not a security issue, just wanted to return once the false,err instead of many times in the function.

guymguym:
@jackyalbo Build failed:
```
pkg/crd/crd.go:126:6: syntax error: unexpected LoadCrds, expecting (
```

PR #48, opened 2019-09-10
some changes to default backing store reconcilation

dannyzaken:
- perform `EnsureCredentialsRequestAndGetBucketName` only in phase 2
- in phase 4 check if credential request exists. if not assumes it is not supported 
- if exists then waiting for secret and creating default backing store


guymguym:
Can you rebase?

pkg/system/phase2_creating.go
```diff
@@ -133,3 +136,70 @@ func (r *Reconciler) SetDesiredCoreApp() {
 		// r.Own(pvc)
 	}
 }
+
+// EnsureCredentialsRequestAndGetBucketName creates a CredentialsRequest resource if neccesary and returns
+// the bucket name allowed for the credentials. nil is returned if cloud credentials are not supported
+func (r *Reconciler) EnsureCredentialsRequestAndGetBucketName(log *logrus.Entry) error {
```
guymguym: I suggest a simpler name, and also seems like you don't need to pass the logger:
```suggestion
func (r *Reconciler) ReconcileCredentialsRequest() error {
```

PR #79, opened 2019-10-03
return non-zero exit code on error

### LINKED ISSUE #77: When bucketclass create failed because we gave it a flag with no parameters the exit status is 0.
When bucketclass create failed because i gave it a flag with no parameters the exit status is 0.

$ build/_output/bin/noobaa-operator-local -n liran --operator-image noobaa/noobaa-operator:2.0.1-rc.1 bucketclass create --backingstores
Error: flag needs an argument: --backingstores

Options:
      --backingstores=[]: Set first tier backing stores (use commas or multiple flags)
      --placement='': Set first tier placement policy - Mirror | Spread | "" (empty defaults to single backing store)

Usage:
  noobaa bucketclass create <bucket-class-name> [flags] [options]

$ echo $?
0

dannyzaken:
Fixes #77

### DIFF
--- a/pkg/cli/cli.go
+++ b/pkg/cli/cli.go
@@ -3,6 +3,7 @@ package cli
 import (
 	"flag"
 	"math/rand"
+	"os"
 	"time"
 
 	"github.com/noobaa/noobaa-operator/v2/pkg/backingstore"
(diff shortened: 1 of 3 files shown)


dannyzaken:
but `Execute()` is called in main, `cli.go` only returns the command. am I missing something?

dannyzaken:
> I want the main to be a one liner.

I think that in go four liner is the new one liner 🤷‍♂

pkg/cli/cli.go
```diff
@@ -46,6 +47,14 @@ const ASCIILogo2 = `
 #      N O O B A A      # 
 `
 
+//Run runs
```
guymguym: priceless comment! :)

PR #82, opened 2019-10-06
Create test cli flow for travis

liranmauda:
Create test cli flow for travis


liranmauda:
Fails due to #136

guymguym:
That's awesom.!

PR #83, opened 2019-10-06
print core and operator versions in Status.Readme

dannyzaken:


pkg/system/phase4_configuring.go
```diff
@@ -60,6 +60,16 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 		}
 	}
 
+	// update noobaa-core version in reconciler struct
+	r.NBClient.SetAuthToken(r.SecretOp.StringData["auth_token"])
+	systemInfo, err := r.NBClient.ReadSystemAPI()
```
guymguym: Small thing - I would move this code into a small function and name it `ReconcileReadSystem()` and keep the result on the reconciler as `r.SystemInfo` as we will probably want it soon for more tasks in configuring the system. You still want to also set `r.CoreVersion = r.SystemInfo.Version` for the template.

pkg/system/phase4_configuring.go
```diff
@@ -60,6 +60,16 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 		}
 	}
 
+	// update noobaa-core version in reconciler struct
+	r.NBClient.SetAuthToken(r.SecretOp.StringData["auth_token"])
```
guymguym: I think this is redundant

guymguym:
@dannyzaken Please rebase

pkg/system/phase4_configuring.go
dannyzaken: didn't work without it for some reason

dannyzaken: I think I know why. I'll fix it

dannyzaken: I remove the `SetAuthToken` in phase 3, since it sets an invalid token (secret was not read yet). setting auth token in `ReconcileSecretOp` instead

PR #85, opened 2019-10-06
port-forward to allow CLI direct access to RPC

guymguym:
This allows `noobaa bucket ...` and `noobaa pvstore` on cloud clusters where the nodes network is private.

Including also these small changes:
- save generated deploy bundle in git to avoid error on clean repo.
- go.mod - tidy and move the `replace` section down.
- align optional feature prints - to look nicer.


nimrod-becker:
This is awesome

PR #97, opened 2019-10-23
fixed default rgw backing store flow

dannyzaken:
* when not running in AWS, trying to create a default backing store on RGW if available
* listing `CephObjectstore` resources in the same namespace, and using the first one that is found as a target S3 compatible resource
* In backingstore reconciler added support to RGW naming convention for access\secret keys (camel case)

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -410,13 +410,7 @@ func (r *Reconciler) MakeExternalConnectionParams() (*nb.AddExternalConnectionPa
 		Name: r.BackingStore.Name,
 	}
 
-	// fix keys to handle the case that the secret holds lowercase keys
-	if r.Secret.StringData["AWS_ACCESS_KEY_ID"] == "" && r.Secret.StringData["aws_access_key_id"] != "" {
-		r.Secret.StringData["AWS_ACCESS_KEY_ID"] = r.Secret.StringData["aws_access_key_id"]
-	}
-	if r.Secret.StringData["AWS_SECRET_ACCESS_KEY"] == "" && r.Secret.StringData["aws_secret_access_key"] != "" {
-		r.Secret.StringData["AWS_SECRET_ACCESS_KEY"] = r.Secret.StringData["aws_secret_access_key"]
-	}
+	r.fixAlternateKeysNames()
 
 	switch r.BackingStore.Spec.Type {
 
@@ -518,6 +512,30 @@ func (r *Reconciler) MakeExternalConnectionParams() (*nb.AddExternalConnectionPa
 	}
 
 	return conn, nil
+}
+
+func (r *Reconciler) fixAlternateKeysNames() {
+
+	alternateAccessKeyNames := []string{"aws_access_key_id", "AccessKey"}
+	alternateSecretKeyNames := []string{"aws_secret_access_key", "SecretKey"}
+
+	if r.Secret.StringData["AWS_ACCESS_KEY_ID"] == "" {
+		for _, key := range alternateAccessKeyNames {
+			if r.Secret.StringData[key] != "" {
+				r.Secret.StringData["AWS_ACCESS_KEY_ID"] = r.Secret.StringData[key]
+				break
+			}
+		}
+	}
+
+	if r.Secret.StringData["AWS_SECRET_ACCESS_KEY"] == "" {
+		for _, key := range alternateSecretKeyNames {
+			if r.Secret.StringData[key] != "" {
+				r.Secret.StringData["AWS_SECRET_ACCESS_KEY"] = r.Secret.StringData[key]
+				break
+			}
+		}
+	}
 }
 
 // ReconcileExternalConnection handles the external connection using noobaa api
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -10,10 +10,12 @@ import (
 	"github.com/noobaa/noobaa-operator/v2/pkg/options"
 	"github.com/noobaa/noobaa-operator/v2/pkg/util"
 	cloudcredsv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
+	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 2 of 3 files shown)


pkg/backingstore/reconciler.go
```diff
@@ -520,6 +514,30 @@ func (r *Reconciler) MakeExternalConnectionParams() (*nb.AddExternalConnectionPa
 	return conn, nil
 }
 
+func (r *Reconciler) fixAlternateKeysNames() {
+
+	alternateAccessKeyNames := []string{"aws_access_key_id", "AccessKey"}
+	alternateSecretKeyNames := []string{"aws_access_key_id", "SecretKey"}
```
guymguym: copy-pasta-typo:
```suggestion
	alternateSecretKeyNames := []string{"aws_secret_access_key", "SecretKey"}
```

pkg/system/phase2_creating.go
```diff
@@ -28,7 +31,7 @@ func (r *Reconciler) ReconcilePhaseCreating() error {
 	// the credentials that are created by cloud-credentials-operator sometimes take time
 	// to be valid (requests sometimes returns InvalidAccessKeyId for 1-2 minutes)
 	// creating the credential request as early as possible to try and avoid it
-	err := r.ReconcileCredentialsRequest()
+	err := r.ReconcileBackinStoreCredentials()
```
guymguym: name typo
```suggestion
	err := r.ReconcileBackingStoreCredentials()
```

pkg/system/phase2_creating.go
```diff
@@ -168,13 +171,49 @@ func (r *Reconciler) SetDesiredCoreApp() {
...
+	return r.ReconcileCephObjectstoreCredentials()
+
+}
+
+// ReconcileCephObjectstoreCredentials creates a ceph objectstore user if a ceph objectstore exists in the same namespace
+func (r *Reconciler) ReconcileCephObjectstoreCredentials() error {
+	r.Logger.Info("Not running in AWS. will attempt to create a ceph objectstore user")
+	util.KubeCheck(r.CephObjectstoreUser)
+	// create user if not already exists
+	if r.CephObjectstoreUser.UID == "" {
```
guymguym: How about returning if UID != "" to keep the rest of the function flat?
```
if r.CephObjectstoreUser.UID != "" {
    return nil
}
...
```

pkg/system/phase2_creating.go
```diff
@@ -168,13 +171,49 @@ func (r *Reconciler) SetDesiredCoreApp() {
...
+		// list ceph objectstores and pick the first one
+		cephObjectStoresList := &cephv1.CephObjectStoreList{}
+		if !util.KubeList(cephObjectStoresList, &client.ListOptions{Namespace: options.Namespace}) {
+			return fmt.Errorf("failed to list ceph objectstores")
+		}
+		if len(cephObjectStoresList.Items) == 0 {
+			// no object stores
+			return nil
+		}
+		// for now take the first one. need to decide what to do if multiple objectstores in one namespace
```
guymguym: Great, but I suggest you print the list of results so we can debug it.

pkg/system/phase2_creating.go
```diff
@@ -168,13 +171,49 @@ func (r *Reconciler) SetDesiredCoreApp() {
...
+	util.KubeCheck(r.CephObjectstoreUser)
+	// create user if not already exists
+	if r.CephObjectstoreUser.UID == "" {
+		// list ceph objectstores and pick the first one
+		cephObjectStoresList := &cephv1.CephObjectStoreList{}
+		if !util.KubeList(cephObjectStoresList, &client.ListOptions{Namespace: options.Namespace}) {
+			return fmt.Errorf("failed to list ceph objectstores")
+		}
+		if len(cephObjectStoresList.Items) == 0 {
+			// no object stores
```
guymguym: Add log.Warn?

pkg/system/phase2_creating.go
```diff
@@ -168,13 +171,49 @@ func (r *Reconciler) SetDesiredCoreApp() {
...
+// ReconcileCephObjectstoreCredentials creates a ceph objectstore user if a ceph objectstore exists in the same namespace
+func (r *Reconciler) ReconcileCephObjectstoreCredentials() error {
+	r.Logger.Info("Not running in AWS. will attempt to create a ceph objectstore user")
+	util.KubeCheck(r.CephObjectstoreUser)
+	// create user if not already exists
+	if r.CephObjectstoreUser.UID == "" {
+		// list ceph objectstores and pick the first one
+		cephObjectStoresList := &cephv1.CephObjectStoreList{}
+		if !util.KubeList(cephObjectStoresList, &client.ListOptions{Namespace: options.Namespace}) {
+			return fmt.Errorf("failed to list ceph objectstores")
```
guymguym: Is this right to error here? What if we install noobaa without ceph? 
We should make sure that path works well too.

pkg/system/phase4_configuring.go
```diff
@@ -334,6 +345,29 @@ func (r *Reconciler) ReconcileOBCStorageClass() error {
 	return nil
 }
 
+func (r *Reconciler) createS3BucketForeBackingStore(s3Config *aws.Config, bucketName string) error {
```
guymguym: typo
```suggestion
func (r *Reconciler) createS3BucketForBackingStore(s3Config *aws.Config, bucketName string) error {
```

pkg/system/phase2_creating.go
```diff
@@ -168,13 +171,49 @@ func (r *Reconciler) SetDesiredCoreApp() {
...
+func (r *Reconciler) ReconcileBackinStoreCredentials() error {
+	if util.IsAWSPlatform() {
+		return r.ReconcileAWSCredentials()
+	}
+	return r.ReconcileCephObjectstoreCredentials()
+
+}
+
+// ReconcileCephObjectstoreCredentials creates a ceph objectstore user if a ceph objectstore exists in the same namespace
+func (r *Reconciler) ReconcileCephObjectstoreCredentials() error {
```
guymguym:
```suggestion
func (r *Reconciler) ReconcileRGWCredentials() error {
```

pkg/system/phase2_creating.go
dannyzaken: you're right. shouldn't error here

pkg/system/phase2_creating.go
dannyzaken: It is a pretty long name😩 I'll change it

PR #102, opened 2019-10-24
Create of secure OpenShift routes for NooBaa services (mgmt/s3)- OpenShift only

nb-ohad:
- Generate a secret holding a certificate for each service (using OpenShift serving cert mechanism)
- Mount the secrets as folders in the noobaa-core pod (to be used by the core  for TLS communication)
- Create a secure route for each service the serve using the ingress operator cert and re-encrypt using the cert generated for each service.

### DIFF
--- a/deploy/internal/route-mgmt.yaml
+++ b/deploy/internal/route-mgmt.yaml
@@ -1,0 +1,16 @@
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  labels:
+    app: noobaa
+  name: noobaa-mgmt
+spec:
+  port:
+    targetPort: mgmt-https
+  tls:
+    termination: reencrypt
+  to:
+    kind: Service
+    name: noobaa-mgmt
+    weight: 100
+  wildcardPolicy: None
--- a/deploy/internal/service-mgmt.yaml
+++ b/deploy/internal/service-mgmt.yaml
@@ -8,6 +8,7 @@ metadata:
     prometheus.io/scrape: "true"
     prometheus.io/scheme: http
     prometheus.io/port: "8080"
+    service.beta.openshift.io/serving-cert-secret-name: noobaa-mgmt-serving-cert
 spec:
   type: LoadBalancer
   selector:
--- a/deploy/internal/statefulset-core.yaml
+++ b/deploy/internal/statefulset-core.yaml
@@ -35,6 +35,14 @@ spec:
       volumes:
       - name: logs
         emptyDir: {}
+      - name: mgmt-secret
+        secret:
+          secretName: noobaa-mgmt-serving-cert
+          optional: true
+      - name: s3-secret
+        secret:
+          secretName: noobaa-s3-serving-cert
+          optional: true
       initContainers:
 #----------------#
 # INIT CONTAINER #
@@ -58,6 +66,12 @@ spec:
         volumeMounts:
         - name: logs
           mountPath: /log
+        - name: mgmt-secret
+          mountPath: /etc/mgmt-secret
+          readOnly: true
+        - name: s3-secret
+          mountPath: /etc/s3-secret
+          readOnly: true
         readinessProbe:
           tcpSocket:
             port: 6001 # ready when s3 port is open
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -46,6 +46,12 @@ func (r *Reconciler) ReconcilePhaseCreating() error {
 	if err := r.ReconcileObject(r.ServiceS3, r.SetDesiredServiceS3); err != nil {
 		return err
 	}
+	if err := r.ReconcileObjectOptional(r.RouteMgmt, nil); err != nil {
+		return err
+	}
+	if err := r.ReconcileObjectOptional(r.RouteS3, nil); err != nil {
+		return err
+	}
 
 	return nil
 }
--- a/pkg/system/phase3_connecting.go
+++ b/pkg/system/phase3_connecting.go
@@ -10,6 +10,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	routev1 "github.com/openshift/api/route/v1"
 )
 
 // ReconcilePhaseConnecting runs the reconcile phase
@@ -21,8 +22,8 @@ func (r *Reconciler) ReconcilePhaseConnecting() error {
 		"noobaa operator started phase 3/4 - \"Connecting\"",
 	)
 
-	r.CheckServiceStatus(r.ServiceMgmt, &r.NooBaa.Status.Services.ServiceMgmt, "mgmt-https")
-	r.CheckServiceStatus(r.ServiceS3, &r.NooBaa.Status.Services.ServiceS3, "s3-https")
+	r.CheckServiceStatus(r.ServiceMgmt, r.RouteMgmt, &r.NooBaa.Status.Services.ServiceMgmt, "mgmt-https")
+	r.CheckServiceStatus(r.ServiceS3, r.RouteS3, &r.NooBaa.Status.Services.ServiceS3, "s3-https")
 
 	// initialize the noobaa client for making calls to the server.
 	if len(r.NooBaa.Status.Services.ServiceMgmt.NodePorts) == 0 {
@@ -59,7 +60,7 @@ func (r *Reconciler) ReconcilePhaseConnecting() error {
 }
 
 // CheckServiceStatus populates the status of a service by detecting all of its addresses
-func (r *Reconciler) CheckServiceStatus(srv *corev1.Service, status *nbv1.ServiceStatus, portName string) {
+func (r *Reconciler) CheckServiceStatus(srv *corev1.Service, route *routev1.Route, status *nbv1.ServiceStatus, portName string) {
 
 	log := r.Logger.WithField("func", "CheckServiceStatus").WithField("service", srv.Name)
 	*status = nbv1.ServiceStatus{}
@@ -105,6 +106,14 @@ func (r *Reconciler) CheckServiceStatus(srv *corev1.Service, status *nbv1.Servic
 		status.InternalDNS = append(
 			status.InternalDNS,
 			fmt.Sprintf("%s://%s.%s:%d", proto, srv.Name, srv.Namespace, servicePort.Port),
+		)
+	}
+
+	// Routes IP:Port (of the service) {
+	if route.Spec.Host != "" {
+		status.ExternalDNS = append(
+			status.ExternalDNS,
+			fmt.Sprintf("%s://%s", proto, route.Spec.Host),
 		)
 	}
 
--- a/deploy/internal/route-s3.yaml
+++ b/deploy/internal/route-s3.yaml
@@ -1,0 +1,16 @@
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  labels:
+    app: noobaa
+  name: s3
+spec:
+  port:
+    targetPort: s3-https
+  tls:
+    termination: reencrypt
+  to:
+    kind: Service
+    name: s3
+    weight: 100
+  wildcardPolicy: None
--- a/deploy/internal/service-s3.yaml
+++ b/deploy/internal/service-s3.yaml
@@ -4,6 +4,8 @@ metadata:
   name: s3
   labels:
     app: noobaa
+  annotations:
+    service.beta.openshift.io/serving-cert-secret-name: 'noobaa-s3-serving-cert'
 spec:
   type: LoadBalancer
   selector:
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -97,3 +97,14 @@ rules:
   - update
   - list
   - watch
+- apiGroups:
+  - route.openshift.io
+  resources:
+  - routes
+  verbs:
+  - get
+  - create
+  - update
+  - list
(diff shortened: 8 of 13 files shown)


ron1:
For performance reasons, consider defining the s3 route as passthrough rather than reencrypt.

Also consider adding property sslCertificateRef to the NooBaa CRD which references a Kubernetes TLS Secret. This is similar to the Rook-Ceph CephObjectStore sslCertificateRef used by Ceph RGW. This would eliminate the need for Operator users to manually create a SSL Certificate in the unfamiliar zip format and upload the cert via the Admin UI. Should I open another issue for this request?

nb-ohad:
@ron1 Thaks for the feedback. 
I understand the performance considerations regarding the re-encrypt TLS termination. 
But changing it to passthrough defeats the purpose because it will not use the cluster ingress cert for encryption of traffic from outside the cluster.

Regarding S3 traffic Inside the cluster (where the client lives in a pod inside the cluster) I am assuming that the service's internal DNS name will be used as an endpoint, which is served using the serving-cert certificate which we load into the NooBaa brain. This way no performance penalties occur. 

Also, I am not totally convinced that we need a prop on the NooBaa CRD because a client can create/update the cert secret object on its own using any standard k8s tool and he does not have to specifically use the management console tp upload a new cert. This is especially important on non-openshift envs where, by default, there is no service that automatically generates certificates.
The only consideration is that the secret name should be publicly known (which is not that different from a name of an optional prop in a custom CRD).

Saying that I still see the convenience of the suggested method.

pkg/system/phase3_connecting.go
```diff
@@ -108,6 +109,14 @@ func (r *Reconciler) CheckServiceStatus(srv *corev1.Service, status *nbv1.Servic
 		)
 	}
 
+	// Routes IP:Port (of the service) {
+	if route.Spec.Host != "" {
+		status.ExternalDNS = append(
+			status.ExternalDNS,
+			fmt.Sprintf("%s://%s:%d", proto, route.Spec.Host, 443),
```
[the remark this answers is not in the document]
nb-ohad: We don't by I didn't want to break the format of the displayed addresses.

deploy/internal/route-mgmt.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  labels:
+    app: noobaa
+  name: SYSNAME-mgmt
```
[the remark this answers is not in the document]
nb-ohad: I just copied the convention from the service's YAML files. Should we get rid of them in all YAMLs?

deploy/internal/route-mgmt.yaml
```diff
@@ -0,0 +1,16 @@
...
+spec:
+  port:
+    targetPort: mgmt-https
+  tls:
+    termination: reencrypt
+  to:
+    kind: Service
+    name: SYSNAME-mgmt
```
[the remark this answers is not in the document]
nb-ohad: Same as above

nb-ohad:
@ron1 
I Just wanted to clarify: 
As part of the process, we are mounting the serviing-cert secrets into the NooBaa brain pod and automatically using them in for the brains' https servers, so no manual upload of a cert zip from the management console is needed anymore. 
As a matter of fact, I removed the upload code and UI from the brain's code.

You can see the corresponding changes at 
https://github.com/noobaa/noobaa-core/pull/5802

ron1:
@nb-ohad Thanks much for your response.

Up until now, we have been using Rook-Ceph RGW for our object storage on OCP 3.11. Since we are happy with the block storage available from our cloud provider, since Rook-Ceph is quite a complex deployment for just object storage alone, and since OCP 3.11 does not support dynamic provisioning of block volumes, we decided to evaluate NooBaa.

Our OCP 3.11 deployment includes a Custom Certificate as described here: https://docs.openshift.com/container-platform/3.11/install_config/certificate_customization.html. Our current Rook-Ceph CephObjectStore RGW sslCertificateRef references a secret with a custom certificate with public name ```s3.apps.<clustername>.<domainname>``` and SAN ```rook-ceph-rgw-<objectstorename>.rook-ceph.svc.cluster.local```. Our Rook-Ceph RGW deployment includes a passthrough Route with Host property ```s3.apps.<clustername>.<domainname>```. The RGW pods successfully use the the custom certificate with the SAN to terminate both external and internal clients. We would like to achieve the same functionality with the S3 NooBaa endpoint where the custom certificate public name is the same and the SAN is simply replaced with ```s3.noobaa.svc.cluster.local```. As far as the noobaa-mgmt endpoint is concerned, it would be nice for it to offer the same flexibility as the s3 endpoint. However, since only our cluster operators access that endpoint and since the volume of traffic on this endpoint is small, we don't mind using port-forwarding to access it if needed.

Let me know if you have questions.

nb-ohad:
@ron1 Thank you for the detailed explanation 
I think I understand your OCP 3.11 configuration.

Things changed a little as of OCP 4.0+, an OpenShift cluster uses different certificates (and CAs) for different kinds of traffic.

For k8s API traffic, it uses an API server certificate defaulted to a CA that is local to the cluster and its cert is distributed to all pods as CA.crt file.
see https://docs.openshift.com/container-platform/4.2/authentication/certificates/api-server.html

For internal service communication, OpenShift recommends using the serving cert facility which has it's own CA which is also local to the cluster and has a different CA cert which is distributed to all pods as a service-ca.crt file (so any service can add it to its root CA list)) 
see https://docs.openshift.com/container-platform/4.2/authentication/certificates/service-serving-certificate.html
 
For ingress traffic, it uses an ingress operator certificate which is, by default, self-signed but can and is recommended to be replaced by a cert that was issued by a well-known authority.
see https://docs.openshift.com/container-platform/4.2/authentication/certificates/replacing-default-ingress-certificate.html

I imagen that it is possible to configure a cluster in a way where the ingress certificate and the serving cert certificate (the CA) are the same with the ingress public name as the main name and the cluster local name is its SAN. This will allow a passthrough configuration on the routes TLS termination policy but I can see a couple of drawbacks:
- First, this is not the recommended or the standard way to configure a cluster and as such this will not be the out of the box solution.
- Second, it ignores completely the CA and cert expiry and rotation mechanism that is implemented for the serving cert facility. And also prevent the cluster operator from manually expiring the current cert by just deleting the current generate cert serving cert (in case the cert key was compromised).
- Third, it enforces service traffic inside the cluster to use the same custom certificate.

Saying all of the above we do not prevent this kind of configuration. If you want to do so, you can configure the cluster to use the same cert for ingress and the serving cert facility as described in the links above (with the public name set to ```*.apps.<clustername>.<domainname>``` and the SAN configured to the cluster local name) then just manually edit the s3/noobaa-mgmt route object and change the TLS termination field to passthrough, Noobaa will automatically use the certificates stored in the noobaa-mgmt-serving-cert/s3-serving-cert without any prejudice.

I think that we need to evaluate if this configuration is a common case and if so support it directly in the NooBaa CRD as a secondary form of configuration.

guymguym:
> As a matter of fact, I removed the upload code and UI from the brain's code.

@nb-ohad As discussed - can you keep the upload code? Perhaps just disable/hide the button temporarily?

guymguym:
@tamireran Can you take a look at the requirements that @ron1 raised above in https://github.com/noobaa/noobaa-operator/pull/102#issuecomment-546154230 ?

ron1:
@nb-ohad Again, thanks for the timely response.

An OCP S3 storage endpoint has unique requirements given the potential volume of external traffic it may support. Since NooBaa is an OCP S3 storage provider, I would expect it to offer similar optimizations to those available in rook ceph rgw. I'll see if I can manually achieve the configuration you described in your comment. For folks like me migrating to NooBaa from ceph rgw, it would be great if minimal docs were available to help achieve such a configuration.

ron1:
@nb-ohad If I manually populate the s3-serving-cert Secret, how do I prevent the service-ca controller from overwriting that Secret with its own? Do I accomplish this on OCP 4.2 by simply patching the Service resource and removing the ```service.beta.openshift.io/serving-cert-secret-name=<secret-name>``` annotation?

Also, will noobaa-operator reconciliation overwrite any customizations I make to the noobaa S3 Route or Service?

nb-ohad:
@ron1 In order to prevent the service-ca from overriding the secret you can safely remove the annotation from the service then update the ```noobaa-s3-serving-cert``` secret.
The NooBaa operator reconciler should not override your changes with a new annotation unless you delete the service object (in which case the reconciler will create a new service with the annotation on).
If you find that the operator behaves differently, its a bug, please let us know by submitting an issue. 

We had a discussion about reintroducing some form of UI in the management console that will allow you to perform this manual update more conveniently, which should take care of all the issues including the removal of the annotation. 

If this kind of configuration is indeed a "best practice", as you describe, I agree that we should make it as easy as possible to allow the user to get to that configuration. I am not this configuration will work as an "out of the box" configuration because it includes the introduction of a user-provided certificate.

@tamireran do you have any input on the subject at hand?

ron1:
@nb-ohad Thanks for the feedback.

Our preference is to configure as much as possible in CRs rather than via the Web UI. This allows us to use Kubernetes tools such as Kustomize, etc. to automate secret generation, configuration, deployment, etc.

As I mentioned here: https://github.com/noobaa/noobaa-operator/pull/102#issuecomment-546088870, we would prefer that an optional property like sslCertificateRef be added to the NooBaa CR. Additional optional properties for a corresponding Route Host and scheme might also be useful. If the sslCertificateRef property is absent, the behavior currently implemented in this PR would be applied. If the sslCertificateRef is provided, a passthrough Route would created and the Service would be created without the annotation.

guymguym:
@nb-ohad Please rebase

pkg/system/phase2_creating.go
```diff
@@ -52,6 +52,22 @@ func (r *Reconciler) ReconcilePhaseCreating() error {
 		return err
 	}
 
+	if err := r.ReconcileObject(r.RouteMgmt, nil); err != nil {
+		if meta.IsNoMatchError(err) || runtime.IsNotRegisteredError(err) {
```
guymguym: @nb-ohad After you rebase from master, see that I created a function to `ReconcileObjectOptional()` so no need to copy these conditions over and over.

guymguym:
Hi @ron1  
First of all thanks - I really appreciate your feedback and the time you took to explain your use case.

Today we continued to talk about on your comments and agreed that we need advanced DNS & SSL (route & certificate) settings. Still considering our priorities so it would be great to communicate further on your timelines.

I want to add a doc that highlights our current implementation of DNS & SSL and what can be overridden. Would you be interested in reviewing this doc once we edit it?

ron1:
@guymguym A doc that describes what can currently be customized/overridden as it relates to DNS and SSL would be great. I would be more than happy to review it.

deploy/internal/service-mgmt.yaml
```diff
@@ -8,6 +8,7 @@ metadata:
     prometheus.io/scrape: "true"
     prometheus.io/scheme: http
     prometheus.io/port: "8080"
+    service.beta.openshift.io/serving-cert-secret-name: noobaa-mgmt-serving-cert
```
guymguym: As @ron1 mentioned, is there a reason not to add the alpha annotation too? 
`service.alpha.openshift.io/serving-cert-secret-name`

nb-ohad: Only relevant to openshift3. Has no benefits on openshift4 and may cause incompatibility issues. I opted to treat earlier version of openshift the same way as we treat k8s

deploy/internal/statefulset-core.yaml
```diff
@@ -35,6 +35,14 @@ spec:
       volumes:
       - name: logs
         emptyDir: {}
+      - name: mgmt-secret
+        secret:
+          secretName: noobaa-mgmt-serving-cert
+          optional: true
```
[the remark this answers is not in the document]
nb-ohad: Yes it is and it is quite new. I missed it the first. Found it in the operator framework codebase. Tested it and it is working

deploy/internal/service-mgmt.yaml
guymguym: Got it.

ron1: @guymguym @nb-ohad Most OpenShift-friendly Kubernetes projects declare OCP 3.11+ support and do not "treat earlier version of openShift the same way as they treat k8s". I would expect upstream NooBaa to do likewise unless there is a major technical reason why OCP 3.11 cannot be supported. This is especially relevant since a). the project is RedHat-sponsored, b). the OpenShift 4 migration tool is still not yet GA AFAIK and c). OKD 4 is still several months off AFAIK. It is understandable that the current NooBaa focus is on supporting OCP 4.2. However, I would hope OCP 3.11 remains supported as well since Red Hat will support it into 2022. It would be great if NooBaa CI leveraged existing Red Hat resources to continue support for OCP 3.11.

guymguym: Good point @ron1. 
Are you planning to move to OCP 4 once the migration tool is available?
We would be happy to support it better upstream and will support you as best as we can.
You can help us by opening issues/PRs with suggested fixes.
However, at this point we don't have delivery pipeline for OCP 3.11 so cannot detect regressions without community help.
@tamireran FYI

ron1: @guymguym Unfortunately, we will be on OCP 3.11 for quite a while since OCP 4 introduces so many incompatible changes all the way down to the OS.

nb-ohad:
@ron1 @guymguym 
I created the requested doc, see PR
https://github.com/noobaa/noobaa-operator/pull/119

PR #105, opened 2019-10-27
changed CustomImage event to NooBaaImage event for every image change

dannyzaken:
* removed event on custom image when not using default image
* generate event for creation and every image change


guymguym (approved):

pkg/system/phase2_creating.go
```diff
@@ -162,6 +172,15 @@ func (r *Reconciler) SetDesiredCoreApp() {
 				}
 			}
 		}
+
+		if coreImageChanged {
+			// generate info event for the first creation of noobaa
```
guymguym: fix the comment...

PR #111, opened 2019-10-28
Addition of diagnostics option to CLI

jeniawhite:
Added diagnostics option to the CLI of the operator.

### DIFF
--- a/pkg/diagnose/diagnose.go
+++ b/pkg/diagnose/diagnose.go
@@ -1,0 +1,154 @@
+package diagnose
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	nbv1 "github.com/noobaa/noobaa-operator/v2/pkg/apis/noobaa/v1alpha1"
+	"github.com/noobaa/noobaa-operator/v2/pkg/options"
+	"github.com/noobaa/noobaa-operator/v2/pkg/util"
+	"github.com/sirupsen/logrus"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/spf13/cobra"
+)
+
+// Collector configuration for diagnostics
+type Collector struct {
+	folderName string
+	log        *logrus.Entry
+}
+
+// Cmd returns a CLI command
+func Cmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "diagnose",
+		Short: "Collect diagnostics",
+		Run:   RunCollect,
+	}
+	return cmd
+}
+
+// RunCollect runs a CLI command
+func RunCollect(cmd *cobra.Command, args []string) {
+	c := Collector{
+		folderName: fmt.Sprintf("%s_%s", "noobaa_diagnostics", time.Now().Format(time.RFC3339)),
+		log:        util.Logger(),
+	}
+
+	c.log.Println("Running collection of diagnostics")
+
+	err := os.Mkdir(c.folderName, os.ModePerm)
+	if err != nil {
+		c.log.Fatalf(`❌ Could not craete directory %s, reason: %s`, c.folderName, err)
+	}
+
+	c.CollectCR(&nbv1.BackingStoreList{
+		TypeMeta: metav1.TypeMeta{Kind: "BackingStoreList"},
+	})
+
+	c.CollectCR(&nbv1.BucketClassList{
+		TypeMeta: metav1.TypeMeta{Kind: "BucketClassList"},
+	})
+
+	c.CollectCR(&nbv1.NooBaaList{
+		TypeMeta: metav1.TypeMeta{Kind: "NooBaaList"},
+	})
+
+	corePodSelector, _ := labels.Parse("noobaa-core=" + options.SystemName)
+	c.CollectPodLogs(corePodSelector)
+
+	operatorPodSelector, _ := labels.Parse("noobaa-operator=deployment")
+	c.CollectPodLogs(operatorPodSelector)
+
+	// collectSystemMetrics()
+
+	c.ExportDiagnostics()
+}
+
+// CollectCR info
+func (c *Collector) CollectCR(list runtime.Object) {
+	gvk := list.GetObjectKind().GroupVersionKind()
+
+	if !util.KubeList(list, &client.ListOptions{Namespace: options.Namespace}) {
+		c.log.Printf(`❌ Failed to list %s\n`, gvk.Kind)
+		return
+	}
+
+	list.GetObjectKind().SetGroupVersionKind(gvk)
+
+	targetFile := fmt.Sprintf("%s/%s_crs.yaml", c.folderName, gvk.Kind)
+	util.SaveCRsToFile(list, targetFile)
+}
+
+// CollectPodLogs info
+func (c *Collector) CollectPodLogs(corePodSelector labels.Selector) {
+	corePodList := &corev1.PodList{}
+	currentPod := strings.Split(corePodSelector.String(), "=")[0]
+	if !util.KubeList(corePodList, &client.ListOptions{Namespace: options.Namespace, LabelSelector: corePodSelector}) {
+		return
+	}
+	if len(corePodList.Items) == 0 {
+		c.log.Printf(`❌ No %s pods found\n`, currentPod)
+		return
+	}
+
+	for i := range corePodList.Items {
+		corePod := &corePodList.Items[i]
+		podLogs, _ := util.GetPodLogs(*corePod)
+		for containerName, containerLog := range podLogs {
+			targetFile := fmt.Sprintf("%s/%s-%s.log", c.folderName, corePod.Name, containerName)
+			util.SaveStreamToFile(containerLog, targetFile)
+		}
+	}
+}
+
+// TODO: Use port forwarding (usePortForwarding in system.go)
+// func collectSystemMetrics() {
+// 	sys := getSystemObject()
+// 	mgmtAddress := sys.Status.Services.ServiceMgmt.ExternalDNS[0]
+// 	mgmtURL, err := url.Parse(mgmtAddress)
+// 	if err != nil {
+// 		log.Fatalf("failed to parse mgmt address %q. got error: %v", mgmtAddress, err)
+// 	}
+
+// 	targetAddress := fmt.Sprintf("%s/metrics/counter", mgmtURL.String())
+// 	log.Printf("JENIA THIS IS THE URL %s", targetAddress)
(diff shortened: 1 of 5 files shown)


pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,295 @@
...
+	collectBucketClasses()
+	log.Println("✅ Collected BucketClass CRs")
+	log.Println("Compressing and packaging diagnostics")
+	exportDiagnostics()
+	log.Println("✅ Compressed and packaged diagnostics")
+	// collectSystemMetrics()
+	log.Println("✅ Collected diagnostics")
+}
+
+func collectBackingStores() {
```
guymguym: Create a generic function `collectCR(list)` instead

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,295 @@
...
+			log.Errorf("%s\n", errRemoveAll)
+			log.Fatalf(`❌ Could not remove previous directory %s`, folderName)
+		}
+		log.Printf("✅ Removed previous directory %s\n", folderName)
+	}
+
+	log.Printf("Creating new directory %s\n", folderName)
+	errMkdir := os.Mkdir(folderName, os.ModePerm)
+	if errMkdir != nil {
+		log.Errorf("%s\n", errMkdir)
```
guymguym: Why separate error? Just add this error to the next Fatal?

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,295 @@
...
+	return cmd
+}
+
+// RunCollect runs a CLI command
+func RunCollect(cmd *cobra.Command, args []string) {
+	log = util.Logger()
+	log.Println("Running collection of diagnostics")
+
+	if len(args) != 1 || args[0] == "" {
+		log.Fatalf(`❌ Missing expected arguments: <diagnostics-package-name> %s`, cmd.UsageString())
```
guymguym: Why do we need this arg? Just create a random folder name.

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	folderName = args[0]
+
+	if _, err := os.Stat(folderName); !os.IsNotExist(err) {
+		log.Printf("Removing previous directory %s\n", folderName)
+		errRemoveAll := os.RemoveAll(folderName)
```
guymguym: I would prefer a random folder name with a prefix, and cleanup at the end

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,295 @@
...
+
+func collectOperatorPodLogs() {
+	operatorPodSelector, _ := labels.Parse("noobaa-operator=deployment")
+	collectPodLogs(operatorPodSelector)
+}
+
+func collectPodLogs(corePodSelector labels.Selector) {
+	log = util.Logger()
+	corePodList := &corev1.PodList{}
+	corePodErr := util.KubeClient().List(util.Context(),
```
guymguym: why not `util.KubeList()` ?

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+	if len(corePodList.Items) == 0 {
+		log.Printf(`❌ No %s pods found\n`, currentPod)
+		return
+	}
+	corePod := &corePodList.Items[0]
+	podLogs, errLogs := util.GetPodLogs(*corePod)
+	if errLogs != nil {
+		log.Errorf("%s\n", errLogs)
+		log.Printf(`❌ Could not get logs for pod %s\n`, corePod.Name)
```
guymguym: Can you move this error logging into `util.GetPodLogs`?

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,295 @@
...
+	}
+
+	for containerName, containerLog := range podLogs {
+		targetFile := fmt.Sprintf("%s/%s-%s.log", folderName, corePod.Name, containerName)
+		_, errSave := util.SaveSteamToFile(containerLog, targetFile)
+		if errSave != nil {
+			log.Errorf("%s\n", errSave)
+			log.Printf(`❌ Could not save logs for pod %s container %s\n`, corePod.Name, containerName)
```
guymguym: Can you move this error logging into `util.SaveSteamToFile()`?

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+}
+
+// TODO: This works with external address which is not always existing
```
guymguym: Use port forwarding:
https://github.com/noobaa/noobaa-operator/blob/fe62a28b78fd88a90677b14c16b20fccc80f734e/pkg/system/system.go#L588

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,295 @@
...
+		log.Errorf("%s\n", errGet)
+		log.Fatalf(`❌ Could not get NooBaa CR %s`, sys.Name)
+	}
+	return *sys
+}
+
+func exportDiagnostics() {
+	log = util.Logger()
+	var buf bytes.Buffer
+	errCompress := util.CompressedTarToBuffer(folderName, &buf)
```
guymguym: Why to memory??? Open the target file and write directly to it.

pkg/util/util.go
```diff
@@ -376,6 +382,76 @@ func GetPodStatusLine(pod *corev1.Pod) string {
...
+		return 0, errClose
+	}
+	defer f.Close()
+	// Probably the optimal buffer size for reads
+	buffer := make([]byte, 4096)
+	var count int64
+	var n int
+	var errRead error
+	for {
+		n, errRead = body.Read(buffer)
```
guymguym: Why not `io.Copy` instead of this loop?

pkg/util/util.go
```diff
@@ -376,6 +382,76 @@ func GetPodStatusLine(pod *corev1.Pod) string {
...
+			break
+		}
+		count += int64(n)
+		f.Write(buffer[:n])
+	}
+	return 0, nil
+}
+
+// SaveStringToFile info
+func SaveStringToFile(body string, path string) error {
```
guymguym: Why not change this to be `SaveYamlToFile`?

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,147 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/spf13/cobra"
+)
+
+var folderName string
```
guymguym: Create a struct for this state

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,147 @@
...
+	corePodList := &corev1.PodList{}
+	currentPod := strings.Split(corePodSelector.String(), "=")[0]
+	if !util.KubeList(corePodList, &client.ListOptions{Namespace: options.Namespace, LabelSelector: corePodSelector}) {
+		return
+	}
+	if len(corePodList.Items) == 0 {
+		log.Printf(`❌ No %s pods found\n`, currentPod)
+		return
+	}
+	corePod := &corePodList.Items[0]
```
guymguym: Loop over pods

pkg/util/util.go
```diff
@@ -760,3 +846,66 @@ func DiscoverOAuthEndpoints() (*OAuth2Endpoints, error) {
...
+			return err
+		}
+
+		// copy file data into tar writer
+		if _, err := io.Copy(tw, f); err != nil {
+			return err
+		}
+
+		// manually close here after each file operation; defering would cause each file close
+		// to wait until all operations have completed.
```
guymguym: I don't think this is right, if you defer `f.Close()` inside the walk callback func it should close when the callback returns...

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,159 @@
...
+}
+
+// RunCollect runs a CLI command
+func RunCollect(cmd *cobra.Command, args []string) {
+	c := Config{
+		folderName: fmt.Sprintf("%s_%s", "noobaa_diagnostics", time.Now().Format(time.RFC3339)),
+		log:        util.Logger(),
+	}
+
+	log.Println("Running collection of diagnostics")
```
guymguym: This is not the correct log...

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,159 @@
...
+	c.CollectPodLogs(operatorPodSelector)
+
+	// collectSystemMetrics()
+
+	c.ExportDiagnostics()
+}
+
+// CollectCR info
+func (c *Config) CollectCR(list runtime.Object) {
+	c.log = util.Logger()
```
guymguym: Shouldn't be done here

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,159 @@
...
+	gvk := list.GetObjectKind().GroupVersionKind()
+
+	if !util.KubeList(list, &client.ListOptions{Namespace: options.Namespace}) {
+		log.Printf(`❌ Failed to list %s\n`, gvk.Kind)
```
guymguym: Not the correct log

pkg/util/util.go
```diff
@@ -392,6 +401,70 @@ func GetPodStatusLine(pod *corev1.Pod) string {
...
+	}
+	defer body.Close()
+	f, err := os.Create(path)
+	if err != nil {
+		log.Errorf(`Could not save stream to file: %s, reason: %s\n`, path, err)
+		return 0, err
+	}
+	defer f.Close()
+	io.Copy(f, body)
+	return 0, nil
```
guymguym: Why do you return a number which is always 0?

pkg/diagnose/diagnose.go
```diff
@@ -0,0 +1,159 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"github.com/spf13/cobra"
+)
+
+// Config configuration for diagnostics
+type Config struct {
```
guymguym: Rename to `Collector`

pkg/util/util.go
```diff
@@ -392,6 +401,70 @@ func GetPodStatusLine(pod *corev1.Pod) string {
...
+}
+
+// SaveCRsToFile info
+func SaveCRsToFile(crs runtime.Object, path string) error {
+	if crs == nil {
+		return nil
+	}
+
+	f, err := os.Create(path)
+
```
guymguym: Prefer not to separate the error handling from the line of the error

PR #118, opened 2019-11-03
Add OAuth2 support

nb-ohad:
- Discover OAuth2 endpoints using the .wellknown auth endpint and relay them to noobaa-core via env vars
- Relay noobaa service account to noobaa-core pod via env variable
- CLI: Add missing service account annotaion and cluster role binding (to system:auth-delegator)

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -27,6 +27,13 @@ func (r *Reconciler) ReconcilePhaseCreating() error {
 		"noobaa operator started phase 2/4 - \"Creating\"",
 	)
 
+	// A failiure to discover OAuth endpoints should not fail the entire reconcile phase.
+	oAuthEndpoints, err := util.DiscoverOAuthEndpoints()
+	if err != nil {
+		r.Logger.Warnf("Discovery of OAuth endpoints failed, got: %v", err)
+	}
+	r.OAuthEndpoints = oAuthEndpoints
+
 	// the credentials that are created by cloud-credentials-operator sometimes take time
 	// to be valid (requests sometimes returns InvalidAccessKeyId for 1-2 minutes)
 	// creating the credential request as early as possible to try and avoid it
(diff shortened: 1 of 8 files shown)


guymguym:
@nb-ohad make test fails in CI:
```
# github.com/noobaa/noobaa-operator/v2/pkg/system
pkg/system/phase2_creating.go:33:18: Errorf call has arguments but no formatting directives
```

pkg/system/phase2_creating.go
```diff
@@ -27,6 +27,13 @@ func (r *Reconciler) ReconcilePhaseCreating() error {
 		"noobaa operator started phase 2/4 - \"Creating\"",
 	)
 
+	// A failiure to discover OAuth endpoints should not fail the entire reconcile phase.
+	oAuthEndpoints, err := util.DiscoverOAuthEndpoints()
+	if err != nil {
+		r.Logger.Errorf("Discovery of OAuth endpoints failed, got:", err)
```
guymguym: This is the line of the CI failure - Errorf requires `%s` in the message for the error arg

nb-ohad: Not specifically the whole operation is optional. Many env will not support the .well-known endpoint: Vennila k8s, OpenShift prior to 4

pkg/util/util.go
```diff
@@ -721,3 +733,30 @@ func PrintThisNoteWhenFinishedApplyingAndStartWaitLoop() {
 	log.Printf("  - From now on, it only loops and reads the status, to monitor the operator work.")
 	log.Printf("  - You may Ctrl-C at any time to stop the loop and watch it manually.")
 }
+
+func DiscoverOAuthEndpoints() (*OAuth2Endpoints, error) {
```
guymguym: I see that oauth-proxy has the same logic here:
https://github.com/openshift/oauth-proxy/blob/fbd062bbcb23538863a735f9823ac3fd6f47429e/providers/openshift/provider.go#L543-L567

Do you think we can/should call that code instead?

pkg/util/util.go
```diff
@@ -721,3 +733,30 @@ func PrintThisNoteWhenFinishedApplyingAndStartWaitLoop() {
...
+	client := http.Client{
+		Timeout: 120 * time.Second,
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
```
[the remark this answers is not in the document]
nb-ohad: In order to remove the InsecureSkipVerify option, the operator needs to understand that service-ca.crt on the operator pod is a valid root ca. Which is not currently the case.
We can find a way to do this and add it in a future change

pkg/system/phase2_creating.go
guymguym: Ah that's right.
So can you make sure this http call doesn't hang when running on vanilla kube?

pkg/util/util.go
nb-ohad: We are not using OAuth proxy or its codebase

pkg/util/util.go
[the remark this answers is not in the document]
nb-ohad: I will open an Issue

PR #119, opened 2019-11-05
A doc describing SSL, Name Resolution and Routing as related to NooBaa

nb-ohad:


nb-ohad:
For better readability, while reviewing you can see the doc as formatted text directly at my branch:
https://github.com/nb-ohad/noobaa-operator/blob/ohad-dns-and-ssl-docs/doc/ssl-dns-routing.md

nb-ohad:
@ron1 a review will be very welcomed

ron1:
My apologies for the delayed review. Overall, I think this document will be helpful to folks like me who are trying to work-around the default noobaa-operator reencrypt behavior for OpenShift 4.2+. A few comments:

- I don't think end-user patching of Operator-generated "internal" resources, in this case, Services and Routes, is a good practice. What happens if an upgrade wants to change the "internal" resource in such a way that is incompatible with the end-user patched version? One of the reasons Operator Custom Resources exist is to eliminate the need for this type of patching. Instead, the NooBaa CR itself should capture properties such as "s3->sslCertificateRef", "s3->port", "s3->securePort", and maybe "s3->ingressTerminationType", "s3->hostname". If this additional information were available, the noobaa-operator could do the right thing and eliminate any need for end-user, s3-related endpoint patching. In fact, robust operator reconciliation logic ensures that resources managed by the operator have the state the operator expects. Any end-user modifications to that state would be immediately reverted. This is clearly not the behavior described in this document.

- The document seems a bit OpenShift-centric. I hope this upstream project continues to target the broader Kubernetes community in addition to Openshift. Unfortunately, Kubernetes Ingress configuration is a complex, not well standardized feature which makes it hard to manage as an "internal" resource. Many operator CRs, such as the Rook CephObjectStore, simply treat this as an "external" issue and rely upon the end-user to create her own Route/Ingress resource as needed. This would work for us.

- In 2.0.6, we have been unable to successfully configure the noobaa-mgmt Route with ssl termination type "passthrough". I suspect the forthcoming disableSSO flag may needed in order for this to work.

nb-ohad:
@ron1 Regarding your comments:
- We are in full agreement. The instructions laid out in the document are meant as a workaround until we get to implementing support in the NooBaa CRD as you suggested.

- We work under the assumption that networking and name resolution in the border Kubernetes world is user a responsibility. Saying that we still explain how to set up certificates in that case. If it was not obvious from the document please let me know and I will update the document.

- Did you tested it on COP3+ or OCP4+. SSO should work on COP4.2+ even if you change the certificate (and use "passthrough"), there is no connection between the two. In the case of OCP3 SSO is not supported at all, but I guess our detection code thinks it should support it and try to configure it which will cause problems even with the default certificates. In the meantime, as a workaround, you can add "?skip-oauth" to the end of the mgmt route to skip the SSO login. We will fix the detection code and add the disableSSO flag in future PRs

ron1:
@nb-ohad My feedback is included below:

- The doc was so thorough I was thinking it might not reflect a work-around solution. Would it make sense to create an issue for the NooBaa CRD changes and reference that issue from the doc?

- Agreed. So NooBaa overachieves for OpenShift since Route usage is simple and pervasive on OCP. The broader Kubernetes world must manage Ingress itself.

- I believe oAuth support in OCP3 and OCP4 is very similar. I am testing with OCP3. Do I add the "?skip-oauth" to the Route Host value itself?

nb-ohad:
@ron1 Thanks for the feedback on the document.

Regarding the doc, it is meant to be a description of the networking solution NooBaa is applying in various environments (with an emphasis on OpenShift). 
When we will implement a proper way to accomplish the issue on hand I will remove all the how-to sections from the document and update it appropriately.
The thing is that we still find value in describing the way NooBaa is configuring network and certificate usage in the repository (via the doc).

Regarding OAuth (SSO), it is similar in OCP3 but not identical to OCP4). 
We should decide when and what do to, regarding the issue) taking into account other workloads we have. For the short term my recommendation is to drop SSO support from OCP3, please let me know if and how it will affect you.

Creating an issue is a great way to track any progress on the matter and I encourage you to do so for any missing/misbehaving feature. In most cases, it will help to start a discussion. 

Last but not least, the workaround. 
Let's say your mgmt address is `noobaa-mgmt-ron.apps.domain`, you can skip SSo and go directly to NooBaa login by using the URL: `https://noobaa-mgmt-ron.apps.domain?skip-oauth`
I hope this helps

PR #120, opened 2019-11-06
Fixing deletion of BackingStore from NooBaa

jeniawhite:


guymguym:
Created an issue #121

Issue #91, opened 2019-10-16
Implement BackingStore type pv-pool in Operator CR & CLI


ron1:
According to the noobaa-operator reconciliation code here: https://github.com/noobaa/noobaa-operator/blob/be83b8fe0467bded6c5ee9dea59853808cc01170/pkg/backingstore/reconciler.go#L512, the noobaa-operator does not yet support BackingStore type "pv-pool". This is a blocker for many folks that want a performant NooBaa deployment on Kubernetes using their own storage.

At a minimum, the BackingStore documentation available here: https://github.com/noobaa/noobaa-operator/blob/master/doc/backing-store-crd.md#pvc-type should be updated to reflect that this feature has not yet been implemented. In addition, consider elevating the priority of this issue in the NooBaa backlog.

tamireran:
@ron1 You are correct. We will add it to the CLI soon. In the meantime, you can use the web management console from your browser. The access details can be found with the command 
`noobaa status` or `kubectl describe noobaa`
Once you are in, you will see a button "Add Storage Resources". Click on it, and click on "Deploy Kubernetes Pool" in the new window. You will see a little wizard that will get you a local storage on local PV. 
Please let us know if you have any issues.

ron1:
@tamireran Thanks for the clarification. I did not realize NooBaa Core supported this feature and that the limitation is just in the Operator and CLI.

Am I correct that pv-pools work with all Kubernetes/OpenShift dynamically-provisioned volumes and not just Kubernetes 1.14 GA Local Persistent Volumes? Based on the reference to "local storage on local PV" in your comment, this is not quite clear.

tamireran:
@ron1, correct. We support Kubernetes/OpenShift dynamically-provisioned
volumes

ron1:
@tamireran @guymguym Unfortunately, pvpool/pvstore type storage resources created via the "Advanced pvstore create" CLI or through the Web UI cannot be referenced from BucketClass CRs, the "Manage bucketclass" CLI, and therefore from ObjectBucketClaim CRs and the "Manage obc" CLI. This essentially makes them unusable unless I am missing something. If this is the case, please elevate the priority of this issue. A patch should be fairly easy to implement since the code needed to create a pvpool BackingStore has already been implemented for the "Advanced pvstore create" CLI.

guymguym:
@ron1 I know unfortunately. That's a hard gap. I'll try to push it.

ron1:
[...]

PR #146, opened 2019-11-17
Feature #91 Enable backingstore type pv-pool

ron1:


ron1:
Could someone take a look at the failed Travis CI Build and let me know if this is a build problem or a problem with my patch? Thx.

guymguym:
@ron1 the CI fails on lint:
```
pkg/backingstore/reconciler.go:522:15: Sprintf call has arguments but no formatting directives
```

ron1:
@guymguym Yep, I missed that. Thx.

pkg/backingstore/reconciler.go
```diff
@@ -350,26 +350,45 @@ func (r *Reconciler) ReadSystemInfo() error {
...
+		if p.Name == r.BackingStore.Name {
+			r.PoolInfo = p
+			break
+		}
+	}
+
+	pool := r.PoolInfo
+	if r.BackingStore.Spec.Type == nbv1.StoreTypePVPool {
+		if pool != nil && pool.ResourceType != "HOSTS" {
+			r.Logger.Warnf("BackingStore %q w/existing pool %+v has unexpected resource type %+v", r.BackingStore.Name, pool, pool.ResourceType)
```
guymguym: **[IMPORTANT]**
That's a good thing to check but I think we want to take a bit more drastic measures.
I suggest to return a persistent error so that this backing store will get to a rejected state.
Also this should probably be checked in the same manner for the opposite situation - that is when `r.BackingStore.Spec.Type != nbv1.StoreTypePVPool && pool != nil && pool.ResourceType != "CLOUD"`

pkg/backingstore/reconciler.go
```diff
@@ -546,6 +565,9 @@ func (r *Reconciler) ReconcileExternalConnection() error {
 		return nil
 	}
 
+	if r.AddExternalConnectionParams == nil {
```
guymguym: **[IMPORTANT]**
I'm trying to figure out why this is needed, can you explain?
Perhaps best as a code comment...

guymguym:
Hey @ron1 
I would like to add this to next release early next week.
Do you think you can go over the comments?
Thanks

guymguym:
@ron1 I can do fixes to my comments in another PR after we merge this.
Can you just rebase your PR so I can merge it?
Thanks

pkg/backingstore/reconciler.go
ron1: If I remember correctly, I believe I added this to avoid a subsequent error on line 615.

ron1:
Updated PR based on feedback.

guymguym:
CI fails that go.sum changed during build.
This is probably because the go version we use in TravisCI is 1.13 but you used 1.12?
In general I think we need to fix these two to match:

https://github.com/noobaa/noobaa-operator/blob/1b2a1ca6134c30299cd0d899c128db2294aef1aa/.travis.yml#L7-L8

https://github.com/noobaa/noobaa-operator/blob/1b2a1ca6134c30299cd0d899c128db2294aef1aa/go.mod#L3

I will take care of it in a separate PR, so I am ignoring the build failure for now.

PR #159, opened 2019-11-21
Remove OwnerReferences on OBCStorageClass

kshlm:
Invalid OwnerReferences can lead to the other children of the owner
being garbage collected. Ref [1][1].

The OBCStorageClass is a non-namespaced (global) resource, which has
OwnerReference set to the NooBaa resource which is a namespaced
resource. This is an invalid OwnerReference.

[1]: https://bugzilla.redhat.com/show_bug.cgi?id=1691546


guymguym:
Thanks @kshlm !

PR #162, opened 2019-11-22
MetaData for creation of Operand form for BackingStore and BucketClass

swati-kale:
Signed-off-by: Swati Kale <swkale@redhat.com>


pkg/olm/olm.go
```diff
@@ -205,6 +205,18 @@ func GenerateCSV(opConf *operator.Conf) *operv1.ClusterServiceVersion {
 	}
 	uiText := []string{"urn:alm:descriptor:com.tectonic.ui:text"}
 	uiResources := []string{"urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
+	uiFieldGroup := "urn:alm:descriptor:com.tectonic.ui:fieldGroup:"
+	uiK8sSecret := "urn:alm:descriptor:io.kubernetes:Secret"
+	uiBooleanSwitch := "urn:alm:descriptor:com.tectonic.ui:booleanSwitch"
+	uiNumber := "urn:alm:descriptor:com.tectonic.ui:number"
+	uiFieldGroupText := "urn:alm:descriptor:com.tectonic.ui:text"
```
guymguym: This is not a fieldGroup right? 
Lets rename it to uiText and since few lines before we defined uiText but in that case it was as an array, then I suggest to replace the other uses of uiText to `[]string{uiText}`.
```suggestion
	const uiText = uiTectonic + "text"
```

pkg/olm/olm.go
```diff
@@ -205,6 +205,18 @@ func GenerateCSV(opConf *operator.Conf) *operv1.ClusterServiceVersion {
...
+	uiFieldGroupAwsS3 := "urn:alm:descriptor:com.tectonic.ui:fieldGroup:awsS3"
```
guymguym:
```suggestion
	const uiFieldGroupAwsS3 = uiFieldGroup + "awsS3"
```

pkg/olm/olm.go
```diff
@@ -205,6 +205,18 @@ func GenerateCSV(opConf *operator.Conf) *operv1.ClusterServiceVersion {
 	}
 	uiText := []string{"urn:alm:descriptor:com.tectonic.ui:text"}
 	uiResources := []string{"urn:alm:descriptor:com.tectonic.ui:resourceRequirements"}
+	uiFieldGroup := "urn:alm:descriptor:com.tectonic.ui:fieldGroup:"
```
guymguym:
```suggestion
	const uiTectonic = "urn:alm:descriptor:com.tectonic.ui:"
	const uiFieldGroup = uiTectonic + "fieldGroup:"
```

pkg/olm/olm.go
```diff
@@ -203,30 +203,117 @@ func GenerateCSV(opConf *operator.Conf) *operv1.ClusterServiceVersion {
...
 		"ObjectBucketClaim": []operv1.SpecDescriptor{},
 		"ObjectBucket":      []operv1.SpecDescriptor{},
 	}
 	for kind := range crdSpecDescriptors {
 		for i := range crdSpecDescriptors[kind] {
 			d := &crdSpecDescriptors[kind][i]
-			d.DisplayName = d.Path
-			d.Description = d.Path
+			d.DisplayName = d.DisplayName
+			d.Description = d.Description
```
guymguym: The reason I wrote this loop originally was that I was too lazy to actually fill `Description` and `DisplayName` for the fields, so I just copied `Path` to both. However this is now the case only for the `NooBaa` CRD and not for `BackingStore` and `BucketClass` which you added, and this assignment is not even doing anything as it is now.

So I just suggest to remove this loop altogether and instead fill those Description and DisplayName for the NooBaa CRD.

swati-kale: Updated the code.

PR #163, opened 2019-11-27
Separate Endpoints into a standalone deployment

nb-ohad:
- Remove endpoint deployment and reconcile for NooBaa core sts
- Open mongo to external connections
- Create an endpoint deployment yaml
- Add endpoints control and status into noobaa CRD
- Change the way we handle resource limit/requets for the NooBaa pods
  * Replace all default values with values decided for base profile.
  * Expose endpoint pod resource configuration inside the endpoint part of the NooBaa crd
  * Changed test-olm to use a low profile for endpoint resources
- Fix test to support new endpoint configuration


pkg/system/phase4_configuring.go
```diff
@@ -51,10 +57,110 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
...
+			for j := range c.Env {
+				switch c.Env[j].Name {
+				case "MGMT_URL":
+					c.Env[j].Value = fmt.Sprintf(`https://%s.%s.svc`,
+						r.ServiceMgmt.Name, r.Request.Namespace)
+
+				case "MONGODB_URL":
+					// TODO: change to mongodb pod name after mongo separation
+					c.Env[j].Value = fmt.Sprintf(`mongodb://%s-0.%s/nbcore`,
+						r.CoreApp.Name, r.ServiceMgmt.Name)
```
dannyzaken: I think the pod dns name should be (r.NooBaaDB.Name)-0.(r.NooBaaDB.Spec.ServiceName)

nb-ohad: Changed the code as of your suggestion

test/cli/test_cli_functions.sh
```diff
@@ -100,14 +100,16 @@ function timeout {
 
 function install {
 # TODO: once we can control the core resource amount we can use it,
-#       for now we will mimic the install command 
+#       for now we will mimic the install command
     noobaa crd create
     noobaa system create
```
guymguym: We already support --core-resources and --db-resources and --endpoint-resources as flags on `noobaa system create` so we can remove those kubectl patches

guymguym: @nb-ohad just like you did for test-olm.sh

nb-ohad: Changed the test to use the new --mini flag in #183

PR #172, opened 2019-12-11
Addition of bucketclass edit reconciler

### LINKED ISSUE #59: Updating existing BucketClass does not update policies for existing OBCs
Updating a BucketClass which was already used by existing OBCs does not update current buckets to the new policy.

jeniawhite:
bucketclass reconciler will update the BE with the edits of the bucketclasses and change the relevant tiering policies and tiers

Fixes #59

### DIFF
--- a/pkg/bucketclass/reconciler.go
+++ b/pkg/bucketclass/reconciler.go
@@ -256,6 +256,7 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 	}
 	r.SystemInfo = &systemInfo
 
+	r.UpdateBucketClass()
 	// TODO ReconcileTieringPolicy
 	// for i := range r.SystemInfo.Buckets {
 	// 	bucket := &r.SystemInfo.Buckets[i]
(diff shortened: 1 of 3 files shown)


pkg/bucketclass/reconciler.go
```diff
@@ -297,3 +298,50 @@ func (r *Reconciler) FinalizeDeletion() error {
...
+		return fmt.Errorf("BucketClass not loaded %#v", r)
+	}
+
+	policyTiers := []nb.TierItem{}
+	tiers := []nb.TierInfo{}
+
+	for i := range r.BucketClass.Spec.PlacementPolicy.Tiers {
+		tier := &r.BucketClass.Spec.PlacementPolicy.Tiers[i]
+		// Tier is irrelevant and will be populated in the BE
+		policyTiers = append(policyTiers, nb.TierItem{Order: int64(i), Tier: "SLOTH"})
```
nimrod-becker: Tier SLOTH ? :)

pkg/bucketclass/reconciler.go
```diff
@@ -297,3 +298,50 @@ func (r *Reconciler) FinalizeDeletion() error {
...
+		// we assume either mirror or spread but no mix and the bucket class controller rejects mixed classes.
+		placement := "SPREAD"
+		if tier.Placement == nbv1.TierPlacementMirror {
+			placement = "MIRROR"
+		}
+		// Name is irrelevant and will be populated in the BE
+		tiers = append(tiers, nb.TierInfo{Name: "SLOTH", AttachedPools: tier.BackingStores, DataPlacement: placement})
+	}
+
+	err := r.NBClient.UpdateBucketClass(nb.UpdateBucketClassParams{
```
dannyzaken: maybe check if update is needed (according to system info) before actually updating noobaa for every reconcile

pkg/bucketclass/reconciler.go
jeniawhite: Any suggestions for a different name? It is irrelevant since I populate that in the BE :)

nimrod-becker: TEMP ?

pkg/bucketclass/reconciler.go
jeniawhite: Opened an issue #173

pkg/bucketclass/reconciler.go
```diff
@@ -297,3 +298,76 @@ func (r *Reconciler) FinalizeDeletion() error {
...
+							r.BucketClass.Spec.PlacementPolicy.Tiers = append(r.BucketClass.Spec.PlacementPolicy.Tiers,
+								nbv1.Tier{Placement: placement, BackingStores: t.AttachedPools})
+							break
+						}
+					}
+				}
+				break
+			}
+		}
+		util.KubeUpdate(r.BucketClass)
```
guymguym: Missing event in this case of overriding the bucket class by the operator...

pkg/bucketclass/reconciler.go
```diff
@@ -297,3 +298,76 @@ func (r *Reconciler) FinalizeDeletion() error {
...
+		// we assume either mirror or spread but no mix and the bucket class controller rejects mixed classes.
+		placement := "SPREAD"
+		if tier.Placement == nbv1.TierPlacementMirror {
+			placement = "MIRROR"
+		}
+		// Name is irrelevant and will be populated in the BE
+		tiers = append(tiers, nb.TierInfo{Name: "TEMP", AttachedPools: tier.BackingStores, DataPlacement: placement})
+	}
+
+	err := r.NBClient.UpdateBucketClass(nb.UpdateBucketClassParams{
```
guymguym: I suggest that this call will return back the applied bucket class by the core, so that the operator will update the CR if needed.

PR #177, opened 2019-12-16
Trim the prerelease part from the version

nb-ohad:
By default, version constraint checks will fail on the prerelease version if the constraint does not exactly match the version. Trimming the prerelease part solves this problem

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,12 +5,12 @@ go 1.13
 require (
 	github.com/aws/aws-sdk-go v1.23.8
 	github.com/blang/semver v3.5.1+incompatible
+	github.com/coreos/go-semver v0.2.0
 	github.com/coreos/prometheus-operator v0.29.0
 	github.com/docker/distribution v2.7.1+incompatible
 	github.com/go-openapi/spec v0.19.2
 	github.com/gobuffalo/flect v0.1.6 // indirect
 	github.com/google/uuid v1.1.1
-	github.com/hashicorp/go-version v1.2.0
 	github.com/kube-object-storage/lib-bucket-provisioner v0.0.0-20190924175516-f3ba69cc601e
 	github.com/openshift/api v3.9.1-0.20190424152011-77b8897ec79a+incompatible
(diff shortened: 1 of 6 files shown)


guymguym:
See this - https://goplay.space/#YBHyEW_Tlo_r

nimrod-becker:
@guymguym + doesn't work in docker, there is an open issue on it

nb-ohad:
@guymguym I checked the code for version.go and constraint.go. 
If a prerelease is present on a version (a part that starts with "-" not "+") all constraints fall back to exact match. you can see this behavior in the prereleaseCheck function (it uses deep equal on the numeric version parts).
A part that starts with a "+" is not considered a prerelease and this is why the code you linked works.
But as nimrod already mentioned it there is an open issue for tags containing the "+" sign.

The best curse of action, in this case, is to cut out everything beginning with the first "-" sign and used the trimmed version only for constraint checks. In any other place, we use the full version name.

guymguym:
OK so I suggest to change to this semver library that supports prerelease:
https://goplay.space/#9OnXIxyaJkt
```golang
package main

import (
	"fmt"

	semver "github.com/coreos/go-semver/semver"
)

func main() {

	minver := semver.New("5.0.0")
	maxver := semver.New("6.0.0")
	fmt.Printf("min = %#v\n", minver)
	fmt.Printf("max = %#v\n", maxver)

	version := semver.New("5.3.11-master20191215")
	fmt.Printf("version = %#v\n", version)

	mincompare := version.Compare(*minver)
	maxcompare := version.Compare(*maxver)
	fmt.Printf("mincompare = %#v\n", mincompare)
	fmt.Printf("maxcompare = %#v\n", maxcompare)
}
```

output:
```
min = &semver.Version{Major:5, Minor:0, Patch:0, PreRelease:"", Metadata:""}
max = &semver.Version{Major:6, Minor:0, Patch:0, PreRelease:"", Metadata:""}
version = &semver.Version{Major:5, Minor:3, Patch:11, PreRelease:"master20191215", Metadata:""}
mincompare = 1
maxcompare = -1
```

PR #179, opened 2019-12-17
Fixing BZ issues - 1781287 and 1779732

jackyalbo:
Fixes for the following issues, some of them not fully:
https://bugzilla.redhat.com/show_bug.cgi?id=1781287 - added handling of Invalid names to KubeCreateSkipExisting instead of just panicing 
https://bugzilla.redhat.com/show_bug.cgi?id=1779732 - added Delete support for pvstore as well as handling special cased of pool being used by buckets/account
https://bugzilla.redhat.com/show_bug.cgi?id=1777295 - added to the CLI special handling of pools being used by buckets/account - it will still hang on deleting directly from kubernetes api, 
we need to think what to do in order to provide better feedback for that kind of failed deletions

### DIFF
--- a/pkg/backingstore/backingstore.go
+++ b/pkg/backingstore/backingstore.go
@@ -8,7 +8,9 @@ import (
 
 	nbv1 "github.com/noobaa/noobaa-operator/v2/pkg/apis/noobaa/v1alpha1"
 	"github.com/noobaa/noobaa-operator/v2/pkg/bundle"
+	"github.com/noobaa/noobaa-operator/v2/pkg/nb"
 	"github.com/noobaa/noobaa-operator/v2/pkg/options"
+	"github.com/noobaa/noobaa-operator/v2/pkg/system"
 	"github.com/noobaa/noobaa-operator/v2/pkg/util"
 
 	"github.com/spf13/cobra"
@@ -300,10 +302,10 @@ func RunCreateAWSS3(cmd *cobra.Command, args []string) {
 // RunCreateS3Compatible runs a CLI command
 func RunCreateS3Compatible(cmd *cobra.Command, args []string) {
 	createCommon(cmd, args, nbv1.StoreTypeS3Compatible, func(backStore *nbv1.BackingStore, secret *corev1.Secret) {
+		endpoint := util.GetFlagStringOrPrompt(cmd, "endpoint")
 		targetBucket := util.GetFlagStringOrPrompt(cmd, "target-bucket")
 		accessKey := util.GetFlagStringOrPromptPassword(cmd, "access-key")
 		secretKey := util.GetFlagStringOrPromptPassword(cmd, "secret-key")
-		endpoint, _ := cmd.Flags().GetString("endpoint")
 		sigVer, _ := cmd.Flags().GetString("signature-version")
 		secret.StringData["AWS_ACCESS_KEY_ID"] = accessKey
 		secret.StringData["AWS_SECRET_ACCESS_KEY"] = secretKey
@@ -427,6 +429,30 @@ func RunDelete(cmd *cobra.Command, args []string) {
 	backStore.Namespace = options.Namespace
 	backStore.Spec = nbv1.BackingStoreSpec{}
 
+	nbClient := system.GetNBClient()
+
+	poolinfo, err := nbClient.ReadPoolAPI(nb.ReadPoolParams{Name: backStore.Name})
+	if err != nil {
+		rpcErr, isRPCErr := err.(*nb.RPCError)
+		if !isRPCErr || rpcErr.RPCCode != "NO_SUCH_POOL" {
+			log.Fatalf(`❌ Failed to read BackingStore info: %s`, err)
+		}
+	} else if poolinfo.Undeletable != "" {
+		switch poolinfo.Undeletable {
+		case "CONNECTED_BUCKET_DELETING":
+			fallthrough
+		case "IN_USE":
+			log.Fatalf(`❌ Could not delete BackingStore %q in namespace %q as it is being used by one or more bucket classes`,
+				backStore.Name, backStore.Namespace)
+
+		case "DEFAULT_RESOURCE":
+			log.Fatalf(`❌ Could not delete BackingStore %q in namespace %q as it is the default resource of one or more accounts`,
+				backStore.Name, backStore.Namespace)
+		default:
+			log.Fatalf(`❌ Could not delete BackingStore %q in namespace %q, undeletable due to %q`,
+				backStore.Name, backStore.Namespace, poolinfo.Undeletable)
+		}
+	}
 	if !util.KubeDelete(backStore) {
 		log.Fatalf(`❌ Could not delete BackingStore %q in namespace %q`,
 			backStore.Name, backStore.Namespace)
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -303,6 +303,16 @@ func (r *Reconciler) ReconcileDeletion() error {
 		}
 		err := r.NBClient.DeletePoolAPI(nb.DeletePoolParams{Name: r.PoolInfo.Name})
 		if err != nil {
+			if rpcErr, isRPCErr := err.(*nb.RPCError); isRPCErr {
+				if rpcErr.RPCCode == "DEFAULT_RESOURCE" {
+					return util.NewPersistentError("DefaultResource",
+						fmt.Sprintf("DeletePoolAPI cannot complete because pool %q is an account default resource", r.PoolInfo.Name))
+				}
+				if rpcErr.RPCCode == "IN_USE" {
+					return util.NewPersistentError("ResourceInUse",
+						fmt.Sprintf("DeletePoolAPI cannot complete because pool %q has buckets attached", r.PoolInfo.Name))
+				}
+			}
 			return err
 		}
 	}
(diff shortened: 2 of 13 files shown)


pkg/pvstore/pvstore.go
```diff
@@ -101,6 +112,34 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
+	}
+
+	nbClient := system.GetNBClient()
+
+	poolName := args[0]
+
+	err := nbClient.DeletePoolAPI(nb.DeletePoolParams{
+		Name: poolName,
+	})
+	if err != nil {
```
nimrod-becker: What if pv-pool doesn't exist?

jackyalbo: then it will panic... we prefer something else? I guess it is pretty bad if to happens

guymguym: Probably best to just continue if the pool doesn't exist, because it means the deletion is essentially successful

pkg/util/util.go
```diff
@@ -207,6 +207,10 @@ func KubeCreateSkipExisting(obj runtime.Object) bool {
 		log.Printf("❌ Forbidden: %s %q: %s\n", gvk.Kind, objKey.Name, err)
 		return false
 	}
+	if errors.IsInvalid(err) {
```
guymguym: What does Invalid means - that the validations of the object failed, like missing a required field for example?

In which case did you get it? 
Is this is a static object that we create in the code? do we generally expect validation errors?

pkg/backingstore/reconciler.go
```diff
@@ -303,7 +303,19 @@ func (r *Reconciler) ReconcileDeletion() error {
...
 		if err != nil {
-			return err
+			if rpcErr, isRPCErr := err.(*nb.RPCError); isRPCErr {
+				if rpcErr.RPCCode == "DEFAULT_RESOURCE" {
+					return util.NewPersistentError("DefaultResource",
+						fmt.Sprintf("DeletePoolAPI cannot complete because pool %q is an account default resource", r.PoolInfo.Name))
+				} else if rpcErr.RPCCode == "IN_USE" {
+					return util.NewPersistentError("ResourceInUse",
+						fmt.Sprintf("DeletePoolAPI cannot complete because pool %q has buckets attached", r.PoolInfo.Name))
+				} else {
```
guymguym: How about just removing these two `else` statements and just keep it as a fallback to `return err` in any other case?

pkg/util/util.go
jackyalbo: we got it when the name of the bucket is illeagal: ```PANI[0004] ☠️  Panic Attack: [Invalid] ObjectBucketClaim.objectbucket.io "obc#1" is invalid: metadata.name: Invalid value: "obc#1"```

pkg/backingstore/backingstore.go
```diff
@@ -420,7 +422,18 @@ func RunDelete(cmd *cobra.Command, args []string) {
...
+	poolinfo, err := nbClient.ReadPoolAPI(nb.ReadPoolParams{Name: backStore.Name})
+	if err != nil {
+		log.Fatalf("❌ Failed to read BackingStore info: %s", err)
+	} else if poolinfo.Undeletable == "IN_USE" {
+		log.Fatalf(`❌ Could not delete BackingStore %q in namespace %q as it is being used by one or more buckets`,
+			backStore.Name, backStore.Namespace)
+	} else if poolinfo.Undeletable == "DEFAULT_RESOURCE" {
+		log.Fatalf(`❌ Could not delete BackingStore %q in namespace %q as it is the default resource of one or more accounts`,
+			backStore.Name, backStore.Namespace)
+	} else if !util.KubeDelete(backStore) {
```
guymguym: I would avoid `else if` here - it's hard to follow the flow when the delete call is hidden inside a compound if statement.

pkg/backingstore/backingstore.go
```diff
@@ -420,7 +422,18 @@ func RunDelete(cmd *cobra.Command, args []string) {
...
 	backStore.Spec = nbv1.BackingStoreSpec{}
 
-	if !util.KubeDelete(backStore) {
+	nbClient := system.GetNBClient()
+
+	poolinfo, err := nbClient.ReadPoolAPI(nb.ReadPoolParams{Name: backStore.Name})
+	if err != nil {
+		log.Fatalf("❌ Failed to read BackingStore info: %s", err)
+	} else if poolinfo.Undeletable == "IN_USE" {
+		log.Fatalf(`❌ Could not delete BackingStore %q in namespace %q as it is being used by one or more buckets`,
```
nimrod-becker:
```suggestion
		log.Fatalf(`❌ Could not delete BackingStore %q in namespace %q as it is being used by one or more buckets classes`,
```

pkg/backingstore/reconciler.go
jackyalbo: changed to switch

jackyalbo:
> Maybe worth adding the flows you fixed/added to test_cli_flow? This way we'd have regression testing on it

I will discuss with @liranmauda as the whole deletion part of the flow is marked out.

pkg/backingstore/backingstore.go
jackyalbo: changed to switch

pkg/backingstore/reconciler.go
jackyalbo: removed the else as recommended

pkg/pvstore/pvstore.go
jackyalbo: added handling for doesn't exist - so all these cases will have same kind of responses on failures

pkg/backingstore/backingstore.go
```diff
@@ -427,7 +429,28 @@ func RunDelete(cmd *cobra.Command, args []string) {
 	backStore.Namespace = options.Namespace
 	backStore.Spec = nbv1.BackingStoreSpec{}
 
-	if !util.KubeDelete(backStore) {
+	nbClient := system.GetNBClient()
+
+	continueDelete := true
+	poolinfo, err := nbClient.ReadPoolAPI(nb.ReadPoolParams{Name: backStore.Name})
+	if err != nil {
+		if rpcErr, isRPCErr := err.(*nb.RPCError); !isRPCErr || rpcErr.RPCCode != "NO_SUCH_POOL" {
```
guymguym: I think this statement is too long for a one liner...
You can break it:
```golang
	rpcErr, isRPCErr := err.(*nb.RPCError)
	if !isRPCErr || rpcErr.RPCCode != "NO_SUCH_POOL" {
```

pkg/backingstore/backingstore.go
```diff
@@ -427,7 +429,28 @@ func RunDelete(cmd *cobra.Command, args []string) {
...
+			log.Fatalf(`❌ Failed to read BackingStore info: %s`, err)
+			continueDelete = false
```
guymguym: log.Fatalf() exits the process after logging, so no need for the code after it, so continueDelete is unneeded

pkg/backingstore/backingstore.go
```diff
@@ -427,7 +429,28 @@ func RunDelete(cmd *cobra.Command, args []string) {
...
+		}
+	} else {
+		switch poolinfo.Undeletable {
```
guymguym: I think we need to fail in any case of `poolinfo.Undeletable != ""`

test/cli/test_cli_functions.sh
```diff
@@ -26,7 +26,13 @@ function kuberun {
 function noobaa {
     local rc
     local run_with_timeout=false
+    local should_fail=false
     local silence=false
+    if [ "${1}" == "failure" ]
```
guymguym: I think this is becoming very confusing that we have an override of the noobaa command that has all these test options that do not exist in the CLI. At least I would rename it to test_noobaa or something.

pkg/options/options.go
```diff
@@ -30,7 +30,7 @@ const (
 	// ContainerImageRepo is the repo of the default image url
 	ContainerImageRepo = "noobaa-core"
 	// ContainerImageTag is the tag of the default image url
-	ContainerImageTag = "5.3.0-master20191219"
+	ContainerImageTag = "5.3.0-master20191223"
```
guymguym: Just making sure - is this a good image for anyone using master?

pkg/backingstore/backingstore.go
jackyalbo: also on deleting? are we sure it will always be permanent error? shouldn't we reconcile on that?

pkg/options/options.go
jackyalbo: suppose to fix some annoying issues: https://github.com/noobaa/noobaa-core/pull/5851

test/cli/test_cli_functions.sh
jackyalbo: changed to test_noobaa

pkg/backingstore/backingstore.go
jackyalbo: failing on all undeletables as bucket_deleting usually suggests a real issue in the core

pkg/backingstore/backingstore.go
jackyalbo: fixed

pkg/backingstore/backingstore.go
jackyalbo: fixed

PR #188, opened 2019-12-24
Add Web Sockets communication

romayalon:
**Explain the changes**
1. Added RPC struct and Initialised GlobalRPC, Added newRPC() function.
2. Added additional structs: RPCConn, RPCPendingRequest, RPCMessagePreread etc.
3. Added WS support in Call function.
4. Added register for alerts function and called it in phase4_configuring.go file.
5. Added rpc_benchmark.

**Issues: Fixed #xxx / Gap #xxx**
1.

**Testing Instructions:**
1.


pkg/nb/rpc.go
```diff
@@ -2,28 +2,56 @@ package nb
...
+var GlobalRPC *RPC
+
+func init() {
+	// GlobalRPC initialization
+	GlobalRPC = NewRPC()
+}
+
+// RPC is a struct that describes the relevant fields upon handeling rpc protocol
+type RPC struct {
+	Protocol    string
```
guymguym: I think Protocol should be a property of the router, but do we even need it now?

pkg/nb/router.go
```diff
@@ -22,7 +22,7 @@ import (
 // APIRouter should be able to map noobaa api names to actual addresses
 // See implementations below: APIRouterNodePort, APIRouterPodPort, APIRouterServicePort
 type APIRouter interface {
-	GetAddress(api string) string
+	GetAddress(api string, protocol string) string
```
guymguym: I think protocol should be determined by the router, rather than provided to it.

pkg/nb/rpc.go
```diff
@@ -89,52 +177,298 @@ func (c *RPCClient) Call(req *RPCRequest, res RPCResponseIfc) error {
...
+
+	address := c.Router.GetAddress(api, c.RPC.Protocol)
+
 	// u := address + strings.TrimSuffix(api, "_api") + "/" + method
 	u := strings.TrimSuffix(api, "_api") + "." + method + "()"
 	logrus.Infof("✈️  RPC: %s Request: %+v", u, req.Params)
 
-	reqBytes, err := json.Marshal(req)
-	fatal(err)
+	switch c.RPC.Protocol {
```
guymguym: use `u.Scheme` instead

pkg/system/phase4_configuring.go
```diff
@@ -36,6 +36,9 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 	if err := r.ReconcileSecretAdmin(); err != nil {
 		return err
 	}
+	if err := r.NBClient.RegisterToCluster(); err != nil {
```
guymguym: For now I would comment this call, because we don't know if this behaves nicely with many notifications and if this doesn't trigger too many reconciles...

pkg/nb/rpc_benchmark_ws.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+var client = flag.Bool("client", false, "client option")
+var port = flag.String("port", "5656", "port option")
+var hostname = flag.String("hostname", "127.0.0.1", "hostname option")
+var wsize = flag.Int64("wsize", 1024*1024, "wsize option")
+var rsize = flag.Int64("rsize", 1024*1024, "rsize option")
+var concur = flag.Int("concur", 16, "concur option")
+
+func benchmark() {
```
guymguym: This shouldn't be in the nb package because it would pollute the global flags.
Move this to its own package main under `test/rpcbench/`
Also add a comment on the function how to use it.

pkg/nb/rpc.go
```diff
@@ -89,52 +177,298 @@ func (c *RPCClient) Call(req *RPCRequest, res RPCResponseIfc) error {
...
+				b.Buffer = buffers[pos:end]
+				pos = end
+			}
+		}
+
+		switch msg.Op {
+		case "req":
+
+			if r.Handler != nil {
+				go func() {
```
guymguym: extract this to function:
```
func (r *RPC) HandleRequest(msg * RPCMessagePreread) { ... }
```

pkg/nb/rpc.go
```diff
@@ -122,19 +205,251 @@ func (c *RPCClient) Call(req *RPCRequest, res RPCResponseIfc) error {
...
+		req.AuthToken = c.AuthToken
+	}
+
+	address := c.Router.GetAddress(api)
+
+	// u := address + strings.TrimSuffix(api, "_api") + "/" + method
+	u := strings.TrimSuffix(api, "_api") + "." + method + "()"
+	logrus.Infof("✈️  RPC: %s Request: %+v", u, req.Params)
+
+	if strings.HasPrefix(address, "wss") {
```
guymguym:
```suggestion
	if strings.HasPrefix(address, "wss:") {
```

pkg/nb/rpc.go
```diff
@@ -122,19 +205,251 @@ func (c *RPCClient) Call(req *RPCRequest, res RPCResponseIfc) error {
...
+	}
+	return nil
+}
+
+// GetConnection finds the connection related to the pending request or creates a new one
+func (r *RPC) GetConnection(pending *RPCPendingRequest, address string) error {
+	r.ConnMapLock.Lock()
+	pending.Conn = r.ConnMap[address]
+	if pending.Conn == nil {
+		ws, _, err := websocket.Dial(context.TODO(), address, &websocket.DialOptions{HTTPClient: &r.HTTPClient})
```
guymguym: Use `r.ConnectWS()` here?

pkg/nb/rpc.go
```diff
@@ -122,19 +205,251 @@ func (c *RPCClient) Call(req *RPCRequest, res RPCResponseIfc) error {
...
+	}
+	if err := binary.Read(reader, binary.BigEndian, &bodySize); err != nil {
+		return nil, err
+	}
+	msgBytes := make([]byte, bodySize)
+	n, err := reader.Read(msgBytes)
+	if err != nil {
+		if websocket.CloseStatus(err) == websocket.StatusNormalClosure {
+			fmt.Printf("websocket closed, err: %+v\n", err)
+			_, err1 := r.ConnectWs(address)
```
guymguym: We probably want a short delay before reconnect to avoid tight loops when the address is wrong or there is no network available...

test/rpcbench/rpc_benchmark.go
```diff
@@ -0,0 +1,69 @@
...
+var server = flag.Bool("server", false, "server option")
+var help = flag.Bool("help", false, "help option")
+var port = flag.String("port", "5656", "port option")
+var hostname = flag.String("hostname", "127.0.0.1", "hostname option")
+var wsize = flag.Int64("wsize", 1024*1024, "wsize option")
+var rsize = flag.Int64("rsize", 1024*1024, "rsize option")
+var concur = flag.Int("concur", 5, "concur option")
+
+// tries connecting via wss protocol
+// Run test: go run {PATH.rpc_benchmark_test.go} --client -v
```
guymguym:
```suggestion
// noobaa-core/     $ node src/rpc/rpc_benchmark.js --server
// noobaa-operator/ $ go run test/rpcbench/rpcbench.go --client -v
```

pkg/nb/rpc_ws.go
```diff
@@ -0,0 +1,334 @@
...
+
+// GetAddress returns the connection address
+func (c *RPCConnWS) GetAddress() string {
+	return c.Address
+}
+
+// Reconnect connects after setting a delay
+func (c *RPCConnWS) Reconnect() {
+	c.ReconnectDelay = 3 * time.Second
+	c.Connect()
```
guymguym: We forgot to lock the connection here because Connect requires a lock...

pkg/nb/rpc_ws.go
```diff
@@ -0,0 +1,334 @@
...
+		return err
+	}
+
+	return <-replyChan
+}
+
+// Connect is opening a ws connection for new connection or after the previous one closed
+// it can delay the reconnect attempts in case of repeated failures
+// such as when the host is unreachable, etc.
+func (c *RPCConnWS) Connect() error {
```
guymguym: Lets rename this func to `ConnectUnderLock()`

pkg/nb/rpc_ws.go
romayalon: true, but I think the lock should be inside connect because before the dial there is a delay..

PR #189, opened 2019-12-24
OBC provisioner will set BUCKET_HOST to the s3 service internal DNS

### LINKED ISSUE #181: OBC provisioner should set BUCKET_HOST to the s3 service internal DNS
Currently it uses the NodePort address because we thought it can be useful to set the address to be accessible for clients outside of the cluster that have access to the nodes network too.

But it seems to be too confusing so we would like to set the `BUCKET_HOST` to be the service DNS address `s3.<namespace>.svc`.

The CLI command `noobaa obc status <name>` should show the external address (or NodePort if no external address) for a user to test the OBC access from an external client machines - at least in the s3 alias command here -
https://github.com/noobaa/noobaa-operator/blob/d36eb7dc53db2e2d4f535de24bdcacf005c82be2/pkg/obc/obc.go#L267

jackyalbo:
fixes #181

### DIFF
--- a/pkg/system/system.go
+++ b/pkg/system/system.go
@@ -615,6 +615,7 @@ type Client struct {
 	NooBaa      *nbv1.NooBaa
 	NBClient    nb.Client
 	S3Client    *s3.S3
+	S3DnsURL    *url.URL
 	MgmtURL     *url.URL
 	ServiceMgmt *corev1.Service
 	SecretOp    *corev1.Secret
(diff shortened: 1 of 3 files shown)


pkg/system/system.go
```diff
@@ -716,6 +726,7 @@ func Connect(usePortForwarding bool) (*Client, error) {
 		NooBaa:      r.NooBaa,
 		NBClient:    nbClient,
 		S3Client:    s3.New(s3Session),
+		S3DnsURL:    s3DnsURL,
```
guymguym: It is not clear that this is the internal DNS.
I would suggest to refrain from doing this logic inside Connect().
The provisioner can simply do this on its own.

jackyalbo: fixed in PR #179

PR #191, opened 2019-12-26
olm.go GenerateCSV fixes 

dannyzaken:
* OCP UI crashed when encountered tolerations in noobaa cr
   * removed tolerations descriptor from OLM to prevent crash
* fixed wrong descriptor for `imagePullSecret`

### DIFF
--- a/pkg/olm/olm.go
+++ b/pkg/olm/olm.go
@@ -273,12 +273,6 @@ func GenerateCSV(opConf *operator.Conf) *operv1.ClusterServiceVersion {
 				DisplayName:  "PV Pool DefaultStorageClass",
 			},
 			operv1.SpecDescriptor{
-				Path:         "tolerations",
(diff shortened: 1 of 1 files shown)


pkg/olm/olm.go
```diff
@@ -272,15 +272,9 @@ func GenerateCSV(opConf *operator.Conf) *operv1.ClusterServiceVersion {
 				Description:  "PVPoolDefaultStorageClass (optional) overrides the default cluster StorageClass for the pv-pool volumes. This affects where the system stores data chunks (encrypted). Updates to this field will only affect new pv-pools, but updates to existing pools are not supported by the operator.",
 				DisplayName:  "PV Pool DefaultStorageClass",
 			},
-			operv1.SpecDescriptor{
```
guymguym: @dannyzaken Instead of deleting can you leave it commented out with a comment that explains the problem?

PR #192, opened 2019-12-29
Expose endpoint group Id to the endpoint pods

nb-ohad:
Using the NooBaa CRD UID because it is always present and unique for each NooBaa.


pkg/system/phase4_configuring.go
```diff
@@ -147,6 +147,9 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() {
 					}
 					c.Env[j].Value = fmt.Sprintf(strings.Join(hosts[:], " "))
 
+				case "ENDPOINT_GROUP_ID":
+					c.Env[j].Value = fmt.Sprint(r.NooBaa.UID)
```
guymguym: BTW UID is a string type, just need a simple cast like this `string(r.NooBaa.UID)`

PR #193, opened 2019-12-31
Update obc.go

tamireran:
typo


guymguym:
Added in #188 instead

PR #196, opened 2020-01-07
Update to use operator-sdk v0.13.0 and controller-runtime v0.4.0

### LINKED ISSUE #168: Update to use operator-sdk v0.13.0 and controller-runtime v0.4.0
https://github.com/operator-framework/operator-sdk/releases/tag/v0.13.0
https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.4.0

jackyalbo:
fixes #168


pkg/system/phase2_creating.go
```diff
@@ -154,6 +154,7 @@ func (r *Reconciler) SetDesiredNooBaaDB() {
 	if r.NooBaaDB.UID == "" {
 		for i := range r.NooBaaDB.Spec.VolumeClaimTemplates {
 			pvc := &r.NooBaaDB.Spec.VolumeClaimTemplates[i]
+			pvc.Namespace = r.NooBaaDB.Namespace
```
guymguym: Is this change related?

go.mod
```diff
@@ -3,55 +3,99 @@ module github.com/noobaa/noobaa-operator/v2
...
-	k8s.io/apiextensions-apiserver v0.0.0-20190315093550-53c4693659ed
-	k8s.io/apimachinery v0.0.0-20191005115455-e71eb83a557c
-	k8s.io/cli-runtime v0.0.0-20191005121332-4d28aef60981
-	k8s.io/client-go v11.0.0+incompatible
-	k8s.io/code-generator v0.0.0-20191003035328-700b1226c0bd
-	k8s.io/gengo v0.0.0-20190822140433-26a664648505
-	k8s.io/kube-openapi v0.0.0-20190816220812-743ec37842bf
-	k8s.io/kubectl v0.0.0-20191005122440-987b623dc1f7
+	golang.org/x/crypto v0.0.0-20191119213627-4f8c1d86b1ba
+	k8s.io/api v0.0.0
```
guymguym: I am not sure but seems like this means there is no version requirement on this at all.
I would try to verify what does this mean and if this is how it should work.

go.mod
```diff
@@ -3,55 +3,99 @@ module github.com/noobaa/noobaa-operator/v2
...
-	sigs.k8s.io/controller-tools v0.1.12
-	sigs.k8s.io/testing_frameworks v0.1.2 // indirect
+	sigs.k8s.io/controller-runtime v0.4.0
 	sigs.k8s.io/yaml v1.1.0
 )
 
-// Pinned to kubernetes-1.13.4
-// This is because operator-sdk v0.10.0 still requires kubernetes-1.13.4 and controller-runtime v0.1.10
-// once we can bump the operator-sdk version we can also bump k8s.io/* and controller-runtime
+// replace (
```
guymguym: Cleanup?

pkg/system/phase2_creating.go
jackyalbo: it failes the next step in the namespace isn't set - so yes

go.mod
jackyalbo: i will set according to the replace

jackyalbo: posted new question to #forum-operator-fw channel in CoreOs slack

guymguym: For future reference, the response was:
> It’s a quirk of k/k and k/k is still a transitive dependency of operator-sdk.
There’s some discussion here: https://github.com/kubernetes/kubernetes/issues/79384
The TL;DR is that you need to use replace to pull in the correct sub project dependencies.
`operator-sdk print-deps` should help.

PR #198, opened 2020-01-07
Externalise errors in backingstore and bucketclass

romayalon:
### Explain the changes
1. Added modes to backingstore and bucketclass structs.
2. Updated the new modes after receiving alert from cluster.
3. Updated backingstore phase if needed and added event in the  relevant backingstore description.
4. Added event in bucketclass description after mode changed.

1. Fixed bugzilla bugs: 1765865 and 1781096
2. Opened #202 - Further discussion on phases for transient/in-progress states such as INIT and DELETE

### Testing Instructions:
1.


pkg/backingstore/reconciler.go
```diff
@@ -24,6 +24,32 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
+var bsPhaseToPoolMode map[string]nbv1.BackingStorePhase
```
dannyzaken: shouldn't it be `bsModeToPhase`?

pkg/backingstore/reconciler.go
```diff
@@ -131,12 +157,26 @@ func (r *Reconciler) Reconcile() (reconcile.Result, error) {
...
-		log.Infof("✅ Done")
+		curMode := &nbv1.BackingStoreMode{
+			ModeMessage: r.BackingStore.Status.Mode.ModeMessage,
+			ChangeTime:  r.BackingStore.Status.Mode.ChangeTime,
+		}
+
+		bsPhase, exist := bsPhaseToPoolMode[curMode.ModeMessage]
+
+		if exist && bsPhase != r.BackingStore.Status.Phase {
+			r.SetPhase(bsPhase, curMode.ModeMessage, curMode.ChangeTime)
```
dannyzaken: maybe pass a more descriptive `reason` and `message`

pkg/backingstore/reconciler.go
```diff
@@ -24,6 +24,32 @@ import (
...
 )
 
+var bsPhaseToPoolMode map[string]nbv1.BackingStorePhase
+
+func init() {
+	bsPhaseToPoolMode = newBSPhaseModeMap()
+}
+func newBSPhaseModeMap() map[string]nbv1.BackingStorePhase {
+	return map[string]nbv1.BackingStorePhase{
+		"INITIALIZING":        nbv1.BackingStorePhaseReady,
```
dannyzaken: I think we should reconsider some of these mappings. 
for example `INITIALIZING` does not promise that the resource is ready or will be ready eventually. also `DELETING` is not really ready.

nimrod-becker: @dannyzaken 
You are basically correct, but I don't see another good option.
- Mapping those to "creating" and "deleting", would result in the operator reconciling this phase and actually taking action according to that phase, which is not the desired behavior.
- Adding new phases that will be there only for visibility and the reconcile for that phase won't do anything also seems kind of weird since they are not actual phases of the resource / CRD ...

nimrod-becker: In addition, thinking about an action that originated from the UI (for example delete bucket) and the current desired behavior we defined (operator should reconcile and re-create the resource), setting DELETING would cause the operator to reconcile delete and actually delete the resource.

 We might want to change our decision and have deleted from either API (CRD or UI), if we do setting DELETION would make sense.

A vote in favor of deleting both would be the fact we need to test and answer what will happen if we delete a resource (UI), getting it into a deleting state within the core (moving data etc.) and then we'll re-create it...

@guymguym a penny for your thoughts

romayalon: What about the following idea: 
Instead of mapping from mode to phase, mapping from mode to struct that contains nbv1.BackingStorePhase, message and reason. 
this way, when pool is for example in deleting mode it will still be mapped to nbv1.BackingStorePhaseReady but the reason/message will be "DELETING"

dannyzaken: @romayalon this is a good idea regardless of the specific mode mappings. 
I still think we need to indicate that a backing store is not ready in certain situations.
It is not necessarily critical to do in this PR, but maybe we should open an issue on that and move the discussion there.

nimrod-becker: @dannyzaken raised a good point about operators which for example create EC2 instances, which can take some time and probably do go directly into ready

PR #200, opened 2020-01-12
fixing crd files

jackyalbo:
1. using operator-sdk generate crds instead of genapi which is deprecated
2. fixing the crds for the new naming convention in operator_sdk:

generate crds generates CRDs or updates them if they exist,
under deploy/crds/<full group>_<resource>_crd.yaml; OpenAPI
V3 validation YAML is generated as a 'validation' object.

Example:
	$ operator-sdk generate crds
	$ tree deploy/crds
	├── deploy/crds/app.example.com_v1alpha1_appservice_cr.yaml
	├── deploy/crds/app.example.com_appservices_crd.yaml


jackyalbo:
> @jackyalbo something looks wrong - where are the newly generated CRDs now??

Well they were already pushed in my previous PR #196

guymguym:
@jackyalbo OK, so how about renaming also the CR files to the same pattern?
Also please search `noobaa_v1alpha1` in the code because I see references in .md files and others.

jackyalbo:
@guymguym as you can see from the operator-sdk help attached... looks like the cr name convention should be different: 
deploy/crds/app.example.com_v1alpha1_appservice_cr.yaml
the resource in singular and still with the version (v1alpha1)
- and it didn't changes automatically the cr names... so not sure if we only need to change the beginning to noobaa.io_v1alpha1 or just leave it. 

I will fix the references of the CRDs

jackyalbo:
Adapted according to the help file: 
CR: deploy/crds/noobaa.io_v1alpha1_[singular-resource-name]_cr.yaml
CRD: deploy/crds/noobaa.io_[plural-resource-name]_crd.yaml

PR #203, opened 2020-01-15
Support deployment of a joined NooBaa on a remote cluster 

nb-ohad:
Currently, we mainly deploy an S3 service and an auto-scaled endpoint group

Changes:
- Create an auth_token secret to be used by endpoints
- Introduce a join secret to hold information on how to join an existing NooBaa cluster
- Add a join secret reference on the NooBaa CRD to indicate that the deployed NooBaa should join an existing cluster using the join information specified in the referenced secret

### DIFF
--- a/deploy/crds/noobaa.io_noobaas_crd.yaml
+++ b/deploy/crds/noobaa.io_noobaas_crd.yaml
@@ -753,6 +753,19 @@ spec:
                     TODO: Add other useful fields. apiVersion, kind, uid?'
                   type: string
               type: object
+            joinSecret:
+              description: JoinSecret (optional) instructs the operator to join another
+                cluster and point to a secret that holds the join information
+              properties:
+                name:
+                  description: Name is unique within a namespace to reference a secret
+                    resource.
+                  type: string
+                namespace:
+                  description: Namespace defines the space within which the secret
+                    name must be unique.
+                  type: string
+              type: object
             pvPoolDefaultStorageClass:
               description: PVPoolDefaultStorageClass (optional) overrides the default
                 cluster StorageClass for the pv-pool volumes. This affects where the
--- a/deploy/internal/deployment-endpoint.yaml
+++ b/deploy/internal/deployment-endpoint.yaml
@@ -46,16 +46,22 @@ spec:
         - containerPort: 6001
         - containerPort: 6443
         env:
-        - name: JWT_SECRET
-          valueFrom:
-            secretKeyRef:
-              name: noobaa-server
-              key: jwt
-        - name: MGMT_URL
+        - name: MGMT_ADDR
+        - name: BG_ADDR
+        - name: MD_ADDR
+        - name: HOSTED_AGENTS_ADDR
         - name: MONGODB_URL
         - name: VIRTUAL_HOSTS
         - name: REGION
         - name: ENDPOINT_GROUP_ID
+        - name: LOCAL_MD_SERVER
+        - name: LOCAL_N2N_AGENT
+        - name: JWT_SECRET
+        - name: NOOBAA_AUTH_TOKEN
+          valueFrom:
+            secretKeyRef:
+              name: noobaa-endpoints
+              key: auth_token
         - name: CONTAINER_CPU_REQUEST
           valueFrom:
             resourceFieldRef:
--- a/pkg/apis/noobaa/v1alpha1/noobaa_types.go
+++ b/pkg/apis/noobaa/v1alpha1/noobaa_types.go
@@ -121,7 +121,13 @@ type NooBaaSpec struct {
 
 	// Endpoints (optional) sets configuration info for the noobaa endpoint
 	// deployment.
+	// +optional
 	Endpoints *EndpointsSpec `json:"endpoints,omitempty"`
+
+	// JoinSecret (optional) instructs the operator to join another cluster
+	// and point to a secret that holds the join information
+	// +optional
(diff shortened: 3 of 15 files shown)


deploy/internal/deployment-endpoint.yaml
```diff
@@ -46,16 +46,22 @@ spec:
...
-        - name: JWT_SECRET
-          valueFrom:
-            secretKeyRef:
-              name: noobaa-server
-              key: jwt
         - name: MGMT_URL
+        - name: MGMT_PORT
+        - name: BG_PORT
+        - name: MD_PORT
+        - name: HOSTED_AGENTS_PORT
```
dannyzaken: do we need `HOSTED_AGENTS_PORT`?

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -121,7 +121,13 @@ type NooBaaSpec struct {
 
 	// Endpoints (optional) sets configuration info for the noobaa endpoint
 	// deployment.
+	// +optional
 	Endpoints *EndpointsSpec `json:"endpoints,omitempty"`
+
+	// JoinSecret (optional) instracts the operator to join another cluster
```
dannyzaken: typo - ~~instracts~~ -> instructs

deploy/crds/noobaa.io_noobaas_crd.yaml
```diff
@@ -753,6 +753,19 @@ spec:
                     TODO: Add other useful fields. apiVersion, kind, uid?'
                   type: string
               type: object
+            joinSecret:
+              description: JoinSecret (optional) instracts the operator to join another
```
nimrod-becker:
```suggestion
              description: JoinSecret (optional) instructs the operator to join another
```

pkg/system/phase2_creating.go
```diff
@@ -31,41 +31,57 @@ func (r *Reconciler) ReconcilePhaseCreating() error {
 		"noobaa operator started phase 2/4 - \"Creating\"",
 	)
 
-	// A failure to discover OAuth endpoints should not fail the entire reconcile phase.
-	oAuthEndpoints, err := util.DiscoverOAuthEndpoints()
-	if err != nil {
-		r.Logger.Warnf("Discovery of OAuth endpoints failed, got: %v", err)
+	if r.JoinSecret == nil {
```
nimrod-becker: Isn't it better to split into two ReconcilePhaseCreating? One for Join nil and for for !nil?

deploy/internal/deployment-endpoint.yaml
nimrod-becker: +1 , Assuming we work with n2n_proxy we probably don't right?

pkg/system/phase4_configuring.go
```diff
@@ -31,10 +31,16 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 		"noobaa operator started phase 4/4 - \"Configuring\"",
 	)
 
-	if err := r.ReconcileSecretOp(); err != nil {
-		return err
+	if r.JoinSecret == nil {
```
nimrod-becker: Same as the comment on phase2? Maybe also 3 (even though its not a lot of changes)

pkg/system/phase2_creating.go
```diff
@@ -31,41 +31,57 @@ func (r *Reconciler) ReconcilePhaseCreating() error {
...
+		// A failure to discover OAuth endpoints should not fail the entire reconcile phase.
+		oAuthEndpoints, err := util.DiscoverOAuthEndpoints()
```
dannyzaken: should this be inside the `if`?

pkg/system/phase2_creating.go
guymguym: sounds better to me too

nb-ohad: @nimrod-becker @guymguym 
I don't think so. It creates duplicate code that needs to be updated for every change in both places.
Also, the end goal (in the future) is to have full NooBaa on the remote cluster which means 
that we will take reconcile function out of the if statements little by little

deploy/internal/deployment-endpoint.yaml
nb-ohad: These ports are moved to the other cluster to support a full router configuration, even if currently we do not invoke any API calls that are targeted in the hosted agents.
In any case, we still need to provide the server_rpc router an address (which includes a port) for the hosted_agents domain on the router

pkg/system/phase4_configuring.go
nb-ohad: Please see my answer above

pkg/system/phase2_creating.go
dannyzaken: at the moment it's not very readable. there is no need to duplicate code. if there is a lot of common code you can extract it to a function

nimrod-becker: lets not create duplicate code. Create a common base which runs in both cases and then a specific func for each case.

deploy/internal/deployment-endpoint.yaml
nb-ohad: Not using ports anymore - Changed to use full address for each service

pkg/system/phase2_creating.go
nb-ohad: I changed the way the code is structured. The main change is that each reconcile function checks for itself if it needs to be skipped

PR #207, opened 2020-01-19
Report on endpoint group config changes to the core

nb-ohad:
Also, support the NBClient for remote deployments
and bump noNooBaa core image reference

### DIFF
--- a/pkg/system/phase3_connecting.go
+++ b/pkg/system/phase3_connecting.go
@@ -24,11 +24,11 @@ func (r *Reconciler) ReconcilePhaseConnecting() error {
 
 	if r.JoinSecret == nil {
 		r.CheckServiceStatus(r.ServiceMgmt, r.RouteMgmt, &r.NooBaa.Status.Services.ServiceMgmt, "mgmt-https")
-
-		if err := r.InitNBClient(); err != nil {
-			return err
-		}
 	}
+	if err := r.InitNBClient(); err != nil {
+		return err
+	}
+
 	r.CheckServiceStatus(r.ServiceS3, r.RouteS3, &r.NooBaa.Status.Services.ServiceS3, "s3-https")
 
 	return nil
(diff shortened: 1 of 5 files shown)


pkg/system/phase3_connecting.go
```diff
@@ -136,24 +136,31 @@ func (r *Reconciler) CheckServiceStatus(srv *corev1.Service, route *routev1.Rout
...
-		ServiceMgmt: r.ServiceMgmt,
-		NodeIP:      mgmtURL.Hostname(),
-	})
+		r.NBClient = nb.NewClient(&nb.APIRouterNodePort{
+			ServiceMgmt: r.ServiceMgmt,
+			NodeIP:      mgmtURL.Hostname(),
+		})
+
+	} else {
+		r.NBClient = nb.NewClient(&nb.SimpleRouter{
```
guymguym: Nice that SimpleRouter is useful here too! check it out @romayalon

pkg/system/phase4_configuring.go
```diff
@@ -220,6 +222,24 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() {
...
 }
 
+// ReconcileHPAEndpoint reconcile the endpoint's HPS and report the configuration
+// back to the noobaa core
+func (r *Reconciler) ReconcileHPAEndpoint() error {
+	if err := r.ReconcileObject(r.HPAEndpoint, r.SetDesiredHPAEndpoint); err != nil {
+		return err
+	}
+
+	endpointsSpec := r.NooBaa.Spec.Endpoints
```
guymguym: What if there is no endpoints spec in the NooBaa CR?
Why not use `r.HPAEndpoint` instead?

nb-ohad: @guymguym Greate catch

PR #208, opened 2020-01-21
Check if storageclass doesn't exist before creating OBC 

romayalon:
### Explain the changes
1. Added a check before obc creation that fails when storageclass doesn't exist.

1. Fixed BZ 1785260

### Testing Instructions:
1.


pkg/obc/obc.go
```diff
@@ -112,6 +112,13 @@ func RunCreate(cmd *cobra.Command, args []string) {
 		obc.Spec.GenerateBucketName = name
 	}
 	obc.Spec.StorageClassName = options.SubDomainNS()
+	sc := &storagev1.StorageClass{}
```
guymguym: Styling suggestion - setting the TypeMeta will help KubeCheck show the right messages, and also the fatal message was a bit confusing since StorageClass has no namespace of its own:

```golang
sc := &storagev1.StorageClass{
	TypeMeta: metav1.TypeMeta{Kind: "StorageClass"},
	ObjectMeta: metav1.ObjectMeta{Name: options.SubDomainNS()},
}
if !util.KubeCheck(sc) {
	log.Fatalf(`❌ Could not get StorageClass %q for system in namespace %q`,
		sc.Name, options.Namespace)
}
obc.Spec.StorageClassName = sc.Name
```

PR #216, opened 2020-01-28
rgw: Disable SSL explicitly on rook-ceph-rgw endpoint

mkogan1:
Signed-off-by: Mark Kogan <mkogan@redhat.com>

### DIFF
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -474,6 +474,7 @@ func (r *Reconciler) prepareCephBackingStore() error {
 	endpoint := "http://rook-ceph-rgw-" + r.CephObjectstoreUser.Spec.Store + "." + options.Namespace + ":80"
 	region := "us-east-1"
 	forcePathStyle := true
+	disableSSL := true
 	s3Config := &aws.Config{
 		Credentials: credentials.NewStaticCredentials(
 			cephObjectUserSecret.StringData["AccessKey"],
@@ -483,6 +484,7 @@ func (r *Reconciler) prepareCephBackingStore() error {
 		Endpoint:         &endpoint,
 		Region:           &region,
 		S3ForcePathStyle: &forcePathStyle,
+		DisableSSL:       &disableSSL,
 	}
 	bucketName := r.generateBackingStoreTargetName()
 	if err := r.createS3BucketForBackingStore(s3Config, bucketName); err != nil {


nimrod-becker:
@mkogan1 Is this the issue? What changed? (Since we've seen and verified the flow without disabling)

mkogan1:
> @mkogan1 Is this the issue? What changed? (Since we've seen and verified the flow without disabling)
Do not have information regarding endpoint configuration in previous testing.
Saw amazon GO API documentation stating that SSL is enabled by default
https://docs.aws.amazon.com/sdk-for-go/api/aws/#Config
```
    // Set this to `true` to disable SSL when sending requests. Defaults
    // to `false`.
    DisableSSL *bool
```

nimrod-becker:
@mkogan1 ok, but I'm a little confused. 
Did you find traces for the request getting to its destination? If so why an HTML error was returned ?

dannyzaken:
@nimrod-becker @mkogan1 according to comment #31 in [the BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1793189), the error is probably due to a cluster-wide proxy, which is bypassing the NO_PROXY settings somehow.

I don't think that the `DisableSSL` option changed its behavior since we tested

nimrod-becker:
> @nimrod-becker @mkogan1 according to comment 31 in [the BZ](https://bugzilla.redhat.com/show_bug.cgi?id=1793189), the error is probably due to a http proxy, which is bypassing the NO_PROXY settings somehow.
> 
> I don't think that the `DisableSSL` option changed its behavior since we tested

Same Build , Same Code that was tested

dannyzaken:
I updated the BZ. my guess is that the cluster-wide proxy settings contain `NO_PROXY=.cluster.local`, but we don't use the `.cluster.local` suffix for the RGW endpoint

https://github.com/noobaa/noobaa-operator/blob/130a4308992b91a7974f6ed75878eb454e1c0c33/pkg/system/phase4_configuring.go#L474

we should add the suffix to the endpoint

dannyzaken:
created PR #217

nimrod-becker:
@mkogan1 as we have the real RC suspect, I'm going to close this one.

mkogan1:
> @mkogan1 as we have the real RC suspect, I'm going to close this one.

Yes, per input on the BZ this can be closed

PR #217, opened 2020-01-28
added svc.cluster.local to RGW endpoint

dannyzaken:
should fix BZ https://bugzilla.redhat.com/show_bug.cgi?id=1793189

still needs to be tested


deploy/operator.yaml
```diff
@@ -22,6 +22,8 @@ spec:
               cpu: "250m"
               memory: "512Mi"
           env:
+            - name: NO_PROXY
```
guymguym: Don't we need it for the core/db pods too?

dannyzaken: I've set it for core. I can do it for DB as well, but not sure we need

PR #218, opened 2020-02-02
validate generated bucket name for backing store targer

dannyzaken:
fixes BZ https://bugzilla.redhat.com/show_bug.cgi?id=1796570

* if truncated bucket name ends with a dash then remove it.
* validate generated bucket name and fall back to a simple name (nb + timestamp)

### DIFF
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -505,10 +505,25 @@ func (r *Reconciler) generateBackingStoreTargetName() string {
 	tsMilli := strconv.FormatInt(time.Now().UnixNano()/int64(time.Millisecond), 10)
 	name := "nb." + tsMilli
 	if r.RouteMgmt.Spec.Host != "" {
-		name += "." + r.RouteMgmt.Spec.Host
+		suffix := ""
+		hostItems := strings.Split(r.RouteMgmt.Spec.Host, ".")
+		for i := len(hostItems) - 1; i >= 0; i-- {
+			hostItem := strings.Trim(hostItems[i], "-.")
+			if len(name)+1+len(hostItem)+1+len(suffix) > 63 {
+				break
+			}
+			suffix = hostItem + "." + suffix
+		}
+		name += "." + suffix
 	}
-	if len(name) > MaxNameLength {
-		name = name[:MaxNameLength]
+	// make sure the name is ended with a valid charecter
+	name = strings.Trim(name, "-.")
+
+	// if for some reason the bucket name is not valid then fallback to nb.timestamp
+	if !util.IsValidS3BucketName(name) {
+		oldName := name
+		name = "nb." + tsMilli
+		logrus.Warnf("generated bucket name (%s) is invalid. falling back to (%s)", oldName, name)
 	}
 	return name
 }
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -15,6 +15,7 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"time"
 	"unicode"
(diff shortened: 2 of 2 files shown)


pkg/util/util.go
```diff
@@ -777,6 +777,40 @@ func GetAWSRegion() (string, error) {
...
 
+// IsValidS3BucketName checks the name according to
+// https://docs.aws.amazon.com/awscloudtrail/latest/userguide/cloudtrail-s3-bucket-naming-requirements.html
+func IsValidS3BucketName(name string) bool {
+	//The bucket name can be between 3 and 63 characters long
+	if len(name) < 3 || len(name) > 63 {
+		return false
+	}
+
+	for i, c := range name {
```
guymguym: Why not regexp?
https://github.com/noobaa/noobaa-core/blob/f8eb884322ddc17848eaaed1cf4c768c7620d9bc/src/server/system_services/bucket_server.js#L39-L40

pkg/system/phase4_configuring.go
```diff
@@ -510,6 +510,17 @@ func (r *Reconciler) generateBackingStoreTargetName() string {
 	if len(name) > MaxNameLength {
 		name = name[:MaxNameLength]
 	}
+	// make sure the name is ended with a valid charecter
+	for string(name[len(name)-1]) == "-" || string(name[len(name)-1]) == "." {
```
guymguym: Better use https://golang.org/pkg/strings/#Trim:
```
name = strings.Trim(name, "-.")
```

pkg/system/phase4_configuring.go
```diff
@@ -510,6 +510,17 @@ func (r *Reconciler) generateBackingStoreTargetName() string {
 	if len(name) > MaxNameLength {
 		name = name[:MaxNameLength]
```
guymguym: I think this is not what we want. Instead of truncating the end of the route name, we should prefer to truncate its beginning and keep the end because in domains this is usually the more identifying part.
So I would change this flow to something like this:
```golang
	name := "nb." + tsMilli
	if r.RouteMgmt.Spec.Host != "" {
		suffix := ""
		hostItems := strings.Split(r.RouteMgmt.Spec.Host, ".")
		for i := len(hostItems) - 1; i >= 0; i-- {
			hostItem := strings.Trim(hostItems[i], "-.")
			if len(name) + 1 + len(hostItem) + 1 + len(suffix) > 63 {
				break
			}
			suffix = hostItem + "." + suffix
		}
		name += "." + suffix
	}
```

pkg/system/phase4_configuring.go
dannyzaken: 👍

pkg/util/util.go
dannyzaken: changed

pkg/system/phase4_configuring.go
dannyzaken: you're right. I'll change

pkg/system/phase4_configuring.go
```diff
@@ -505,10 +505,25 @@ func (r *Reconciler) generateBackingStoreTargetName() string {
 	tsMilli := strconv.FormatInt(time.Now().UnixNano()/int64(time.Millisecond), 10)
 	name := "nb." + tsMilli
 	if r.RouteMgmt.Spec.Host != "" {
-		name += "." + r.RouteMgmt.Spec.Host
+		suffix := ""
+		hostItems := strings.Split(r.RouteMgmt.Spec.Host, ".")
+		for i := len(hostItems) - 1; i >= 0; i-- {
+			hostItem := strings.Trim(hostItems[i], "-.")
+			if len(name)+1+len(hostItem)+1+len(suffix) > 63 {
```
guymguym: Suffix always includes an extra dot at the end, so this check will limit the length to one character less than the max.
This is fine, but we can also fix it.

PR #224, opened 2020-02-05
Backport: added svc.cluster.local to RGW endpoint

nimrod-becker:


guymguym:
We suspect that the CI failure are because minikube might be configured different than locally.
@dannyzaken verified manually this PR manually on his minikube.
So I am force merging now and we should followup and fix 2.0 CI next week.

PR #227, opened 2020-02-09
use default volume size instead of crashing when VolumeResources is not set in backingstore

### LINKED ISSUE #222: BackingStore pv-pool has optional resources but operator crashes when null
optional resources in spec:
https://github.com/noobaa/noobaa-operator/blob/8fb75596137a9359e4c00cc7c6f84eb3c5e5d8a9/pkg/apis/noobaa/v1alpha1/backingstore_types.go#L221

operator crash location:
https://github.com/noobaa/noobaa-operator/blob/8fb75596137a9359e4c00cc7c6f84eb3c5e5d8a9/pkg/backingstore/reconciler.go#L417

dannyzaken:
Fixes #222 

* use 20GB as default PV size

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -413,16 +413,23 @@ func (r *Reconciler) ReadSystemInfo() error {
 				r.BackingStore.Name, pool, pool.ResourceType,
 			))
 		}
+		const defaultVolumeSize = int64(21474836480) // 20Gi=20*1024^3
+		var volumeSize int64
 		pvPool := r.BackingStore.Spec.PVPool
-		qty := pvPool.VolumeResources.Requests[corev1.ResourceName(corev1.ResourceStorage)]
-		gbsize, _ := qty.AsInt64()
+		if pvPool.VolumeResources != nil {
+			qty := pvPool.VolumeResources.Requests[corev1.ResourceName(corev1.ResourceStorage)]
+			volumeSize = qty.Value()
+		}
+		if volumeSize == 0 {
+			volumeSize = int64(defaultVolumeSize)
+		}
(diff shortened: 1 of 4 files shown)


pkg/backingstore/reconciler.go
```diff
@@ -413,16 +413,22 @@ func (r *Reconciler) ReadSystemInfo() error {
 				r.BackingStore.Name, pool, pool.ResourceType,
 			))
 		}
+		gbsize := int64(20)
 		pvPool := r.BackingStore.Spec.PVPool
-		qty := pvPool.VolumeResources.Requests[corev1.ResourceName(corev1.ResourceStorage)]
-		gbsize, _ := qty.AsInt64()
+		if pvPool.VolumeResources != nil {
+			qty := pvPool.VolumeResources.Requests[corev1.ResourceName(corev1.ResourceStorage)]
+			gbsize, _ = qty.AsInt64()
```
guymguym: Can we use `qty.Value()`?
https://pkg.go.dev/k8s.io/apimachinery/pkg/api/resource?tab=doc#Quantity.Value

pkg/backingstore/reconciler.go
```diff
@@ -413,16 +413,22 @@ func (r *Reconciler) ReadSystemInfo() error {
...
+			if gbsize == 0 {
```
guymguym: The control flow is a bit cumbersome now. May I suggest something like this:
```golang
var gbsize int64
if pvPool.VolumeResources != nil {
	gbsize = pvPool.VolumeResources.Requests[corev1.ResourceStorage].Value()
}
if gbsize == 0 {
	gbsize = 20
}
```

PR #230, opened 2020-02-12
changed RPC client in the operator to connect using the management service DNS name

dannyzaken:
* when connecting through the node port, it is possible that there is no connection between different nodes in the cluster using the node ports (e.g. firewall)
* instead of the node port, using the mgmt service DNS name `noobaa-mgmt.[namespace].svc.cluster.local`

### DIFF
--- a/pkg/system/phase3_connecting.go
+++ b/pkg/system/phase3_connecting.go
@@ -2,7 +2,6 @@ package system
 
 import (
 	"fmt"
-	"net/url"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


pkg/system/phase3_connecting.go
```diff
@@ -141,15 +140,8 @@ func (r *Reconciler) InitNBClient() error {
 			return fmt.Errorf("mgmt port not ready yet")
```
guymguym: I guess this check is not really making much sense now...

dannyzaken: correct

dannyzaken: correct

PR #236, opened 2020-02-18
Reverting the changes for operator-sdk 0.13

jackyalbo:
revert 2 PRs - one for updating the operator-sdk to 0.13, and the second for updating the cd/crds for the new operator-sdk api format
A 3rd commit is for small extra fixes that were missing - and for the reverted apis being rebuilt with the new code that was added after the PRs


pkg/system/phase2_creating.go
```diff
@@ -170,7 +170,6 @@ func (r *Reconciler) SetDesiredNooBaaDB() error {
 	if r.NooBaaDB.UID == "" {
 		for i := range r.NooBaaDB.Spec.VolumeClaimTemplates {
 			pvc := &r.NooBaaDB.Spec.VolumeClaimTemplates[i]
-			pvc.Namespace = r.NooBaaDB.Namespace
```
nb-ohad: I think setting the PVC namespace is full proofing even if r.own does not need it in this version of the SDK

test/cli/test_cli_functions.sh
```diff
@@ -68,19 +64,14 @@ function test_noobaa {
...
             then
-                echo_time "✅  ${noobaa} ${options} failed - as should"
+                echo "✅  ${noobaa} ${options} failed - as should"
             else 
-                echo_time "❌  ${noobaa} ${options} failed, Exiting"
-                local pod_operator=$(kuberun get pod | grep noobaa-operator | awk '{print $1}')
-                echo_time "==============OPERATOR LOGS============"
-                kuberun logs ${pod_operator}
-                echo_time "==============CORE LOGS============"
-                kuberun logs noobaa-core-0
```
nb-ohad: The print above might be helpful even after the revert

jackyalbo: reverted back

pkg/system/phase2_creating.go
jackyalbo: You are probably right, but as this is not mandatory in this operator-sdk version, we can keep it out for now - for the sake of having a complete revert of the PR.
the line is still in the master - So we will keep it for future releases

PR #237, opened 2020-02-18
Allow to disable compression

jeniawhite:
Added additional variable to disable compression
https://bugzilla.redhat.com/show_bug.cgi?id=1800599


guymguym:
@jeniawhite 
1. You are just adding a false env to the sts, but you don't add any spec support? so if we need to disable compression it will be a manual operation on our two sts? a bit funky...
2. Can you get rid of all that whitespace change in the yamls?

PR #255, opened 2020-03-04
The endpoint pod should use the secrets as well

YiannisGkoufas:
The endpoint pod will fail to pull the image if it's in a private docker registry. Therefore we need to specify the image pull secrets the same way we do in the other deployments


guymguym:
Thanks @YiannisGkoufas!

PR #258, opened 2020-03-09
Fix OOMKilled problem

romayalon:
### Explain the changes
1. Instead of creating http.Client  in every call to util.DiscoverOAuthEndpoints() - Moved the http.Client creation to phase2_creating and passed the same client as argument. 
2. Called res.Body.Close() when it's empty.

1. Fixed BZ #1799077

### Testing Instructions:
1.

### DIFF
--- a/pkg/diagnose/diagnose.go
+++ b/pkg/diagnose/diagnose.go
@@ -120,10 +120,7 @@ func (c *Collector) CollectPodLogs(corePodSelector labels.Selector) {
 
 // 	targetAddress := fmt.Sprintf("%s/metrics/counter", mgmtURL.String())
 // 	log.Printf("JENIA THIS IS THE URL %s", targetAddress)
-// 	tr := &http.Transport{
-// 		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-// 	}
-// 	client := &http.Client{Transport: tr}
+// 	client := &http.Client{Transport: util.InsecureHTTPTransport}
 // 	resp, err := client.Get(targetAddress)
 // 	if err != nil {
 // 		log.Printf(`%s`, err)
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"crypto/rand"
 	"crypto/tls"
+	"crypto/x509"
 	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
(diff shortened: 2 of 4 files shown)


pkg/system/phase2_creating.go
```diff
@@ -22,6 +24,17 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+var oauthClient *http.Client
```
nb-ohad: Let's save the client on the reconciler, global var seems wrong.

guymguym: @nb-ohad @romayalon Wouldn't that cause the same problem though? The reconciler object gets created on every reconcile...

nb-ohad: Yes, you are right.

pkg/util/util.go
```diff
@@ -64,8 +68,37 @@ var (
...
+	// SecureHTTPClient is a global secured http client
+	SecureHTTPTransport = newSecureHTTPTransport()
 )
 
+// newSecureHTTPTransport initializes a secured http transport. http.Transport may cause memory leaks when overused
+func newSecureHTTPTransport() *http.Transport {
+	// Load CA cert
+	caCert, err := ioutil.ReadFile(serviceCAPath)
+	if err != nil {
+		return &http.Transport{}
```
guymguym: Do we prefer `return nil` to make it use the default transport?

pkg/diagnose/diagnose.go
```diff
@@ -120,10 +120,7 @@ func (c *Collector) CollectPodLogs(corePodSelector labels.Selector) {
 
 // 	targetAddress := fmt.Sprintf("%s/metrics/counter", mgmtURL.String())
 // 	log.Printf("JENIA THIS IS THE URL %s", targetAddress)
-// 	tr := &http.Transport{
-// 		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
-// 	}
-// 	client := &http.Client{Transport: tr}
+// 	client := &http.Client{Transport: util.SecureHTTPTransport}
```
guymguym:
```suggestion
// 	client := &http.Client{Transport: util.InsecureHTTPTransport}
```

pkg/util/util.go
romayalon: so, returning nil there caused later on a runtime error, even if I checked it for nil and didn't assign anything to the client's transport. returning  &http.Transport{} worked. do you have other suggestion that will be ok ?

guymguym: I'm curious - what is the error when it's nil?

romayalon: Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)

guymguym: of course, but where? what is the stack trace?

PR #261, opened 2020-03-12
update core version to 5.3.0-rc1

jackyalbo:


guymguym:
@jackyalbo @dannyzaken I think we can close this older PR in favor of #269

PR #273, opened 2020-03-25
Addition of disable compression configuration

jeniawhite:


jeniawhite:
@nb-ohad 
We work with snappy and if the user wants a different compression it would require a change in the NB Native and obviously testing cycles for such compressions.
Since we don't really have anything besides snappy right now (that was used heavily and tested), I don't see a point of adding this option right now.
Not sure if there are users that want to change the compression algorithm if there are then they would probably want to configure that round-robin of picking nodes for allocations as well (where do we set the limit?).

nb-ohad:
@jeniawhite it is not about introducing a new compression. It is about API design. The NooBaa CRD is a public API. and we should think about it in this way (forward compatibility) when it makes sense.

In this case, if we take your suggestion but down the line will want to add another compression we will need to do one of two:
1. Break backward compatibility of the public API (CRD)
2. Introduce both fields and have some weird logic in the operator the decide which field has priority (which can be a pain in the ass for the user).

My suggestion solves this problem without introducing a lot of changes to the current code.
The only major thing is to add validation in the operator for 'snappy' or 'none' values in the reconcile.

I hope this clarifies why I think this change is important at the moment even if we have no plan to introduce additional compressions in the near future.

jeniawhite:
@nb-ohad 
I understand the urge to push that while I'm touching it, but all I'm saying is that some may argue the significance and value to the users of such change.
Making complex APIs isn't always a good choice (especially when you don't even have any future plans of adding another compression and no one is interested in that, at least of what I know). Like you've said: "which can be very confusing to the user".
Regarding the 'Introduce both fields', there is no complex logic the enable/disable compression is of the highest priority. It is just a preference of what is better, holding one ENUM or having a boolean with an additional compression name.
I just wanted to clarify that we can add configuration for every little thing in our system but is there a point of investing time in that if no one requested it and if we do not have any future plans for it?
I'm not breaking the API nor making it more complex if we will want to add compression in the future.

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -118,6 +118,10 @@ type NooBaaSpec struct {
 	// +optional
 	Region *string `json:"region,omitempty"`
 
+	// NoobaaDisableCompression (optional) disable compression
+	// +optional
+	NoobaaDisableCompression *string `json:"noobaaDisableCompression,omitempty"`
```
dannyzaken: why `string` and not `bool`?

pkg/system/phase2_creating.go
```diff
@@ -258,6 +258,13 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 					if r.OAuthEndpoints != nil {
 						c.Env[j].Value = r.OAuthEndpoints.TokenEndpoint
 					}
+
+				case "NOOBAA_DISABLE_COMPRESSION":
```
dannyzaken: I know it's already implemented like that in noobaa-core but it bugs me anyway.
I think that for env vars its clearer and less error prone if it doesn't contain double negative (disabled=false)

PR #278, opened 2020-04-06
Cleanup flag added randomly to test cli flow

kerencarmelwedel:
Uninstall cleanup flag added to the flow

PR #280, opened 2020-04-16
uninstall noobaa added with check cleanup flag

kerencarmelwedel:
noobaa uninstall function added with random cleanup flag on each iteration. 
check cleanup flag function added.


liranmauda:
This PR is a Duplicate of #278 
Closing

PR #282, opened 2020-04-19
moving pvpool backstore resposibilty to operator

jackyalbo:
- operator will add, scale-up and delete pvpools during reconcile
- Add support to getting agent-config from noobaa-core
- Add util command wrapper for kube client DeleteAllOf
- small fixes to backingstore commands

### DIFF
--- a/deploy/internal/pod-agent.yaml
+++ b/deploy/internal/pod-agent.yaml
@@ -1,0 +1,46 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  labels:
+    app: noobaa
+  name: noobaa-agent
+  finalizers:
+    - noobaa.io/finalizer
+spec:
+  containers:
+    - name: noobaa-agent
+      image: NOOBAA_CORE_IMAGE
+      imagePullPolicy: IfNotPresent
+      resources:
+        # https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/
+        requests:
+          cpu: "100m"
+          memory: "200Mi"
+        limits:
+          cpu: "100m"
+          memory: "200Mi"
+      env:
+        # Insert the relevant config for the current agent
+        - name: CONTAINER_PLATFORM
+          value: KUBERNETES
+        - name: AGENT_CONFIG
+          valueFrom:
+            secretKeyRef:
+              name: secret-name
+              key: AGENT_CONFIG
+      command: ["/noobaa_init_files/noobaa_init.sh", "agent"]
+      # Insert the relevant image for the agent
+      ports:
+        # This should change according to the allocation from the NooBaa server
+        - containerPort: 60101
+      volumeMounts:
+        - name: noobaastorage
+          mountPath: /noobaa_storage
+        - name: tmp-logs-vol
+          mountPath: /usr/local/noobaa/logs
+  volumes:
+    - name: tmp-logs-vol
+      emptyDir: {}
+    - name: noobaastorage
+      persistentVolumeClaim:
+        claimName: noobaa-pv-claim
--- a/deploy/internal/pvc-agent.yaml
+++ b/deploy/internal/pvc-agent.yaml
@@ -1,0 +1,12 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    app: noobaa
+  name: noobaa-pv-claim
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 30Gi
--- a/pkg/apis/noobaa/v1alpha1/backingstore_types.go
+++ b/pkg/apis/noobaa/v1alpha1/backingstore_types.go
@@ -241,6 +241,10 @@ type PVPoolSpec struct {
 
 	// VolumeResources represents the minimum resources each volume should have.
 	VolumeResources *corev1.ResourceRequirements `json:"resources,omitempty"`
+
+	// Secret refers to a secret that provides the agent configuration
+	// The secret should define AGENT_CONFIG containing agent_configuration from noobaa-core.
+	Secret corev1.SecretReference `json:"secret"`
 }
 
 // S3SignatureVersion specifies the client signature version to use when signing requests.
--- a/pkg/backingstore/backingstore.go
+++ b/pkg/backingstore/backingstore.go
@@ -40,6 +40,7 @@ func Cmd() *cobra.Command {
 		CmdStatus(),
 		CmdList(),
 		CmdReconcile(),
+		CmdRunRemovePendingPods(),
 	)
 	return cmd
 }
@@ -221,6 +222,16 @@ func CmdStatus() *cobra.Command {
 		Use:   "status <backing-store-name>",
 		Short: "Status backing store",
 		Run:   RunStatus,
+	}
+	return cmd
+}
+
+// CmdRunRemovePendingPods returns a CLI command
+func CmdRunRemovePendingPods() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "remove_pending <backing-store-name>",
+		Short: "Deletes all the pending pods that failed to connect to server",
+		Run:   RunRemovePendingPods,
 	}
 	return cmd
 }
@@ -478,8 +489,12 @@ func RunCreatePVPool(cmd *cobra.Command, args []string) {
 			NumVolumes:   int(numVolumes),
 			VolumeResources: &corev1.ResourceRequirements{
 				Requests: corev1.ResourceList{
-					corev1.ResourceStorage: *resource.NewScaledQuantity(int64(pvSizeGB), resource.Giga),
+					corev1.ResourceStorage: *resource.NewQuantity(int64(pvSizeGB)*1024*1024*1024, resource.BinarySI),
 				},
+			},
+			Secret: corev1.SecretReference{
+				Name:      secret.Name,
+				Namespace: secret.Namespace,
 			},
 		}
 	})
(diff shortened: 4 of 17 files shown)


pkg/apis/noobaa/v1alpha1/backingstore_types.go
```diff
@@ -244,6 +244,9 @@ type PVPoolSpec struct {
 
 	// VolumeResources represents the minimum resources each volume should have.
 	VolumeResources *corev1.ResourceRequirements `json:"resources,omitempty"`
+
```
dannyzaken, line 247: I think that agent_config should be stored in a secret, similar to cloud credentials in other backingstore types

deploy/internal/pvc-agent.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  labels:
+    app: noobaa
+  name: noobaa-pv-claim
+spec:
+  storageClassName: standard
```
dannyzaken, line 8: I'm not sure that `standard` is a predifined storage class. If you just ommit the storageclass I think the default stoarge class (defined in the cluster) will be used

https://kubernetes.io/docs/concepts/storage/storage-classes/#the-storageclass-resource

pkg/apis/noobaa/v1alpha1/backingstore_types.go
jackyalbo: changed to secret

deploy/internal/pod-agent.yaml
```diff
@@ -0,0 +1,41 @@
...
+          memory: "200Mi"
+        limits:
+          cpu: "100m"
+          memory: "200Mi"
+      env:
+        # Insert the relevant config for the current agent
+        - name: CONTAINER_PLATFORM
+          value: KUBERNETES
+        - name: AGENT_CONFIG
+          value: "AGENT_CONFIG_VALUE"
```
dannyzaken, line 25: also use `valueFrom` and `secretKeyRef` inside the pod. that way you don't expose the secret in the pod CR.

```suggestion
        - name: AGENT_CONFIG
           valueFrom:
             secretKeyRef:
               name: secret-name
               key: AGENT_CONFIG
```

pkg/backingstore/reconciler.go
```diff
@@ -761,18 +793,28 @@ func (r *Reconciler) ReconcileExternalConnection() error {
...
+			}
+		}
+		return r.reconcilePvPool()
 	}
 
 	poolName := ""
 
 	if r.CreateHostsPoolParams != nil {
-		err := r.NBClient.CreateHostsPoolAPI(*r.CreateHostsPoolParams)
+		res, err := r.NBClient.CreateHostsPoolAPI(*r.CreateHostsPoolParams)
```
dannyzaken, line 809: same as before. I think that it's better if the actual values are being read by noobaa-core directly from the cluster.

pkg/backingstore/reconciler.go
```diff
@@ -795,3 +837,116 @@ func (r *Reconciler) ReconcilePool() error {
 
 	return nil
 }
+
+func (r *Reconciler) reconcilePvPool() error {
```
dannyzaken, line 841: several issues I see with `reconcilePvPool`:

1. why isn't scale down supported? it was supported before
2. the naming convention for the pods and pvc can cause problems once we will support scale down.   
when you use sequential names (backingstore-1, backingstore-2, etc.) and you need to scale down then you are forced to remove the last one so you can keep the naming order. then if the user wants to scale back up before the previous scale down ended, it's becoming complicated to keep track of the state of previous scale down

deploy/internal/pod-agent.yaml
jackyalbo: using secret - fixed

pkg/backingstore/reconciler.go
jackyalbo: changing naming convention - scale down only to the number of already connected hosts in noobaa

pkg/backingstore/reconciler.go
```diff
@@ -761,18 +793,28 @@ func (r *Reconciler) ReconcileExternalConnection() error {
 // ReconcilePool handles the pool using noobaa api
 func (r *Reconciler) ReconcilePool() error {
 
-	// TODO we only support creation here, but not updates
 	if r.PoolInfo != nil {
-		return nil
+		if r.ScaleHostsPoolParams != nil {
+			err := r.NBClient.ScaleHostsPoolAPI(*r.ScaleHostsPoolParams)
```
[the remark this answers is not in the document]
jackyalbo: using new update api

pkg/backingstore/reconciler.go
jackyalbo: for now as discussed we will keep - as 2 different creates suppose to be impossible

pkg/backingstore/backingstore.go
```diff
@@ -530,6 +545,66 @@ func RunDelete(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		rpcErr, isRPCErr := err.(*nb.RPCError)
+		if !isRPCErr || rpcErr.RPCCode != "NO_SUCH_POOL" {
+			log.Fatalf(`❌ Failed to read BackingStore host info: %s`, err)
+		}
+	}
+	podsList := &corev1.PodList{}
+	util.KubeList(podsList, client.InNamespace(options.Namespace), client.MatchingLabels{"pool": backStore.Name})
+	for _, pod := range podsList.Items {
+		if !isPodinNoobaa(&pod, &hostsInfo.Hosts) {
```
dannyzaken, line 580: Maybe I'm being too strict but there is a theoretical race condition here. 

between the time of `listsHosts` and `kubeDelete(&pod)` the agent might already be connected.

a possible solution can be to recheck the hosts after the pod is deleted and before the PVC is deleted. if the pod became registered in noobaa then you can just deploy the pod again.

it might not be critical to fix right now but we should at least have a technical debt regarding this.

pkg/backingstore/backingstore.go
```diff
@@ -530,6 +545,66 @@ func RunDelete(cmd *cobra.Command, args []string) {
...
+// RunRemovePendingPods runs a CLI command
+func RunRemovePendingPods(cmd *cobra.Command, args []string) {
+	log := util.Logger()
+
+	if len(args) != 1 || args[0] == "" {
+		log.Fatalf(`❌ Missing expected arguments: <backing-store-name> %s`, cmd.UsageString())
+	}
+
+	o := util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_backingstore_cr_yaml)
+	backStore := o.(*nbv1.BackingStore)
```
dannyzaken, line 557: how will the backingstore reconciler will behave once you are messing with its pods? also, don't you need to update the numVolumes?

pkg/backingstore/backingstore.go
jackyalbo: We want this command to be running on stable system. User asks for 10 see that he got 3 and can't get more. he will change the NumVolumes to what he see as the max. and just to cleanup after himself he runs this command. This command will be removed as soon as we have scale down.

pkg/backingstore/backingstore.go
jackyalbo: changing the numVolumes won't help as the reconcile will start running during the deletion of the pods and try to recreate them. So we need for him to first change the numVolumes wait for it to stable. and then run this command

pkg/backingstore/backingstore.go
dannyzaken: you can never guarantee a stable system. for example some cpu\mem can become available (an unrelated pod got evicted) and a pod that was not attached becomes attached

dannyzaken:
I think it is the wrong approach to fix the count in a CLI command
if the user will decrease the numVolumes then his backingstore will be in a persistent error state. I think it makes the behavior even more confusing for the user.

in my opinion this function should be part of the reconcile loop. maybe the user can set a flag on the CR to `cleanPendingPods` and the reconcile will perform this code. the CLI command can just set this flag

jackyalbo:
> what are we doing with pv-pools that were created and managed by noobaa-core (old implementation)?
> 
> at the moment, if there is an old backingstore with a statefulset pv-pool, after upgrading to this version the operator will create a new pool manged by the operator, and the old statefulset pods will keep running. am I right?

We don't create new pool. We are deleting the sts, and moving all of his pods to be under the backingstore directly, sadly with the old naming convention. Scaling up this pvpool will add new pods with the new naming convention, but other than the names, everything else should be the same.

pkg/backingstore/reconciler.go
```diff
@@ -795,3 +862,180 @@ func (r *Reconciler) ReconcilePool() error {
...
+		}
+	}
+	util.KubeDeleteAllOf(&corev1.Pod{}, client.InNamespace(options.Namespace), client.MatchingLabels{"pool": r.BackingStore.Name})
+	util.KubeDeleteAllOf(&corev1.PersistentVolumeClaim{}, client.InNamespace(options.Namespace), client.MatchingLabels{"pool": r.BackingStore.Name})
+	return nil
+}
+
+func (r *Reconciler) upgradeBackingStore(sts *appsv1.StatefulSet) error {
+	r.Logger.Infof("Deleting old statefulset: %s", sts.Name)
+	agentConfig := sts.Spec.Template.Spec.Containers[0].Env[1].Value
```
dannyzaken, line 1003: will it always be `Env[1]`? maybe it's better to find by name. 
you can add a util function to find Env by name

pkg/backingstore/reconciler.go
```diff
@@ -795,3 +862,180 @@ func (r *Reconciler) ReconcilePool() error {
...
+	secret.Name = fmt.Sprintf("backing-store-%s-%s", nbv1.StoreTypePVPool, r.BackingStore.Name)
+	secret.Namespace = options.Namespace
+	util.KubeCheck(secret)
+	secret.StringData["AGENT_CONFIG"] = agentConfig // update secret for future pods
+	util.KubeUpdate(secret)
+	for i := 0; i < int(*replicas); i++ {
+		pod := &corev1.Pod{}
+		pod.Name = fmt.Sprintf("%s-%d", stsName, i)
+		pod.Namespace = r.BackingStore.Namespace
+		util.KubeCheck(pod)
```
dannyzaken, line 1018: `replicas` can be different from the actual number of pods. check that the pod\PVC really exist

jackyalbo: Yeah I will put the kubeCheck in If... good catch

pkg/backingstore/reconciler.go
jackyalbo: add function to util

pkg/util/util.go
```diff
@@ -1023,3 +1040,15 @@ func WriteYamlFile(name string, obj runtime.Object, moreObjects ...runtime.Objec
...
+
+// GetEnvVariable is looknig for env variable called name in env and return a pointer to the variable
+func GetEnvVariable(env *[]corev1.EnvVar, name string) *corev1.EnvVar {
+	for i := range *env {
+		e := (*env)[i]
+		if e.Name == name {
+			return &e
+		}
+	}
+	log.Printf("❌ Not Found Env variable: %s\n", name)
```
dannyzaken, line 1052: would remove the log here. sometimes it is expected to not find an env and the logs are just annoying. let the caller decide if it should be logged

PR #288, opened 2020-04-22
Travis: fix to quit immediately after the first script fails

liranmauda:
.travis.yml:

- Added || exit 1 on all scripts calls

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,8 +24,8 @@ install:
   - bash .travis/install-minikube.sh
 
 script:
-  - make gen-api-fail-if-dirty --always-make
-  - make build
-  - make test
-  - make test-olm
-  - make test-cli-flow
+  - make gen-api-fail-if-dirty --always-make || exit 1
+  - make build || exit 1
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -24,8 +24,8 @@ install:
   - bash .travis/install-minikube.sh
 
 script:
-  - make gen-api-fail-if-dirty --always-make
-  - make build
-  - make test
-  - make test-olm
-  - make test-cli-flow
+  - make gen-api-fail-if-dirty --always-make || exit 1
```
guymguym, line 27: What went wrong - doesn't make also return 1 when failing?
Or maybe its because the travis script needs explicit exit to be called.

nimrod-becker: it keeps on running unless we exit

liranmauda: Travis will run all the scripts sequentially (it is travis design) even if it fails.
the way to tell it not to can be with `|| exit 1` or `|| travis_terminate 1`

PR #290, opened 2020-04-27
automatically download the correct operator-sdk version

dannyzaken:


.travis/install-operator-sdk.sh
```diff
@@ -1,8 +1,3 @@
 #!/bin/bash
-SDK_VERSION="v0.16.0"
-SDK_RELEASE="https://github.com/operator-framework/operator-sdk/releases/download/${SDK_VERSION}/operator-sdk-${SDK_VERSION}-x86_64-linux-gnu"
-
-echo "install-sdk - installing version ${SDK_VERSION}"
-curl "${SDK_RELEASE}" -Lo $GOPATH/bin/operator-sdk
-chmod +x $GOPATH/bin/operator-sdk
-$GOPATH/bin/operator-sdk version
+bash build/install-operator-sdk.sh && \
+cp build/_tools/operator-sdk ${GOPATH}/bin/operator-sdk
```
liranmauda, line 3: add new line

.travis/install-operator-sdk.sh
```diff
@@ -1,8 +1,3 @@
 #!/bin/bash
-SDK_VERSION="v0.16.0"
-SDK_RELEASE="https://github.com/operator-framework/operator-sdk/releases/download/${SDK_VERSION}/operator-sdk-${SDK_VERSION}-x86_64-linux-gnu"
-
-echo "install-sdk - installing version ${SDK_VERSION}"
-curl "${SDK_RELEASE}" -Lo $GOPATH/bin/operator-sdk
-chmod +x $GOPATH/bin/operator-sdk
-$GOPATH/bin/operator-sdk version
+bash build/install-operator-sdk.sh && \
```
liranmauda, line 2: I think that we don't need to specify bash  

```suggestion
build/install-operator-sdk.sh && \
```

guymguym: @liranmauda why not? are we targeting another shell?

dannyzaken: any reason to not specify bash?

liranmauda: @guymguym using shebang line tells that this interpreter for the shell is bash.

dannyzaken:
@guymguym 👍 I like that

.travis/install-operator-sdk.sh
guymguym: I know it will work. OK it's a super duper minor point. I bet each of us will change his mind about which style is better. My argument for now is that we don't want to allow various shell support in our code base, and this code style supports this fact.

dannyzaken:
@guymguym regarding your suggestion - currently we are calling operator-sdk from `test-olm.sh`, and counting on having it in the PATH. any idea of how we can get around it?

guymguym:
@dannyzaken set `OPERATOR-SDK` env from makefile to test-olm 
if not provided use default `operator-sdk` binary from path...

build/install-operator-sdk.sh
```diff
@@ -0,0 +1,39 @@
...
+        exit 0
+    fi
+fi
+
+PLATFORM=$(uname)
+if [ "${PLATFORM}" == "Darwin" ] 
+then 
+    SDK_RELEASE="https://github.com/operator-framework/operator-sdk/releases/download/${OPERATOR_SDK_VERSION}/operator-sdk-${OPERATOR_SDK_VERSION}-x86_64-apple-darwin"
+else 
+    # Assuming that if not darwin then running on linux
```
mykaul, line 28: Well, uname on Linux returns 'Linux'.

nimrod-becker: Which also gets into the else, do you see a problem here?
We don't test compiling over WIN, but that's the only issue I see with a generic else ....

mykaul: I don't see a problem, was just stating that instead of assuming we can use 'Linux' just as we test for 'Darwin', that's all.

PR #299, opened 2020-04-30
Addition of Azure default BackingStore

jeniawhite:
Added default backing store on azure using the cloud credentials operator.
Gaps:
Handle deletion of resources (storage account, container, the secret that saves the info)


pkg/system/phase4_configuring.go
```diff
@@ -749,3 +833,166 @@ func (r *Reconciler) ReconcileDeploymentEndpointStatus() error {
 
 	return nil
 }
+
```
dannyzaken, line 836: looks like there are quite a lot of azure specific functions. maybe consider moving those to a different file (`azure_utils` or something) in `system` package

PR #307, opened 2020-05-20
added SecurityContext to mongo statefulset to with a know uid

dannyzaken:
* added SecurityContext to mongo statefulset to with a know uid
   * solves permissions error in Mongo when SCC is changed in openshift to `anyuid`
* changed /mongo_data mountPath to /data in init container
   * to be consistent with DB container

Fixing BZ #1835125


dannyzaken:
closed. this fix doesn't work without setting SCC that matches the SecurityContext request. requires a more complete fix

PR #309, opened 2020-05-21
Added graceful uninstall support 

romayalon:
### Explain the changes
1. Added support for graceful uninstall.
2. Added a check in system reconcile  - if  the system has deletionTimestamp and if graceful_finalizer exist and num of object buckets == 0, remove the graceful_finalizer. else, return error.
3. Added a check in Provision and Grant - if the system has deletionTimestamp and it has graceful_finalizer - return error.

1. need to create a PR for ocs-operator that adds graceful_finalizer to noobaa.

### Testing Instructions:
1.

### DIFF
--- a/pkg/system/reconciler.go
+++ b/pkg/system/reconciler.go
@@ -26,6 +26,7 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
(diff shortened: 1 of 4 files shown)


pkg/system/reconciler.go
```diff
@@ -286,6 +291,45 @@ func (r *Reconciler) Reconcile() (reconcile.Result, error) {
...
+				util.RemoveFinalizer(r.NooBaa, nbv1.GracefulFinalizer)
+				if !util.KubeUpdate(r.NooBaa) {
+					log.Errorf("NooBaa %q failed to remove finalizer %q", r.NooBaa.Name, nbv1.GracefulFinalizer)
+				}
+
+			} else {
+				var bucketNames []string
+				for i := range objectBuckets.Items {
+					ob := &objectBuckets.Items[i]
+					bucketName := ob.Spec.Endpoint.BucketName
```
jackyalbo, line 320: Why can't we print the ob.Name? it's not clearer for the user what's not deleted yet?

romayalon: it is printed in line 323 in an array of all the object bucket that were not deleted yet

jackyalbo: but why the name of the acutal buckets and not just ob.Name? that's what I'm missing...

romayalon: I checked and it's not the same name. After a discussion with Jacky, we agreed to change it to ob.Name.

PR #313, opened 2020-05-26
Reconcile noobaa pods HTTP* envs based on operator pod HTTP* envs

nb-ohad:
When running in a proxied environment the HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables control which of a pod's outgoing traffic is directed to a proxy and which not.

When cluster-wide proxy setting is used OLM update the operator's env variables to the proper values. In turn, the operator should set the environment variables on the resources that it is responsible for.

This PR update the operator to read the values of the HTTP* environment variables set on the operator container and set them for the `noobaa-core` statefulset and `nobaa-endpoint` deployments

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1819549

Edit: 
We have a gap where changes to the env vars will not be reconciled for pv pool agent pods.
This is because a pod spec cannot be changed and we need takedown the pod and start a new one which we currently do not support with pv pool pods.

### DIFF
--- a/deploy/internal/pod-agent.yaml
+++ b/deploy/internal/pod-agent.yaml
@@ -24,10 +24,6 @@ spec:
         - name: CONTAINER_PLATFORM
           value: KUBERNETES
         - name: AGENT_CONFIG
-          valueFrom:
-            secretKeyRef:
-              name: secret-name
-              key: AGENT_CONFIG
       command: ["/noobaa_init_files/noobaa_init.sh", "agent"]
       # Insert the relevant image for the agent
       ports:
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	math "math"
 	"net/url"
+	"os"
 	"regexp"
 	"strings"
 	"time"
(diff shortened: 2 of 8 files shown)


pkg/system/phase2_creating.go
```diff
@@ -258,6 +258,21 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 					if r.OAuthEndpoints != nil {
 						c.Env[j].Value = r.OAuthEndpoints.TokenEndpoint
 					}
+
```
dannyzaken, line 261: if the proxies envs are not set on the operator pod then core\endpoints are deployed with empty vars (just names and no value). 
does that work?

nb-ohad: yes it does work

pkg/backingstore/reconciler.go
```diff
@@ -37,22 +37,22 @@ func init() {
...
-		"OPTIMAL":             nbv1.BackingStorePhaseInfo{nbv1.BackingStorePhaseReady, "BackingStorePhaseReady", "Backing store mode: OPTIMAL"},
-		"HAS_NO_NODES":        nbv1.BackingStorePhaseInfo{nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: HAS_NO_NODES"},
-		"ALL_NODES_OFFLINE":   nbv1.BackingStorePhaseInfo{nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: ALL_NODES_OFFLINE"},
-		"NO_CAPACITY":         nbv1.BackingStorePhaseInfo{nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: NO_CAPACITY"},
-		"IO_ERRORS":           nbv1.BackingStorePhaseInfo{nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: IO_ERRORS"},
-		"STORAGE_NOT_EXIST":   nbv1.BackingStorePhaseInfo{nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: STORAGE_NOT_EXIST"},
-		"AUTH_FAILED":         nbv1.BackingStorePhaseInfo{nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: AUTH_FAILED"},
+		"INITIALIZING":        {nbv1.BackingStorePhaseReady, "BackingStorePhaseReady", "Backing store mode: INITALIZING"},
```
dannyzaken, line 40: ?? 
is that just a formatter thing? if it is we must align on the same rules so we won't have irrelevant changes in PRs

deploy/internal/pod-agent.yaml
```diff
@@ -24,10 +24,9 @@ spec:
         - name: CONTAINER_PLATFORM
           value: KUBERNETES
         - name: AGENT_CONFIG
-          valueFrom:
```
dannyzaken, line 27: why did you remove that?

pkg/system/phase2_creating.go
dannyzaken: I think there is a subtle difference between empty vars and non-existing vars. I am not an expert on that, but I think it's better to forward the correct value. you can use `LookupEnv` to distinguish between the cases.

from the documentation of Getenv:
```
// Getenv retrieves the value of the environment variable named by the key.
// It returns the value, which will be empty if the variable is not present.
// To distinguish between an empty value and an unset value, use LookupEnv.
```

pkg/backingstore/reconciler.go
```diff
@@ -890,6 +890,26 @@ func (r *Reconciler) reconcileExistingPods(podsList *corev1.PodList) error {
 	noneAttachingAgents := 0
 	failedAttachingAgents := 0
 	for _, pod := range podsList.Items {
+		for j := range pod.Spec.Containers.Containers[0].Env {
+			switch c.Env[j].Name {
+			case "HTTP_PROXY":
+				value := os.Getenv("HTTP_PROXY")
+				if value != c.Env[j].Value {
```
dannyzaken, line 897: not sure I understand the condition. you should use `LookupEnv`to set the correct value

pkg/backingstore/reconciler.go
nimrod-becker: I agree, Ohad please revert

deploy/internal/pod-agent.yaml
nb-ohad: Jacky's code is overwriting the entire env not just replacing a value. This declaration has no meaning and is actually misleading. I talked with him and we decided to remove this.

pkg/backingstore/reconciler.go
nb-ohad: The new GoFmt is removing unnecessary type declaration.  We all use the same gofmt

nb-ohad:
@dannyzaken Why? This is the way we handle all of our env, even the ones that are not used in some circumstances. I see no benefit in doing it the other way around.  Declaring them ahead of time give the reader a sense that understanding of what we are going to reconcile. 
And functionally it behaves the same.

pkg/backingstore/reconciler.go
nb-ohad: No really, `os.GetEnv` will return an empty string for empty or non-existing envs. This condition was set just because Jacky didn't want me to reset the envs in the case where the value didn't change, Which I understand is pointless, resetting to the same value will give the same result.

dannyzaken:
@nb-ohad are you sure it behaves the same? these envs are not used by noobaa, you are just propagating them from the operator to other pods. I think that in this case, you should forward them as is

pkg/backingstore/reconciler.go
nb-ohad: Just an update. It's not the GoFmt version. It the gofmt settings. My editor is running gofmt with the simplify code flag while vscode does not (by default). I will try to check if I can disable the simplify behavior to be more consistent but as a rule of thumb I think this behavior can benefit us

nb-ohad:
@all 
Please read the update at the end of PR description

pkg/backingstore/reconciler.go
```diff
@@ -890,6 +890,20 @@ func (r *Reconciler) reconcileExistingPods(podsList *corev1.PodList) error {
...
+
+		// GAP: reconcile proxy env vars.
+		// The only thing that can be changed on a pod spec is the image
+		// trying to change anything else, including env vars, will create a panic.
+		// Deployment and Statefulset are handling this by taking down the pod and
+		// starting a new one.
+		// In order to support reconciling changes to the proxy env we need to mimic
+		// this behaviour which is not trival in the case of agent pods.
+
+		// var c = &pod.Spec.Containers[0]
```
dannyzaken, line 902: for now, maybe log if there is a mismatch between the pod and ther operator envs

nb-ohad: done

PR #318, opened 2020-06-01
Added to NooBaa CLI new install and uninstall flags

romayalon:
1. Added cleanup_data flag for uninstall. From now on every uninstallation of noobaa will require the user to cleanup his OBs/OBCs by him self / use the new cleanup_data flag that will remove forcefully all his OBs/OBCs.
2. Added use_graceful_finalizer flag for install. by default, noobaa will have graceful_finalizer, by using the use_graceful_flag=false, the finalizer will be removed.

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -22,6 +22,7 @@ func CmdInstall() *cobra.Command {
 		Short: "Install the operator and create the noobaa system",
 		Run:   RunInstall,
 	}
+	cmd.Flags().Bool("use-obc-cleanup-policy", false, "Create NooBaa system with obc cleanup policy")
 	return cmd
 }
 
@@ -33,6 +34,7 @@ func CmdUninstall() *cobra.Command {
 		Run:   RunUninstall,
 	}
 	cmd.Flags().Bool("cleanup", false, "Enable deletion of Namespace and CRD's")
+	cmd.Flags().Bool("cleanup_data", false, "Clean object buckets")
 	return cmd
 }
 
--- a/test/cli/test_cli_functions.sh
+++ b/test/cli/test_cli_functions.sh
@@ -119,7 +119,11 @@ function timeout {
 }
 
 function install {
-    test_noobaa install --mini
+    local use-obc-cleanup-policy
+    
+    [ $((RANDOM%2)) -gt 0 ] && use-obc-cleanup-policy="--use-obc-cleanup-policy"
+    test_noobaa install --mini ${use-obc-cleanup-policy}
+
     local status=$(kuberun get noobaa noobaa -o json | jq -r '.status.phase' 2> /dev/null)
     while [ "${status}" != "Ready" ]
     do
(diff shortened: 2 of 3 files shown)


pkg/install/install.go
```diff
@@ -22,6 +22,7 @@ func CmdInstall() *cobra.Command {
 		Short: "Install the operator and create the noobaa system",
 		Run:   RunInstall,
 	}
+	cmd.Flags().Bool("use_graceful_finalizer", true, "Create NooBaa system witout graceful finalizer")
```
liranmauda, line 25: No 100% sure, but setting the default value as true can be misleading.
As far as I understand using flags (like use_graceful_finalizer) will set the value to true and in order to reverse it one needs to use `-use_graceful_finalizer=flase`
I think it is more straight forward to use a default bool flag as false.
also, the flag and the description are confusing to me as the flag is "use" and the description is "without"

What do you think?

romayalon: Changed the flag's name to skip_graceful_flag and reversed the default value to false
 :)

romayalon:
Hey @obnoxxx :)
There is a misunderstanding, all the uninstallation support is in https://github.com/noobaa/noobaa-operator/pull/309/files

in PR #309 we are actually waiting for the OBs/OBCs removal (support graceful mode uninstallation) before removing the graceful_finalizer.
As I talked to Talur, we would like him to remove the graceful_finalizer in Forced mode.

in this PR:
cleanup_data flag is a new flag for noobaa uninstall in noobaa CLI 
also skip_graceful_finalizer is a new flag for noobaa install in noobaa CLI.

I'm sorry if it wasn't clear I meant to the CLI options.

obnoxxx:
@romayalon  - thanks for the clarification. I am reviewing PR #309 now (after the fact.) So the discussion about graceful vs forceful mode should not continue here but elsewhere. I will switch to chat for now. :-)

pkg/install/install.go
```diff
@@ -22,6 +22,7 @@ func CmdInstall() *cobra.Command {
 		Short: "Install the operator and create the noobaa system",
 		Run:   RunInstall,
 	}
+	cmd.Flags().Bool("skip_graceful_finalizer", false, "Create NooBaa system witout graceful finalizer")
```
guymguym, line 25: style: how about renaming it to `no_graceful_finalizer` mainly because it's hard to imagine what does skipping over a graceful finalizer actually means - sounds like poetry though...

romayalon: accepted :)

test/cli/test_cli_functions.sh
```diff
@@ -334,15 +342,30 @@ function crd_arr {
...
+        if [ ${check_cleanup_data_flag} -gt 0 ] 
+        then
+            echo_time "Running uninstall without --cleanup and with --cleanup_data”"
+            test_noobaa uninstall --cleanup_data
+        else
+            echo_time "Running uninstall without --cleanup and without --cleanup_data”"
+            test_noobaa uninstall 
+        fi 
     fi  
 }
```
liranmauda, line 370: We can change the function to be like this:
```
function noobaa_uninstall {
    local cleanup cleanup_data
    local check_cleanflag=$((RANDOM%2))
    local check_cleanup_data_flag=$((RANDOM%2))

    [ ${check_cleanflag} -eq 0 ] &&  cleanup="--cleanup"
    [ ${check_cleanup_data_flag} -gt 0 ] && cleanup_data="--cleanup_data"

    echo_time "Running uninstall ${cleanup} ${cleanup_data}"
    test_noobaa uninstall ${cleanup} ${cleanup_data}
    if [ ${check_cleanflag} -eq 0 ]
    then
        check_if_cleanup
    fi
}
```

What do you think?

test/cli/test_cli_functions.sh
```diff
@@ -119,7 +119,15 @@ function timeout {
...
 function install {
-    test_noobaa install --mini
+    use_obc_cleanup_flag=$((RANDOM%2))
+    
+    if [ ${use_obc_cleanup_flag} -gt 0 ] 
+    then
+        test_noobaa install --mini --use-obc-cleanup-policy
+    else      
+        test_noobaa install --mini
+    fi
```
liranmauda, line 129:
```suggestion
    local use-obc-cleanup-policy
    
   [ $((RANDOM%2)) -gt 0 ] && use-obc-cleanup-policy="--use-obc-cleanup-policy"
   test_noobaa install --mini ${use-obc-cleanup-policy}
```
Not sure how readable this suggestion, what do you think?

romayalon: I think it's great :) i'll change it

PR #323, opened 2020-06-03
Use the GracefulFinalizer constant

obnoxxx:
Some places uses the string literal where it could have used the
constant. This lets all places use the constant.

Follow-up from reviewing PR #309 

Signed-off-by: Michael Adam <obnox@redhat.com>


obnoxxx:
@romayalon @dannyzaken @nimrod-becker fyi. just a trivial small cleanup

Issue #325, opened 2020-06-04
make requires a valid kubeconfig


guymguym:
CC @obnoxxx

If we stop minikube make results in:
```
$ make
go mod tidy
go mod vendor
✅ vendor
✅ gen
build/_tools/operator-sdk-v0.17.0 run --local --operator-flags "version"
Error: error getting kubeconfig and default namespace: invalid configuration: no configuration has been provided
Usage:
  operator-sdk run [flags]
...
make: *** [Makefile:42: cli] Error 1
```

This happens because we use `operator-sdk run --local --operator-flags "version"` to build just the CLI - see here:
https://github.com/noobaa/noobaa-operator/blob/1cf4cbabde9a6ab4024421054e8ee5e48735b704/Makefile#L41-L44

We could have called `go build ...` directly but preferred to have operator-sdk call the go build in the same manner it always does. Perhaps this was changed with newer operator-sdk that it now looks for kubeconfig on its own.

However an actual running cluster is not required - it works even if I just use this valid temp kubeconfig:
```
$ export KUBECONFIG=empty-kubeconfig

$ cat empty-kubeconfig
apiVersion: v1
kind: Config
preferences: {}
current-context: temp
clusters:
- name: temp
  cluster:
    server: temp
users:
- name: temp
contexts:
- name: temp
  context:
    cluster: temp
    user: temp
    namespace: temp

$ make cli
checking operator-sdk version
operator-sdk version | grep -q "\"v0.16.0\", commit: \"55f1446c5f472e7d8e308dcdf36d0d7fc44fc4fd\""
✅ operator-sdk
go mod tidy
go mod vendor
✅ vendor
✅ gen
operator-sdk run --local --operator-flags "version"
INFO[0000] Running the operator locally in namespace temp.
INFO[0000] CLI version: 2.3.0
INFO[0000] noobaa-image: noobaa/noobaa-core:5.4.0-rc1
INFO[0000] operator-image: noobaa/noobaa-operator:2.3.0
✅ cli
```

so we might want to ask operator-sdk to provide a command to just build the binary of the operator instead of our make cli calling operator-sdk run to achieve that, or find another way to avoid this failure when no valid kubeconfig is available during build.

obnoxxx:
@guymguym - thanks for capturing our discussion here!

If this empty kubeconfig is sufficient (and correct) to do the build, we could just add it to the `devenv.sh` script and thereby remove the dependency to minikube running.

PR #326, opened 2020-06-04
Don't require minikube for building.

obnoxxx:
`make cli` runs `openshift-sdk run ...` which requires a kubeconfig.
Up to now, this target relied on minikube being started.

This modifies `devenv.sh` to generate an empty kubeconfig file and set the
KUBECONFIG variable to avoid failure if minikube is not running.

Fixes: #325

Signed-off-by: Michael Adam <obnox@redhat.com>

### DIFF
--- a/build/empty-kubeconfig
+++ b/build/empty-kubeconfig
@@ -1,0 +1,22 @@
+#
+# This trival and invalid kubeconfig file has the purpose of allowing us to
+# build some targets using `operator-sdk run ...` without requiring a running
+# kubernetes environment.
+# It will not allow you to really run and use the operator.
+#
+apiVersion: v1
+kind: Config
+preferences: {}
+current-context: temp
+clusters:
+- name: temp
+  cluster:
+    server: temp
+users:
+- name: temp
+contexts:
+- name: temp
+  context:
+    cluster: temp
+    user: temp
+    namespace: temp
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,8 @@ VENV ?= $(OUTPUT)/venv
 export OPERATOR_SDK_VERSION ?= v0.17.0
 export OPERATOR_SDK ?= build/_tools/operator-sdk-$(OPERATOR_SDK_VERSION)
 
+KUBECONFIG ?= build/empty-kubeconfig
+
 #------------#
 #- Building -#
 #------------#
@@ -39,7 +41,7 @@ build: cli image gen-olm
 .PHONY: build
 
 cli: $(OPERATOR_SDK) gen
-	$(OPERATOR_SDK) run --local --operator-flags "version"
+	$(OPERATOR_SDK) run --kubeconfig="$(KUBECONFIG)" --local --operator-flags "version"
 	@echo "✅ cli"
 .PHONY: cli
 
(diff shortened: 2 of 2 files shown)


obnoxxx:
@guymguym - FYI, putting your solution into devenv. works for me. Not sure this is how one would really do it.

obnoxxx:
Possible solution for #325

devenv.sh
```diff
@@ -11,4 +11,27 @@ then
   eval $(minikube docker-env)
 else
   echo "WARNING: minikube is not started - cannot change docker-env"
+
+  echo "Generating empty kubeconfig file"
+  KUBECONFIG_FILE=./empty-kubeconfig
```
guymguym, line 16: Can you add it to .gitignore?

obnoxxx: Done (amended existing patch).

guymguym: @obnoxxx Oh I think my comment was too quick.

I have two changes to suggest:
1. Keep that empty-kubeconfig file in git under `build/`, and then devenv will just update the KUBECONFIG to point to it. The .gitignore change is uneeded for this.
2. I think we should prefer that devenv will not override existing KUBECONFIG unless there is really no configuration available, because it can interfere with working on remote clusters. Can we detect this case with `kubectl config view` - I am not sure what error it gives when the configuration is empty in your minikube stop case.

What do you think?

obnoxxx: @guymguym  - Sounds good. 

Question to clarify: Do you suggest to commit the empty kubeconfig file under `build/` or to generate there?

guymguym: I think to commit it makes more sense, since this is a static asset for building that we can also change if needed.

obnoxxx: Committing makes sense - will do that.

Regarding not overriding KUBECONFIG: checking for the variable is easy. But that might override a working minikube kube config, which is coming from `$HOME/.kube/config`.  But using `kubectl config view` does not seem to work:

The problem with `kubectl config view` is that it will spit out its config no matter whether the cluster is reachable or up. If minikube is stopped, it spits out the valid config. But the later commands triggered by `operator-sdk run` don't work if the cluster is not up.

Since we can build with the empty kubeconfig, the question would be: wouldn't it be fine to always build with this?

guymguym: Right. So we can pass this directly in the Makefile to the call `operator-sdk run --kubeconfig=build/empty-kubeconfig ...`

obnoxxx: I like it when things get more simple. :-)

obnoxxx: Updated. Only modified targets `cli` and `gen-olm`. `run` still requires a kubernetes env. The setting can still be overridden by the KUBCONFIG env variable. But building now works without.

build/empty-kubeconfig
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
```
guymguym, line 1: Would you write a short comment why this file is needed (for next generations)?

devenv.sh
guymguym: Magnificent.

build/empty-kubeconfig
obnoxxx: Done. Also rebased.

PR #334, opened 2020-06-15
WIP - Adding to the test crd flows

kerencarmelwedel:
crd function to check noobaa crd commands added to the test

### DIFF
--- a/test/cli/test_cli_flow.sh
+++ b/test/cli/test_cli_flow.sh
@@ -69,6 +69,7 @@ function main {
     check_S3_compatible
     bucketclass_cycle
     obc_cycle
+    crd_cycle
     check_deletes
     noobaa_uninstall
  }
--- a/test/cli/test_cli_functions.sh
+++ b/test/cli/test_cli_functions.sh
@@ -123,9 +123,9 @@ function install {
     local status=$(kuberun get noobaa noobaa -o json | jq -r '.status.phase' 2> /dev/null)
     while [ "${status}" != "Ready" ]
     do
-        echo_time "Waiting for status Ready, Status is ${status}"
-        sleep 10
-        status=$(kuberun get noobaa noobaa -o json | jq -r '.status.phase' 2> /dev/null)
+       echo_time "Waiting for status Ready, Status is ${status}"
+       sleep 10
+       status=$(kuberun get noobaa noobaa -o json | jq -r '.status.phase' 2> /dev/null)
     done
 }
 
@@ -234,7 +234,6 @@ function bucketclass_cycle {
     # test_noobaa bucketclass create ${bucketclass_names[1]} --placement Mirror --backingstores nb1,aws1 ❌
     # test_noobaa bucketclass create ${bucketclass_names[2]} --placement Spread --backingstores aws1,aws2 ❌
     test_noobaa bucketclass create ${bucketclass_names[3]} --backingstores ${backingstore[0]},${backingstore[1]}
-
     local bucketclass_list_array=($(test_noobaa silence bucketclass list | awk '{print $1}' | grep -v NAME))
     for bucketclass in ${bucketclass_list_array[@]}
     do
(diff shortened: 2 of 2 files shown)


test/cli/test_cli_functions.sh
```diff
@@ -284,10 +284,89 @@ function obc_cycle {
...
+    echo_time "checking if crd noobaa create command equals to crd status command"
+    for crd_status_array in ${crd_array[@]}
+    do
+        if [[ ${crd_create_array[@]} =~ ${crd_status_array} ]]
+        then            
+            echo_time "✅ ${crd_status_array} exists in the crds list"
+        else
+            echo_time "❌ ${crd_status_array} is not existed in the crds list"         
+        fi               
+    done
```
liranmauda, line 303: If the crd is not exists, why not exit?

test/cli/test_cli_functions.sh
```diff
@@ -284,10 +284,89 @@ function obc_cycle {
...
 
+function crd_cycle {     
+    echo_time "Starting the crd cycle" 
+    local crd_create_array=($(test_noobaa silence crd create &>/dev/stdout | awk '{print $6}' | sed -e 's/[""\\]//g' | sed 's/.$//'))
+    local crd_status_array=($(test_noobaa silence crd status &>/dev/stdout | awk '{print $6}' | grep -v "Exists" | sed -e 's/[""\\]//g' | sed 's/.$//'))
+    crd_array=($(kubectl get crd | awk '{print $1}' | grep -v "NAME"))
+    
+    #comparing crd status and create arrays
+    echo_time "checking if crd noobaa create command equals to crd status command"
+    for crd_status_array in ${crd_array[@]}
```
liranmauda, line 295: This is not doing what the `echo_time` describes. 
it is checking that the crd status command is equal to the kubectl get crd command output.

test/cli/test_cli_functions.sh
```diff
@@ -284,10 +284,89 @@ function obc_cycle {
...
+    done
+         
+    kuberun get crd  
+    if [ $? -ne 0 ]
+    then
+        echo_time "❌ kubectl get crd failed"
+        exit 1 
+    else 
+        echo_time "✅ kubectl get crd passed"    
+    fi
```
liranmauda, line 312: This is checking the kubrun function exit code.
why do we need to check kubrun/kubectl?

test/cli/test_cli_functions.sh
```diff
@@ -284,10 +284,89 @@ function obc_cycle {
...
+        fi
+    done
+    kuberun get crd  
+    if [ $? -ne 0 ]
+    then
+        echo_time "❌ kubectl get crd failed"
+        exit 1 
+    else 
+        echo_time "✅ kubectl get crd passed"    
+    fi
```
liranmauda, line 334: This is checking the kubrun function exit code.
why do we need to check kubrun/kubectl?

test/cli/test_cli_functions.sh
```diff
@@ -284,10 +284,89 @@ function obc_cycle {
...
+        exit 1 
+    else 
+        echo_time "✅ kubectl get crd passed"    
+    fi 
+
+    echo_time "creating crd again for checks..."     
+    test_noobaa timeout crd create 
+    local crd_after_create=($(test_noobaa silence crd status &>/dev/stdout | awk '{print $6}' | sed -e 's/[""\\]//g' | sed 's/.$//'))
+    #comparing crds after running create command
+    for crd_after_create in ${crd_array[@]}
```
liranmauda, line 340: it runs on the crd that we get using kubectl get crd command. it includes non-noobaa crds.

test/cli/test_cli_functions.sh
```diff
@@ -284,10 +284,89 @@ function obc_cycle {
...
+    done
+
+    kuberun get crd  
+    if [ $? -ne 0 ]
+    then
+        echo_time "❌ kubectl get crd failed"
+        exit 1 
+    else 
+        echo_time "✅ kubectl get crd passed"    
+    fi
```
liranmauda, line 357: Same as above

test/cli/test_cli_functions.sh
```diff
@@ -284,10 +284,89 @@ function obc_cycle {
...
+       
+    echo_time "✅  crd cycle is done"
+}
+
 function delete_backingstore_path {
     local object_bucket backing_store
     local backingstore=($(test_noobaa silence backingstore list | grep -v "NAME" | awk '{print $1}'))
-    local bucketclass=($(test_noobaa silence bucketclass list  | grep ${backingstore[1]} | awk '{print $1}'))
+    local bucketclass=($(test_noobaa bucketclass list  | grep ${backingstore[1]} | awk '{print $1}'))
+    echo "${bucketclass}"
```
liranmauda, line 368: Why are we printing the bucketclass here?

test/cli/test_cli_flow.sh
```diff
@@ -64,6 +64,7 @@ function clean {
 }
 
 function main {
+    crd_cycle
```
liranmauda, line 67: Why crd cycle is here before the install?
How should we check the crd cycle if there is no NooBaa?

kerencarmelwedel: I consulted with Ohad and we thought its the best way, after noobaa is created it tries to delete crd while its not empty

test/cli/test_cli_functions.sh
kerencarmelwedel: I will fix it for the next pr. Tried to check something here and forgot to delete it

test/cli/test_cli_functions.sh
kerencarmelwedel: But they have exactly the same values...

test/cli/test_cli_functions.sh
kerencarmelwedel: i'll add exit.

test/cli/test_cli_functions.sh
liranmauda: Next commit 🙂

test/cli/test_cli_functions.sh
liranmauda: They have also non-noobaa crds

test/cli/test_cli_flow.sh
liranmauda: we can run this cycle prior to the install but also need to do it after the install and exclude the parts that we don't want to run.

test/cli/test_cli_functions.sh
kerencarmelwedel: Asking again to make sure - you told me to run kubectl after every noobaa command i'm checking. Should i just run it as it is or should i check its' exit status? Or is just comparing between noobaa crd and kubectl get crd enough ?

liranmauda: We need to check that the kubectl returns all the expected crds.
so, we should have a list of crd (that is only noobaa’s), and then check that all appear in the kubectl get crd.
Cacking the exit code of our function does not do that.

test/cli/test_cli_functions.sh
kerencarmelwedel: They don't, they have exactly the same values, at least how i wrote it

liranmauda: On minikube yes, on openshit there are more crds.

PR #336, opened 2020-06-16
Added pvpool as default backingstore after 2 minutes if it does not exist

romayalon:
1. Added a check if 2 minutes passed from the creation of the system in reconcile default backingstore at the part when there is still no default backingstore and still no credentials.
2. If the check returned true, created a secret for the pv-pool and set values of r.DefaultBackingStore

### DIFF
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -15,6 +15,7 @@ import (
 	"github.com/noobaa/noobaa-operator/v2/pkg/util"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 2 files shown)


pkg/system/phase4_configuring.go
```diff
@@ -437,6 +445,35 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
...
+	// create backing store
+	r.DefaultBackingStore.Spec.Type = nbv1.StoreTypePVPool
+	r.DefaultBackingStore.Spec.PVPool = &nbv1.PVPoolSpec{}
+	r.DefaultBackingStore.Spec.PVPool.Secret.Name = secret.Name
+	r.DefaultBackingStore.Spec.PVPool.Secret.Namespace = secret.Namespace
+	r.DefaultBackingStore.Spec.PVPool.StorageClass = ""
+	r.DefaultBackingStore.Spec.PVPool.NumVolumes = 1
+	r.DefaultBackingStore.Spec.PVPool.VolumeResources = &corev1.ResourceRequirements{
+		Requests: corev1.ResourceList{
+			corev1.ResourceStorage: *resource.NewQuantity(int64(50)*1024*1024*1024, resource.BinarySI),
```
jackyalbo, line 471: would take it out to a constant: default_pv_size or something like that

pkg/system/phase4_configuring.go
```diff
@@ -437,6 +445,35 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
...
+	if !util.KubeCreateSkipExisting(secret) {
+		log.Fatalf(`❌ Could not create Secret %q in Namespace %q (conflict)`, secret.Name, secret.Namespace)
+	}
+
+	// create backing store
+	r.DefaultBackingStore.Spec.Type = nbv1.StoreTypePVPool
+	r.DefaultBackingStore.Spec.PVPool = &nbv1.PVPoolSpec{}
+	r.DefaultBackingStore.Spec.PVPool.Secret.Name = secret.Name
+	r.DefaultBackingStore.Spec.PVPool.Secret.Namespace = secret.Namespace
+	r.DefaultBackingStore.Spec.PVPool.StorageClass = ""
```
jackyalbo, line 467: Do we have to provide a storageclass? it doesn't work without it? I think it suppose to take the default in that case...

pkg/system/phase4_configuring.go
```diff
@@ -426,7 +429,12 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
 		}
 	} else {
 
-		return nil
+		if time.Since(r.NooBaa.CreationTimestamp.Time).Minutes() < 2 {
```
jackyalbo, line 432: would take it out to a constant, wait_time_for_backingstore (or something a bit better :) )

pkg/system/phase4_configuring.go
romayalon: checked it and it works fine without providing anything :)

PR #337, opened 2020-06-17
TEST_CLI | Add the ability to run a function when reaching the timeout

liranmauda:
- Add the ability to run a function when reaching the timeout
- Running check_obc when reaching timeout on obc create
- Add timeout to noobaa uninstall
- Add the ability to pass new timeout in sec to timeout function


liranmauda:
Converted to draft as it is not ready for review yet.

PR #339, opened 2020-06-18
Get an RGW endpoint address from a label, as a fallback to non-existent CephObjectStore

nb-ohad:
This is a workaround for missing cephObjectStore in independent mode.

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -328,22 +328,30 @@ func (r *Reconciler) ReconcileRGWCredentials() error {
 
 	// create user if not already exists
 	// list ceph objectstores and pick the first one
+	r.CephObjectstoreUser.Spec.Store = ""
 	cephObjectStoresList := &cephv1.CephObjectStoreList{}
-	if !util.KubeList(cephObjectStoresList, &client.ListOptions{Namespace: options.Namespace}) {
-		r.Logger.Warn("failed to list ceph objectstore to use as backing store")
-		// no object stores
-		return nil
+	if util.KubeList(cephObjectStoresList, &client.ListOptions{Namespace: options.Namespace}) {
+		if len(cephObjectStoresList.Items) > 0 {
+			r.Logger.Infof("found %d ceph objectstores: %v", len(cephObjectStoresList.Items), cephObjectStoresList.Items)
+			// for now take the first one. need to decide what to do if multiple objectstores in one namespace
+			storeName := cephObjectStoresList.Items[0].ObjectMeta.Name
+			r.Logger.Infof("using objectstore %q as a default backing store", storeName)
+			r.CephObjectstoreUser.Spec.Store = storeName
+
+		} else {
+			r.Logger.Infof("did not find any ceph objectstore to use as backing store, assuming independent mode")
+		}
+
+	} else {
+		r.Logger.Infof("failed to list ceph objectstore to use as backing store, assuming independent mode")
 	}
-	if len(cephObjectStoresList.Items) == 0 {
-		r.Logger.Warn("did not find any ceph objectstore to use as backing store")
-		// no object stores
-		return nil
+
+	if r.CephObjectstoreUser.Spec.Store == "" {
+		if r.NooBaa.Labels == nil || r.NooBaa.Labels["rgw-endpoint"] == "" {
+			r.Logger.Warnf("did not find an rgw-endpoint label on the noobaa CR")
+			return nil
+		}
 	}
-	r.Logger.Infof("found %d ceph objectstores: %v", len(cephObjectStoresList.Items), cephObjectStoresList.Items)
-	// for now take the first one. need to decide what to do if multiple objectstores in one namespace
-	storeName := cephObjectStoresList.Items[0].ObjectMeta.Name
-	r.Logger.Infof("using objectstore %q as a default backing store", storeName)
-	r.CephObjectstoreUser.Spec.Store = storeName
 
 	r.Own(r.CephObjectstoreUser)
 	// create ceph objectstore user
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -600,7 +600,17 @@ func (r *Reconciler) prepareCephBackingStore() error {
 		return fmt.Errorf("Ceph object user secret %q is not ready yet", secretName)
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/system/phase4_configuring.go
```diff
@@ -579,7 +579,17 @@ func (r *Reconciler) prepareCephBackingStore() error {
 		return fmt.Errorf("Ceph object user secret %q is not ready yet", secretName)
 	}
 
-	endpoint := "http://rook-ceph-rgw-" + r.CephObjectstoreUser.Spec.Store + "." + options.Namespace + ".svc.cluster.local:80"
+	endpoint := ""
+	if r.CephObjectstoreUser.Spec.Store != "" {
+		endpoint = "http://rook-ceph-rgw-" + r.CephObjectstoreUser.Spec.Store + "." + options.Namespace + ".svc.cluster.local:80"
+
+	} else if r.NooBaa.Labels != nil && r.NooBaa.Labels["rgw-endpoint"] != "" {
```
leseb, line 586:
```suggestion
	} else if r.NooBaa.Labels != nil {
```

pkg/system/phase4_configuring.go
```diff
@@ -579,7 +579,17 @@ func (r *Reconciler) prepareCephBackingStore() error {
...
+		endpoint = r.NooBaa.Labels["rgw-endpoint"]
```
leseb, line 587:
```suggestion
	if value, ok := r.NooBaa.Labels["rgw-endpoint"]; ok {
		endpoint = fmt.Sprintf("http://%s", r.NooBaa.Labels[v])
	} else {
		return fmt.Errorf("Ceph RGW endpoint address is not available")
	}
```

pkg/system/phase2_creating.go
```diff
@@ -329,21 +329,28 @@ func (r *Reconciler) ReconcileRGWCredentials() error {
...
+	} else {
+		r.Logger.Infof("failed to list ceph objectstore to use as backing store, assuming independent mode")
 	}
-	if len(cephObjectStoresList.Items) == 0 {
-		r.Logger.Warn("did not find any ceph objectstore to use as backing store")
-		// no object stores
-		return nil
+
+	if r.CephObjectstoreUser.Spec.Store == "" {
+		if r.NooBaa.Labels == nil || r.NooBaa.Labels["rgw-endpoint"] == "" {
```
leseb, line 349: What about using `if value, ok := r.NooBaa.Labels["rgw-endpoint"]; ok {`?

pkg/system/phase4_configuring.go
nb-ohad: @leseb 
I want to bail out even in the case where we have the label but it is empty

pkg/system/phase2_creating.go
nb-ohad: This will not catch the case where the label exists but the value is empty

pkg/system/phase4_configuring.go
nb-ohad: @leseb I added the protocol but I prefer to bail out also If I find the label but it is an empty value. This is why I am looking at the returned value and not the ok flag

pkg/system/phase2_creating.go
guymguym, line 349: @nb-ohad If this second condition is false, which is the typical case when running without OCS, what happens?
Seems like you still continue and create CephObjectstoreUser with an empty store? 
If the CRD itself is missing the create will fail too.
Is this really the intended result? or am I missing something?

nb-ohad: It is the other way around. On minikube we assume that the label is not set which means that we will return nil and exit before the create

guymguym: @nb-ohad What I still don't understand is when the label is provided, so we do not return here, why do you continue to create CephObjectstoreUser? Should you create it without a store at all?

guymguym: Oh so the rgw-endpoint label is only an override, but we still create the user locally without CephObjectStore being present. Is that right?

If so I this is right, I do expect the flow to separate the cases better.
So first I would check if we should create CephObjectstoreUser, which is the goal of this method, and if that CRD is missing just return gracefully. If the CRD exists, we can then go and get CephObjectStoreList to set the store, or leave it empty for rgw endpoint outside the cluster. But I don't see how we should take into account the existence of the label in here, because all it does is to overwrite the url we use for the actual endpoint.

What do you think @nb-ohad ?

nb-ohad: @guymguym 
In the case where the level is provided, we should create a user.

I think the confusing part is that although `CephObjectStore` is not supported in independent mode, `CephObjectStoreUser` is. 

The only information we need from a `CephObjectStore` is the store name which in turn allows us to construct an RGW endpoint address. 
The new label is a workaround that enables OCS to provide an RGW endpoint address even in the case where a `CephObjectStore` is not available.

guymguym: Not sure if you saw my last comment before replying,
but in any case - I think we should return immediately from the function when the `CephObjectStoreUser` CRD is missing, and if the CRD is available, we should create the user anyhow (with/out a store). Regardless of the label. 

Providing the label just allows us to override the endpoint in some way, but we might choose other ways to get the endpoint, and that's unrelated to this function that is responsible just to ask for credentials...

I am just trying to separate these concerns...

nb-ohad: @guymguym, regarding your second comment. 
In both cases, both CRDs exist in the cluster. it is just that there is no instance of a CephObjectStore in the cluster.

I think it is reasonable to assume that both CRDs are installed or not installed together. 
If we really want we can avoid entering the whole RGW function based on the existence of `CephObjectStore`.

The reason we are checking the label here is to prevent the creation of the `CephObjectStoreUser` if we know that we will not have a way to create a Ceph backing store on phase 4.

nb-ohad: And please bear in mind that this is just a workaround that we are going to remove as soon as `CephObjectStore` will be supported in independent mode.

guymguym: My point is simple - I want to rearrange the conditions for bailing outt of this function to include just the existence of the CRD. And not the existence of the store or the label...

nb-ohad: @guymguym Pelase see https://github.com/noobaa/noobaa-operator/pull/348

PR #348, opened 2020-06-24
Change the order and relocate checks in order to separate concerns

nb-ohad:
After the change creation and configuration are clearly defined.

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -301,7 +301,7 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 	return nil
 }
 
-// ReconcileBackingStoreCredentials creates a CredentialsRequest resource if neccesary and returns
+// ReconcileBackingStoreCredentials creates a CredentialsRequest resource if necessary and returns
 // the bucket name allowed for the credentials. nil is returned if cloud credentials are not supported
 func (r *Reconciler) ReconcileBackingStoreCredentials() error {
 	// Skip if joining another NooBaa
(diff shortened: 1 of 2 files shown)


pkg/system/phase2_creating.go
```diff
@@ -321,13 +321,19 @@ func (r *Reconciler) ReconcileBackingStoreCredentials() error {
 // ReconcileRGWCredentials creates a ceph objectstore user if a ceph objectstore exists in the same namespace
 func (r *Reconciler) ReconcileRGWCredentials() error {
 	r.Logger.Info("Not running in AWS. will attempt to create a ceph objectstore user")
-	util.KubeCheck(r.CephObjectstoreUser)
+
+	// load the user and skip in case of an error, in the majority of cases an error will happen
+	// because the CRD is not installed in the cluster.
+	if !util.KubeCheck(r.CephObjectstoreUser) {
```
guymguym, line 327: But KubeCheck returns false also when the item is not found - that's not what you meant right?
What I suggest is that when `KubeCheck(r.CephObjectstoreUser)` returns false to then check if `KubeList(&cephv1.CephObjectstoreUserList{}, Options{Namespace: myNamespace})` and if that returns false then you can return nil, otherwise continue and create the user.

guymguym: The only question is if we need to change also the conditions in `ReconcileDefaultBackingStore()` in phase4 because now we create the user in any case where the CRD exists, so it will be created on AWS/Azure too, right?

nimrod-becker:
@nb-ohad any relevant still ?

nimrod-becker:
Closing, @nb-ohad if needed, plase rebase/ reopen

PR #354, opened 2020-06-28
Refactor backingstore mode info mapping and add severity

nb-ohad:
- Moved `BackingStorePhaseInfo` from `backingstore_types.go` to the reconciler (the struct only serve as an implementation detail and should not be exposed as an API)
- Refactor out all redundant info from `BackingStorePhaseInfo`
- Added severity to `BackingStorePhaseInfo`, to represent an event severity attached to the mode
- Renamed `BackingStorePhaseInfo` to better fit its actual usage.

Fixing BZ 1847815


pkg/backingstore/reconciler.go
```diff
@@ -30,30 +30,35 @@ import (
...
+		"DELETING":            {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
+		"SCALING":             {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
+		"MOST_NODES_ISSUES":   {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
+		"MANY_NODES_ISSUES":   {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
+		"MOST_STORAGE_ISSUES": {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
+		"MANY_STORAGE_ISSUES": {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
+		"MANY_NODES_OFFLINE":  {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
+		"LOW_CAPACITY":        {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
+		"OPTIMAL":             {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
+		"HAS_NO_NODES":        {nbv1.BackingStorePhaseRejected, corev1.EventTypeNormal},
```
romayalon, line 56: why all of them are normal? didn't we want to change it for some to be warnings and some normal?

nb-ohad: Yed I saw it after the push, and already pushed a fix.
Thanks

pkg/backingstore/reconciler.go
```diff
@@ -30,30 +30,36 @@ import (
...
-		"MANY_NODES_OFFLINE":  {nbv1.BackingStorePhaseReady, "BackingStorePhaseReady", "Backing store mode: MANY_NODES_OFFLINE"},
-		"LOW_CAPACITY":        {nbv1.BackingStorePhaseReady, "BackingStorePhaseReady", "Backing store mode: LOW_CAPACITY"},
-		"OPTIMAL":             {nbv1.BackingStorePhaseReady, "BackingStorePhaseReady", "Backing store mode: OPTIMAL"},
-		"HAS_NO_NODES":        {nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: HAS_NO_NODES"},
-		"ALL_NODES_OFFLINE":   {nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: ALL_NODES_OFFLINE"},
-		"NO_CAPACITY":         {nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: NO_CAPACITY"},
-		"IO_ERRORS":           {nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: IO_ERRORS"},
-		"STORAGE_NOT_EXIST":   {nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: STORAGE_NOT_EXIST"},
-		"AUTH_FAILED":         {nbv1.BackingStorePhaseRejected, "BackingStorePhaseRejected", "Backing store mode: AUTH_FAILED"},
+func modeInfoMap() map[string]ModeInfo {
```
guymguym, line 45: @nb-ohad Why do you use an `init()` function to call another function just to return a map when you can create the map directly on the global var - am I missing something?
```
var bsModeInfoMap = map[string]ModeInfo{
	"INITIALIZING":        {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
	"DELETING":            {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
	"SCALING":             {nbv1.BackingStorePhaseReady, corev1.EventTypeNormal},
	...
}
```

nb-ohad: @guymguym this was in the original code. I didn't change it. But you are perfectly right.

PR #360, opened 2020-07-06
Replace RGW endpoint label encoding (base64) with a simple ":" to "_" encoding

nb-ohad:
Base64 padding char `=` cannot be used in a k8s label.
So we dropped base64 encoding in favor of a simple replacement for the host-port separator char `:`

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -339,16 +339,16 @@ func (r *Reconciler) ReconcileRGWCredentials() error {
 			r.CephObjectstoreUser.Spec.Store = storeName
 
 		} else {
-			r.Logger.Infof("did not find any ceph objectstore to use as backing store, assuming independent mode")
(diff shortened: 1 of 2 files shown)


pkg/system/phase2_creating.go
```diff
@@ -347,8 +347,8 @@ func (r *Reconciler) ReconcileRGWCredentials() error {
 	}
 
 	if r.CephObjectstoreUser.Spec.Store == "" {
-		if r.NooBaa.Labels == nil || r.NooBaa.Labels["rgw-endpoint-base64"] == "" {
-			r.Logger.Warnf("did not find an rgw-endpoint-base64 label on the noobaa CR")
+		if r.NooBaa.Labels == nil || r.NooBaa.Labels["rgw-endpoint"] == "" {
+			r.Logger.Warnf("did not find an rgw-endpoint label on the noobaa CR")
```
leseb, line 351:
```suggestion
			r.Logger.Warn("did not find an rgw-endpoint label on the noobaa CR")
```

PR #372, opened 2020-07-19
Remove app label and finalizer from CR templates and add them in code

nb-ohad:
The `BackingStore` object sent to `KubeCheck` at the start of the backing store reconcile loop is already filled with a finalizer and label (loaded from the backing store cr template). This means that even in the case where the k8s resource has no finalizers or labels, the in-memory object will contain values for those fields. 
This prevents the operator from adding a finalizer at the start of the reconcile loop (because the operator believes that a finalizer is already present on the backing store).

This PR removed any meta info from the template and adds it back in code after checking that the loaded backing store is missing some or all of the desired meta fields.

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1858385

### DIFF
--- a/deploy/crds/noobaa.io_v1alpha1_backingstore_cr.yaml
+++ b/deploy/crds/noobaa.io_v1alpha1_backingstore_cr.yaml
@@ -1,9 +1,5 @@
 apiVersion: noobaa.io/v1alpha1
 kind: BackingStore
 metadata:
-  name: aws1
-  labels:
-    app: noobaa
-  finalizers:
-    - noobaa.io/finalizer
+  name: default
 spec:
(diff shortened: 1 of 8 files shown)


nb-ohad:
@jackyalbo . This is out of the scope of this PR.  and different CRs need diff fields.
Let's approve this one and then talk about the rest

jackyalbo:
@nb-ohad I just think we need to keep the same behaviour to all our CR. Or just remove the finalizer here and keep the old labeling like in the others, or to change to all. I would just remove the finalizer here and do like in all our other CRs.

nb-ohad:
@jackyalbo If we go this way the issue will be that CRs created from anywhere but the noobaa CLI will have the label missing. 
I decided to update the code for both bucketclass and noobaa CRs. It is not a simple copy-paste solution as the noobaa CR needs a different finalizer. But in any case, it will be handled.

PR #378, opened 2020-07-27
Add predicates to filter out non-relevant events and log relevant events

nb-ohad:
Also, remove manager SyncPeriod time dev code left by accident in the code

This PR solves the infinite reconcile problem we have in the operator. 
The main reason for the problem is that the operator is watching resources (`NooBaa`, `BackingStore`, `BucketClass`) that it is also responsible for updating,  specifically writing the status of each of these resources on every reconcile cycle.

The fix is to filter the events the reconcile reacts to. Allowing only meaningful changes to passthrough  

Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1849309

### DIFF
--- a/pkg/util/predicates.go
+++ b/pkg/util/predicates.go
@@ -1,0 +1,112 @@
+package util
+
+import (
+	"reflect"
+
+	"github.com/sirupsen/logrus"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// ComposePredicates will compose a variable number of predicte and return a predicate that
+// will allow events that are allowed by any of the given predicates.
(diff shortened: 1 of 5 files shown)


pkg/util/predicates.go
```diff
@@ -0,0 +1,101 @@
...
+		!reflect.DeepEqual(e.MetaOld.GetLabels(), e.MetaNew.GetLabels())
+}
+
+// FinalizersChangedPredicate will only allow events that changed Metadata.Finalizers
+type FinalizersChangedPredicate struct {
+	predicate.Funcs
+}
+
+// Update implements the update event trap for LabelsChangedPredicate
+func (p FinalizersChangedPredicate) Update(e event.UpdateEvent) bool {
```
romayalon, line 68: change the comment to // Update implements the update event trap for FinalizersChangedPredicate

nb-ohad: done

PR #384, opened 2020-08-03
Add db-type flag and the ability to install postgress using the CLI

liranmauda:
### Added
- Added db-type flag that can get mongodb or postgres
- Adding statfulset postgres db
- Implemeting the db-type functionality, when using --db-type postgres the db to be installed is postgres

### Missing 
- In statfulset postgres db get the cradentials should be from a secret, currently is is hard coded defaults.
- In statfulset postgres db the mountPath is the same as mongo's

- ~We override NooBaaDB with the postgres yaml when we are calling db-type flag. probebly better way is to keep one NooBaaDB and instead of replacing it, assigned it with NooBaaMongoDB or NooBaaPostgresDB~ 

### Limitations
- noobaa-core is not connecting - as planned for this stage. 
- When using this flag it revokes the db-image flag and currently the postgres image is hard coded (in the options). 
- We might want to change also the `/noobaa_init_files/noobaa_init.sh` in the core side. 

### DIFF
--- a/deploy/internal/statefulset-postgres-db.yaml
+++ b/deploy/internal/statefulset-postgres-db.yaml
@@ -1,0 +1,78 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: noobaa-db
+  labels:
+    app: noobaa
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      noobaa-db: noobaa
+  serviceName: noobaa-db
+  updateStrategy:
+    type: RollingUpdate
+  template:
+    metadata:
+      labels:
+        app: noobaa
+        noobaa-db: noobaa
+    spec:
+      serviceAccountName: noobaa
+      initContainers:
+        #----------------#
+        # INIT CONTAINER #
+        #----------------#
+        - name: init
+          image: NOOBAA_CORE_IMAGE
+          command:
+            - /noobaa_init_files/noobaa_init.sh
+            - init_mongo
+          resources:
+            requests:
+              cpu: "500m"
+              memory: "500Mi"
+            limits:
+              cpu: "500m"
+              memory: "500Mi"
+          volumeMounts:
+            - name: db
+              mountPath: /mongo_data
+      containers:
+        #--------------------#
+        # Postgres CONTAINER #
+        #--------------------#
+        - name: db
+          image: NOOBAA_DB_IMAGE
+          env:
+            - name: POSTGRES_DB
+              value: nbcore
+            - name: POSTGRES_USER
+              value: postgres
+            - name: POSTGRES_PASSWORD
+              value: noobaa
+          magePullPolicy: "IfNotPresent"
+          ports:
+            - containerPort: 5432
+          resources:
+            requests:
+              cpu: "2"
+              memory: "4Gi"
+            limits:
+              cpu: "2"
+              memory: "4Gi"
+          volumeMounts:
+            - name: db
+              # mountPath: /var/lib/postgresql/data
+              mountPath: /data
+  volumeClaimTemplates:
+    - metadata:
+        name: db
+        labels:
+          app: noobaa
+      spec:
+        accessModes:
+          - ReadWriteOnce
+        resources:
+          requests:
+            storage: 50Gi
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -77,7 +77,7 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 	if err := r.ReconcileObject(r.CoreApp, r.SetDesiredCoreApp); err != nil {
 		return err
 	}
-	if err := r.ReconcileObject(r.NooBaaDB, r.SetDesiredNooBaaDB); err != nil {
+	if err := r.ReconcileDB(); err != nil {
 		return err
 	}
 	if err := r.ReconcileObject(r.ServiceMgmt, r.SetDesiredServiceMgmt); err != nil {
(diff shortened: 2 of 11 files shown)


pkg/system/phase2_creating.go
```diff
@@ -61,6 +62,15 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
 
+	if r.NooBaa.Spec.DBType == "postgres" {
+		r.NooBaaDB = util.KubeObject(bundle.File_deploy_internal_statefulset_postgres_db_yaml).(*appsv1.StatefulSet)
+		r.ServiceDb = util.KubeObject(bundle.File_deploy_internal_service_postgres_db_yaml).(*corev1.Service)
+		r.NooBaaDB.Namespace = r.Request.Namespace
+		r.NooBaaDB.Name = r.Request.Name + "-db"
+		r.ServiceDb.Namespace = r.Request.Namespace
+		r.ServiceDb.Name = r.Request.Name + "-db"
+	}
+
```
nb-ohad, line 73: Why not create both DB statefulsets and DB services on the reconciler (same name and namespace) and choose which one to update to k8s ?

pkg/system/reconciler.go
```diff
@@ -99,21 +101,23 @@ func NewReconciler(
...
+		NooBaa:          util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_noobaa_cr_yaml).(*nbv1.NooBaa),
+		ServiceAccount:  util.KubeObject(bundle.File_deploy_service_account_yaml).(*corev1.ServiceAccount),
+		CoreApp:         util.KubeObject(bundle.File_deploy_internal_statefulset_core_yaml).(*appsv1.StatefulSet),
+		NooBaaDB:        util.KubeObject(bundle.File_deploy_internal_statefulset_db_yaml).(*appsv1.StatefulSet),
+		// NooBaaPostgresDB:    util.KubeObject(bundle.File_deploy_internal_statefulset_postgres_db_yaml).(*appsv1.StatefulSet),
+		ServiceDb: util.KubeObject(bundle.File_deploy_internal_service_db_yaml).(*corev1.Service),
+		// ServicePostgresDb:   util.KubeObject(bundle.File_deploy_internal_service_postgres_db_yaml).(*corev1.Service),
 		ServiceMgmt:         util.KubeObject(bundle.File_deploy_internal_service_mgmt_yaml).(*corev1.Service),
 		ServiceS3:           util.KubeObject(bundle.File_deploy_internal_service_s3_yaml).(*corev1.Service),
-		ServiceDb:           util.KubeObject(bundle.File_deploy_internal_service_db_yaml).(*corev1.Service),
```
nb-ohad, line 116: Are you sure this should be removed, I could find the lines that reconcile the service for the mongodb case

liranmauda: It is not being removed, It has been reorder

nb-ohad: Sorry missed it. I now see it was moved. But in any case it would be better if you have both versions on the reconciler then choose which one to update k8s with

pkg/system/phase2_creating.go
nb-ohad: You are also not loading the current resource from k8s which mean that any allowed changes to the sts or service will be overwritten

pkg/system/system.go
```diff
@@ -56,7 +56,7 @@ func CmdCreate() *cobra.Command {
 	cmd.Flags().String("core-resources", "", "Core resources JSON")
 	cmd.Flags().String("db-resources", "", "DB resources JSON")
 	cmd.Flags().String("endpoint-resources", "", "Endpoint resources JSON")
-	cmd.Flags().Bool("obc-clenup-policy", false, "Create NooBaa system with obc cleanup policy")
+	cmd.Flags().Bool("obc-cleanup-policy", false, "Create NooBaa system with obc cleanup policy")
```
romayalon, line 59:
```suggestion
	cmd.Flags().Bool("use-obc-cleanup-policy", false, "Create NooBaa system with obc cleanup policy")
```

pkg/system/reconciler.go
```diff
@@ -206,12 +210,16 @@ func NewReconciler(
 
 // CheckAll checks the state of all the objects controlled by the system
 func (r *Reconciler) CheckAll() {
+
 	CheckSystem(r.NooBaa)
 	util.KubeCheck(r.CoreApp)
-	util.KubeCheck(r.NooBaaDB)
 	util.KubeCheck(r.ServiceMgmt)
 	util.KubeCheck(r.ServiceS3)
-	util.KubeCheck(r.ServiceDb)
```
romayalon, line 214: why `util.KubeCheck(r.ServiceDb)` is not needed anymore?

nb-ohad:
LGTM, just see the comment about the removed `util.KubeCheck(r.ServiceDb)` Romy found

pkg/system/system.go
liranmauda: Fixed

pkg/system/reconciler.go
liranmauda: Fixed

pkg/system/system.go
```diff
@@ -131,6 +133,11 @@ func LoadSystemDefaults() *nbv1.NooBaa {
 		dbImage := options.DBImage
 		sys.Spec.DBImage = &dbImage
 	}
+	//  naiveally changing the db postgres image to the hardcoded one if the db type is postgres
```
jeniawhite, line 136: naively not naiveally

deploy/internal/statefulset-postgres-db.yaml
```diff
@@ -0,0 +1,78 @@
...
+          resources:
+            requests:
+              cpu: "2"
+              memory: "4Gi"
+            limits:
+              cpu: "2"
+              memory: "4Gi"
+          volumeMounts:
+            - name: db
+              # mountPath: /var/lib/postgresql/data
```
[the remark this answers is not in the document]
liranmauda: You are **probably** right, In mongo db we are chmoding things on noobaa_init.sh
That's why this is in a comment so we will have the proper mountPath written in the yaml. 

once we do it on the core side, we might want to change that back to `/var/lib/postgresql/data`

pkg/system/system.go
liranmauda: Fixed

PR #386, opened 2020-08-09
Handle updateStatus errors - to continue recocnciles on errors

jackyalbo:
Fixing BZ https://bugzilla.redhat.com/show_bug.cgi?id=1866781
updateStatus errors wasn't handled (only printed) causing that an error happening while updating the status (just in the end of the the reconcile cycle) to leave the phase not changed until the next reconcile which will never run as no new events will happen (as the backingstore is really in Ready state - just wasn't able to update the CR)

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -210,7 +210,13 @@ func (r *Reconciler) Reconcile() (reconcile.Result, error) {
 		}
 	}
 
-	r.UpdateStatus()
+	err = r.UpdateStatus()
+	if err != nil {
+		res.RequeueAfter = 3 * time.Second
+		// leave current phase as is
+		r.SetPhase("", "TemporaryError", err.Error())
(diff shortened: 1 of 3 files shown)


guymguym:
@jackyalbo @nimrod-becker 

Calling SetPhase is not going to have an effect - since we failed to update the status, updating the phase in memory for a reconciler object that is about to be thrown away, is not going to get persisted anywhere. In any case your fix to requeue to retry reconcile is the important thing.

Regarding the api server errors - we might get temporary errors, not-found errors if the CR got removed while we are reconciling, schema issues, missing CRD, etc. But in any of these cases it is still a good idea to requeue and let the reconcile state machine start over and apply its amazing logic.

To sum up - I would just remove this call to r.SetPhase() to avoid confusion, and add a good comment that explains our method of error handling for UpdateStatus as we explained.

jackyalbo:
I agree that setPhase don't have a real effect - except for another print
but as we already have enough of them - removed and added the requested comments - find it here:
#388

PR #389, opened 2020-08-10
CLI noobaa status and uninstall shows actual installed image

liranmauda:
CLI noobaa status and uninstall shows actual installed image:

- When there is a system CLI noobaa status will print the actual image that we used
- Add prints of the db image 

Fixes: #64 

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -76,7 +76,7 @@ func RunInstall(cmd *cobra.Command, args []string) {
 // RunUninstall runs a CLI command
 func RunUninstall(cmd *cobra.Command, args []string) {
 	log := util.Logger()
-	version.RunVersion(cmd, args)
+	system.RunSystemVersionsStatus(cmd, args)
 	log.Printf("Namespace: %s", options.Namespace)
 	log.Printf("")
 	log.Printf("System Delete:")
@@ -99,7 +99,8 @@ func RunUninstall(cmd *cobra.Command, args []string) {
 // RunStatus runs a CLI command
 func RunStatus(cmd *cobra.Command, args []string) {
 	log := util.Logger()
-	version.RunVersion(cmd, args)
+
+	system.RunSystemVersionsStatus(cmd, args)
 	log.Printf("Namespace: %s", options.Namespace)
 	log.Printf("")
 	log.Printf("CRD Status:")
--- a/pkg/system/system.go
+++ b/pkg/system/system.go
@@ -13,6 +13,7 @@ import (
 	"github.com/noobaa/noobaa-operator/v2/pkg/nb"
 	"github.com/noobaa/noobaa-operator/v2/pkg/options"
 	"github.com/noobaa/noobaa-operator/v2/pkg/util"
+	"github.com/noobaa/noobaa-operator/v2/version"
 
 	"github.com/spf13/cobra"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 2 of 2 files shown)


pkg/system/system.go
```diff
@@ -394,6 +395,38 @@ func RunYaml(cmd *cobra.Command, args []string) {
...
+
+	sysKey := client.ObjectKey{Namespace: options.Namespace, Name: options.SystemName}
+	r := NewReconciler(sysKey, klient, scheme.Scheme, nil)
+	util.KubeCheckQuiet(r.NooBaa)
+
+	noobaaImage := ""
+	noobaaDbImage := ""
+	noobaaOperatorImage := ""
+
+	if r.NooBaa.Spec.DBImage == nil {
```
guymguym, line 411: Why is the existence of spec.DBImage is the condition to all?? Doesn't make sense?
Also if the spec is empty, it doesn't mean that the installed versions are matching the CLI versions, so we should read the image versions from the pods (sts/deployment).

So I suggest this function will unconditionally read the actual images from the deploy/sts (if they even exist), and then also calculate the desired images based on the spec, and then print those images and add a warning if the desired and actual differ.

pkg/system/system.go
```diff
@@ -394,6 +395,38 @@ func RunYaml(cmd *cobra.Command, args []string) {
 	util.Panic(p.PrintObj(sys, os.Stdout))
 }
 
+// RunSystemVersionsStatus runs a CLI command
+func RunSystemVersionsStatus(cmd *cobra.Command, args []string) {
+	log := util.Logger()
+	klient := util.KubeClient()
+
+	sysKey := client.ObjectKey{Namespace: options.Namespace, Name: options.SystemName}
+	r := NewReconciler(sysKey, klient, scheme.Scheme, nil)
```
guymguym, line 404: You don't seem to need the reconciler. Just declare a nbv1.NooBaa instance and KubeCheck it.

pkg/install/install.go
```diff
@@ -99,7 +99,8 @@ func RunUninstall(cmd *cobra.Command, args []string) {
 // RunStatus runs a CLI command
 func RunStatus(cmd *cobra.Command, args []string) {
 	log := util.Logger()
-	version.RunVersion(cmd, args)
+
+	system.RunSystemVersionsStatus(cmd, args)
```
guymguym, line 103: Few lines below we are calling system.RunStatus which already loads the system and a lot of the current status, so why aren't we just adding this functionality over there? I think in general we should show both the CLI version information clearly, and also the current system installation versions clearly, each one in its own  section. But as it is now I think it will still be confusing.

liranmauda: 2 reasons for that:
1. I wanted to preserve the structure of the output. 
2. RunUninstall is not calling system.RunStatus and I wanted to fix the images there also.

pkg/system/system.go
liranmauda: 👍

pkg/system/system.go
liranmauda: Regarding the spec.DBImage you are correct :) 

Regarding the check of the images, not sure I fully understand, I will push a change with 2 more functions that checking the images and printing  a warning. Hopefully it answers this point.

pkg/install/install.go
guymguym: ok

pkg/system/system.go
```diff
@@ -401,6 +402,70 @@ func RunYaml(cmd *cobra.Command, args []string) {
...
+// CheckNooBaaDBImages runs a CLI command
+func CheckNooBaaDBImages(cmd *cobra.Command, sys *nbv1.NooBaa, args []string) string {
+	log := util.Logger()
+
+	noobaaDbImage := *sys.Spec.DBImage
+	noobaaDBStatefulset := util.KubeObject(bundle.File_deploy_internal_statefulset_db_yaml).(*appsv1.StatefulSet)
+	noobaaDBStatefulset.Namespace = options.Namespace
+	util.KubeCheckQuiet(noobaaDBStatefulset)
+	noobaaDBStatefulsetImage := noobaaDBStatefulset.Spec.Template.Spec.Containers[0].Image
+	if noobaaDbImage != noobaaDBStatefulsetImage {
```
guymguym, line 430: I think you will need to restrict this only when `sys.Spec.DBImage != nil` because if the spec did not specify the dbImage and the operator applied it's default image we don't want to fail or warn. Same for the running image - need to check if the statefulset even exists to avoid failing -
```suggestion
	desiredImage := ""
	runningImage := ""
	if sys.Spec.DBImage != nil {
		desiredImage = *sys.Spec.DBImage
	}
	sts := util.KubeObject(bundle.File_deploy_internal_statefulset_db_yaml).(*appsv1.StatefulSet)
	sts.Namespace = options.Namespace
	if util.KubeCheckQuiet(sts) {
		runningImage = sts.Spec.Template.Spec.Containers[0].Image
	}
	if desiredImage != "" && runningImage != "" && desiredImage != runningImage {
```

pkg/system/system.go
```diff
@@ -401,6 +402,70 @@ func RunYaml(cmd *cobra.Command, args []string) {
...
+func CheckNooBaaImages(cmd *cobra.Command, sys *nbv1.NooBaa, args []string) string {
+	log := util.Logger()
+
+	noobaaImage := *sys.Spec.Image
+	noobaaActualImage := sys.Status.ActualImage
+	noobaaCoreStatefulset := util.KubeObject(bundle.File_deploy_internal_statefulset_core_yaml).(*appsv1.StatefulSet)
+	noobaaCoreStatefulset.Namespace = options.Namespace
+	util.KubeCheckQuiet(noobaaCoreStatefulset)
+	noobaaCoreStatefulsetImage := noobaaCoreStatefulset.Spec.Template.Spec.Containers[0].Image
+	if noobaaActualImage != noobaaCoreStatefulsetImage {
```
guymguym, line 415: I don't think spec.Image 
```suggestion
	desiredImage := ""
	runningImage := ""
	if sys.Status.ActualImage != "" {
		desiredImage = sys.Status.ActualImage
	} else if sys.Spec.Image != nil {
		desiredImage = *sys.Spec.DBImage
	}
	sts := util.KubeObject(bundle.File_deploy_internal_statefulset_core_yaml).(*appsv1.StatefulSet)
	sts.Namespace = options.Namespace
	if util.KubeCheckQuiet(sts) {
		runningImage = sts.Spec.Template.Spec.Containers[0].Image
	}
	if desiredImage != "" && runningImage != "" && desiredImage != runningImage {
```

guymguym (approved):

PR #390, opened 2020-08-10
Will reconcile on mode change for backingstores

jackyalbo:
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1867762

### DIFF
--- a/pkg/controller/backingstore/backingstore_controller.go
+++ b/pkg/controller/backingstore/backingstore_controller.go
@@ -7,6 +7,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 
 	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
@@ -45,6 +46,7 @@ func Add(mgr manager.Manager) error {
 		predicate.GenerationChangedPredicate{},
 		util.LabelsChangedPredicate{},
 		util.FinalizersChangedPredicate{},
+		backingStoreModeChangedPredicate{},
 	)
 
 	// Watch for changes on resources to trigger reconcile
@@ -65,4 +67,23 @@ func Add(mgr manager.Manager) error {
 	}
 
 	return nil
+}
+
+// backingStoreModeChangedPredicate will only allow events that changed Status.Mode.ModeCode.
+// This predicate should be used only for BackingsStore objects!
+type backingStoreModeChangedPredicate struct {
+	predicate.Funcs
+}
+
+// Update implements the update event trap for LabelsChangedPredicate
+func (p backingStoreModeChangedPredicate) Update(e event.UpdateEvent) bool {
+	if e.ObjectOld == nil || e.ObjectNew == nil {
+		return false
+	}
+	oldBackingStore, oldCastOk := e.ObjectOld.(*nbv1.BackingStore)
+	newBackingStore, newCastOk := e.ObjectNew.(*nbv1.BackingStore)
+	if !oldCastOk || !newCastOk {
+		return false
+	}
+	return oldBackingStore.Status.Mode.ModeCode != newBackingStore.Status.Mode.ModeCode
 }


pkg/controller/backingstore/backingstore_controller.go
```diff
@@ -66,3 +68,19 @@ func Add(mgr manager.Manager) error {
 
 	return nil
 }
+
+// ModeChangedPredicate will only allow events that changed Status.Mode.ModeCode.
+// This predicate should be used only for BackingsStore objects!
+type ModeChangedPredicate struct {
```
nb-ohad, line 74: Can you please prefix the name with BackingStore because this contains a direct cast to the backing store object and it is not a general-purpose predicate. 

Also, this predicate is checking the status section, can we add this to the name as well?

I am suggesting: `BackingStoreStatusModeCodeChanged`

pkg/controller/backingstore/backingstore_controller.go
```diff
@@ -66,3 +68,19 @@ func Add(mgr manager.Manager) error {
...
+	predicate.Funcs
+}
+
+// Update implements the update event trap for LabelsChangedPredicate
+func (p ModeChangedPredicate) Update(e event.UpdateEvent) bool {
+	if e.ObjectOld == nil || e.ObjectNew == nil {
+		return false
+	}
+	oldBackingStore := e.ObjectOld.(*nbv1.BackingStore)
+	newBackingStore := e.ObjectNew.(*nbv1.BackingStore)
```
nb-ohad, line 84: You should check the ok return value for the cast in order to validate a successful cast and return false if not

pkg/controller/backingstore/backingstore_controller.go
guymguym: That's a hell of a name!
 
long-name

pkg/controller/backingstore/backingstore_controller.go
```diff
@@ -66,3 +68,19 @@ func Add(mgr manager.Manager) error {
...
+// ModeChangedPredicate will only allow events that changed Status.Mode.ModeCode.
+// This predicate should be used only for BackingsStore objects!
+type ModeChangedPredicate struct {
+	predicate.Funcs
+}
+
+// Update implements the update event trap for LabelsChangedPredicate
+func (p ModeChangedPredicate) Update(e event.UpdateEvent) bool {
+	if e.ObjectOld == nil || e.ObjectNew == nil {
+		return false
```
guymguym, line 81: What does false mean? I didn't read the entire context but wanted to make sure this was intentional - after returning false here are we still queueing this backing store to reconcile?

pkg/controller/backingstore/backingstore_controller.go
guymguym: I don't think that's necessarily a good idea - let me explain:

The current form of type-assertion (with just one ret value) is simpler and more assertive, because if it fails it will cause a runtime panic. Since most of our code is using concrete types (compared to say generic libraries) then a type assert mismatch is not a real functional error that requires any status to the user etc, but it is just a bug and the best thing we could do is make sure we put a spotlight on the issue for devs to easily see the issue.

pkg/controller/backingstore/backingstore_controller.go
nb-ohad: false means we drop the event - not queueing. 

For Update events, both new and old objects must be non-nil. (and also new and old meta)

Please note that the first time a resource is introduced into the client cache a create event (not update) will be raised, even if the resource has been created a long time ago. This means that whenever the controller is initialized (when the operator is initializing) all watched resources will get a create even dispatched for them and because we do not filter any create events, in this case, a reconcile will be queued.

The checks here are a safeguard that Jacly has copied from the predicates that I wrote and I, in turn, copied from the built-in predicates.

pkg/controller/backingstore/backingstore_controller.go
nb-ohad: Agreed, resolving.

pkg/controller/backingstore/backingstore_controller.go
nb-ohad: I still think the name should explain that this predicate is diverging from the established path (not reconciling status changes - because status should always be updated from the owning controller)

pkg/controller/backingstore/backingstore_controller.go
jackyalbo: false mean don't process:
https://godoc.org/sigs.k8s.io/controller-runtime/pkg/predicate#Predicate

PR #366, opened 2020-07-13
GCP:Creation of default backing store

anubhav-here:
Solves #367 

Signed-off-by: Anubhav Deep <adeep@redhat.com>

PR #392, opened 2020-08-13
GCP:Creation of default backing store

rajatsing:
This PR is the same as PR #366 but since I didn't have permission to push commits so raising a new PR.

Signed-off-by: RAJAT SINGH <rajasing@redhat.com>

### DIFF
--- a/deploy/internal/cloud_creds_gcp_cr.yaml
+++ b/deploy/internal/cloud_creds_gcp_cr.yaml
@@ -1,0 +1,17 @@
+apiVersion: cloudcredential.openshift.io/v1
+kind: CredentialsRequest
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: CRED-REQ-NAME
+  namespace: CRED-REQ-NAMESPACE
+spec:
+  secretRef:
+    name: CRED-SECRET-NAME
+    namespace: CRED-SECRET-NAMESPACE
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: GCPProviderSpec
+    predefinedRoles:
+    - roles/storage.admin
+    skipServiceCheck: true
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -1810,6 +1810,27 @@ spec:
     namespace: CRED-SECRET-NAMESPACE
 `
 
+const Sha256_deploy_internal_cloud_creds_gcp_cr_yaml = "f4415e851da03426e8c31a7cb5b904b4438d958a5297c70b967ca6c2881d360f"
+
+const File_deploy_internal_cloud_creds_gcp_cr_yaml = `apiVersion: cloudcredential.openshift.io/v1
+kind: CredentialsRequest
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: CRED-REQ-NAME
+  namespace: CRED-REQ-NAMESPACE
+spec:
+  secretRef:
+    name: CRED-SECRET-NAME
+    namespace: CRED-SECRET-NAMESPACE
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: GCPProviderSpec
+    predefinedRoles:
+    - roles/storage.admin
+    skipServiceCheck: true
+`
+
 const Sha256_deploy_internal_configmap_empty_yaml = "6405c531c6522ecd54808f5cb531c1001b9ad01a73917427c523a92be44f348f"
 
 const File_deploy_internal_configmap_empty_yaml = `apiVersion: v1
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -1,12 +1,18 @@
 package system
 
 import (
+	"context"
 	"fmt"
 	"net"
 	"net/url"
 	"strconv"
 	"strings"
 	"time"
+
+	"encoding/json"
+
+	"cloud.google.com/go/storage"
+	"google.golang.org/api/option"
 
 	"github.com/marstr/randname"
 	nbv1 "github.com/noobaa/noobaa-operator/v2/pkg/apis/noobaa/v1alpha1"
@@ -25,6 +31,10 @@ import (
 	"github.com/aws/aws-sdk-go/aws/session"
 	"github.com/aws/aws-sdk-go/service/s3"
 )
+
+type gcpAuthJSON struct {
+	ProjectID string `json:"project_id"`
+}
 
 // ReconcilePhaseConfiguring runs the reconcile phase
 func (r *Reconciler) ReconcilePhaseConfiguring() error {
@@ -442,6 +452,11 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
 		if err := r.prepareAzureBackingStore(); err != nil {
 			return err
 		}
+	} else if r.GCPCloudCreds.UID != "" {
+		log.Infof("CredentialsRequest %q created.  creating default backing store on GCP objectstore", r.GCPCloudCreds.Name)
+		if err := r.prepareGCPBackingStore(); err != nil {
+			return err
+		}
 	} else {
 		minutesSinceCreation := time.Since(r.NooBaa.CreationTimestamp.Time).Minutes()
 		if minutesSinceCreation < 2 {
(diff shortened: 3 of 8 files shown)


pkg/system/phase4_configuring.go
```diff
@@ -597,6 +604,64 @@ func (r *Reconciler) prepareAzureBackingStore() error {
...
+	return nil
+}
+
+func (r *Reconciler) createGCPBucketForBackingStore(client *storage.Client, projectID, bucketName string) error {
+	// [START create_bucket]
+	ctx := context.Background()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
+	defer cancel()
+	if err := client.Bucket(bucketName).Create(ctx, projectID, &storage.BucketAttrs{
```
jeniawhite, line 653: I wouldn't send BucketAttrs, just send nil and use the defaults.
Not sure if the majority customers would want to deploy low-cost storage in Asia.

deploy/internal/cloud_creds_gcp_cr.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: cloudcredential.openshift.io/v1
```
jeniawhite, line 1: Is there a particular reason to use these values of names and namespaces instead of following the convention of other CR creations (Azure and AWS):
```
openshift-ingress-gcp,
openshift-cloud-credential-operator,
cloud-credentials, 
openshift-ingress-operator
```

deploy/internal/cloud_creds_gcp_cr.yaml
```diff
@@ -0,0 +1,17 @@
...
+spec:
+  secretRef:
+    name: cloud-credentials
+    namespace: openshift-ingress-operator
+  providerSpec:
+    apiVersion: cloudcredential.openshift.io/v1
+    kind: GCPProviderSpec
+    predefinedRoles:
+    - roles/storage.admin
+    skipServiceCheck: true
```
jeniawhite, line 17: Add a newline at the end of the YAML

pkg/bundle/deploy.go
```diff
@@ -1810,6 +1810,26 @@ spec:
     namespace: CRED-SECRET-NAMESPACE
```
jeniawhite, line 1810: If you will change the CR (names and namespaces), make sure that you will build the updated deploy.go

pkg/system/phase4_configuring.go
```diff
@@ -597,6 +604,64 @@ func (r *Reconciler) prepareAzureBackingStore() error {
...
+	}
+	r.Logger.Infof("Secret %s was created successfully by cloud-credentials operator", r.GCPCloudCreds.Spec.SecretRef.Name)
+	// create bucket in GCP
+	projectID := cloudCredsSecret.StringData["project_id"]
+	ctx := context.Background()
+	gcpclient, err := storage.NewClient(ctx)
+	if err != nil {
+		r.Logger.Info(err)
+	}
+	var bucketName = strings.ToLower(randname.GenerateWithPrefix("noobaacontainer", 5))
```
jeniawhite, line 631: Make sure that this is compatible with the naming convention of GCP Buckets.
Also please change the container to a bucket (container is a term for Azure and not for GCP).

pkg/system/phase4_configuring.go
```diff
@@ -597,6 +604,64 @@ func (r *Reconciler) prepareAzureBackingStore() error {
...
+	if err := r.createGCPBucketForBackingStore(gcpclient, projectID, bucketName); err != nil {
+		r.Logger.Info(err)
+	}
+	// create backing store
+	r.DefaultBackingStore.Spec.Type = nbv1.StoreTypeGoogleCloudStorage
+	r.DefaultBackingStore.Spec.GoogleCloudStorage = &nbv1.GoogleCloudStorageSpec{
+		TargetBucket: bucketName,
+		Secret: corev1.SecretReference{
```
jeniawhite, line 639: Make sure that there are no problems regarding base64 encodings of the JSON credentials and etc...

deploy/internal/cloud_creds_gcp_cr.yaml
rajatsing: Done!

deploy/internal/cloud_creds_gcp_cr.yaml
rajatsing: done, I have changed the naming conventions according to the Azure and AWS

pkg/system/phase4_configuring.go
rajatsing: Done, changed it to `noobaaBucket`

pkg/system/phase4_configuring.go
rajatsing: Done, I have decoded the string, took its bytes part and then unmarshalled it.

pkg/system/phase4_configuring.go
```diff
@@ -597,6 +608,82 @@ func (r *Reconciler) prepareAzureBackingStore() error {
...
+
+	util.KubeCheck(r.GCPContainerCreds)
+	if r.GCPContainerCreds.UID == "" {
+		r.GCPContainerCreds.StringData = cloudCredsSecret.StringData
+		r.Own(r.GCPContainerCreds)
+		if err := r.Client.Create(r.Ctx, r.GCPContainerCreds); err != nil {
+			return fmt.Errorf("got error on GCPContainerCreds creation. error: %v", err)
+		}
+	}
+	bytes, err := b64.StdEncoding.DecodeString(string(cloudCredsSecret.Data["service_account.json"])) // Converting data
```
sabose, line 636: unmarshal the cloudCredsSecret.Data["service_account.json"] rather than decode base64.. see https://github.com/openshift/cloud-credential-operator/blob/f7e71249b9d8c6d9135c035142b7057443a6676c/pkg/gcp/actuator/utils.go#L61

rajatsing:
@jeniawhite 
Backing store creation works fine.
```
oc get backingstores.noobaa.io                                                                                                                                                                      
NAME                           TYPE                   PHASE   AGE
noobaa-default-backing-store   google-cloud-storage   Ready   118m
```
image

image

pkg/system/phase4_configuring.go
rajatsing: Updated!.

pkg/system/reconciler.go
```diff
@@ -77,6 +77,8 @@ type Reconciler struct {
 	AWSCloudCreds       *cloudcredsv1.CredentialsRequest
 	AzureCloudCreds     *cloudcredsv1.CredentialsRequest
 	AzureContainerCreds *corev1.Secret
+	GCPContainerCreds   *corev1.Secret
```
jeniawhite, line 80: Please change the `GCPContainerCreds` to `GCPBucketCreds`.
Azure uses Containers but Google uses Buckets.
Make sure to change that in all of the occurrences in code.
Thank you.

pkg/system/phase4_configuring.go
```diff
@@ -597,6 +612,78 @@ func (r *Reconciler) prepareAzureBackingStore() error {
...
+		},
+	}
+	return nil
+}
+
+func (r *Reconciler) createGCPBucketForBackingStore(client *storage.Client, projectID, bucketName string) error {
+	// [START create_bucket]
+	ctx := context.Background()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second*10)
```
jeniawhite, line 678: Not sure if a timeout of 10 seconds is safe on slow connections.
Maybe raise it a bit? Like 30 seconds or so?
What do you think?
@rajatsing @nimrod-becker

pkg/system/reconciler.go
rajatsing: I see, will do. I followed the naming convention of other platforms. Thanks for the suggestion, I will make the changes accordingly.

pkg/system/phase4_configuring.go
rajatsing: Yes, this can be increased but, in our tests, we never got timed out, we were not on very fast internet connection though.

nimrod-becker: Lets do the same as other backing stores

rajatsing: I changed it to 30 seconds.

pkg/system/reconciler.go
rajatsing: Done!

pkg/system/phase4_configuring.go
rajatsing: I changed it to 30 seconds

rajatsing:
@jeniawhite any other changes that are needed?. I have addressed all the other changes.
Thanks.

pkg/system/reconciler.go
```diff
@@ -179,11 +186,14 @@ func NewReconciler(
 	r.SecretOp.Name = r.Request.Name + "-operator"
 	r.SecretAdmin.Name = r.Request.Name + "-admin"
 	r.SecretEndpoints.Name = r.Request.Name + "-endpoints"
-	r.AzureContainerCreds.Name = r.Request.Name + "-azure-container-creds"
 	r.AWSCloudCreds.Name = r.Request.Name + "-aws-cloud-creds"
 	r.AWSCloudCreds.Spec.SecretRef.Name = r.Request.Name + "-aws-cloud-creds-secret"
+	r.AzureContainerCreds.Name = r.Request.Name + "-azure-container-creds"
 	r.AzureCloudCreds.Name = r.Request.Name + "-azure-cloud-creds"
 	r.AzureCloudCreds.Spec.SecretRef.Name = r.Request.Name + "-azure-cloud-creds-secret"
+	r.GCPBucketCreds.Name = r.Request.Name + "-gcp-container-creds"
```
jeniawhite, line 194: Please change the "-gcp-container-creds" to "-gcp-bucket-creds"

rajatsing: Done, please check.

pkg/system/phase4_configuring.go
```diff
@@ -597,6 +612,78 @@ func (r *Reconciler) prepareAzureBackingStore() error {
...
+		r.Logger.Info(err)
+	}
+
+	var bucketName = strings.ToLower(randname.GenerateWithPrefix("noobaaBucket", 5))
+	if err := r.createGCPBucketForBackingStore(gcpclient, projectID, bucketName); err != nil {
+		r.Logger.Info(err)
+	}
+
+	if errUpdate := r.Client.Update(r.Ctx, r.GCPBucketCreds); errUpdate != nil {
+		return fmt.Errorf("got error on GCPContainerCreds update. error: %v", errUpdate)
```
jeniawhite, line 660: GCPContainerCreds to GCPBucketCreds

pkg/system/phase4_configuring.go
```diff
@@ -597,6 +612,78 @@ func (r *Reconciler) prepareAzureBackingStore() error {
...
+		r.GCPBucketCreds.StringData = cloudCredsSecret.StringData
+		r.Own(r.GCPBucketCreds)
+		if err := r.Client.Create(r.Ctx, r.GCPBucketCreds); err != nil {
+			return fmt.Errorf("got error on GCPBucketCreds creation. error: %v", err)
+		}
+	}
+	authJSON := &gcpAuthJSON{}
+	err := json.Unmarshal([]byte(cloudCredsSecret.StringData["service_account.json"]), authJSON)
+	if err != nil {
+		fmt.Println("Failed to parse secret", err)
```
jeniawhite, line 644: If we fail to pass we shouldn't continue and return an error, since everything below that will fail as well.

pkg/system/phase4_configuring.go
```diff
@@ -597,6 +612,78 @@ func (r *Reconciler) prepareAzureBackingStore() error {
...
+	}
+	projectID := authJSON.ProjectID
+	r.GCPBucketCreds.StringData["GoogleServiceAccountPrivateKeyJson"] = cloudCredsSecret.StringData["service_account.json"]
+	ctx := context.Background()
+	gcpclient, err := storage.NewClient(ctx, option.WithCredentialsJSON([]byte(cloudCredsSecret.StringData["service_account.json"])))
+	if err != nil {
+		r.Logger.Info(err)
```
jeniawhite, line 651: Same here if we did not succeed in creating a client or parsing.
Should return an error and not continue.

pkg/system/phase4_configuring.go
```diff
@@ -597,6 +612,78 @@ func (r *Reconciler) prepareAzureBackingStore() error {
...
+	}
+
+	var bucketName = strings.ToLower(randname.GenerateWithPrefix("noobaaBucket", 5))
```
jeniawhite, line 654: Your bucket names must meet the following requirements:

Bucket names must contain only lowercase letters, numbers, dashes (-), underscores (_), and dots (.). Spaces are not allowed. Names containing dots require verification.
Bucket names must start and end with a number or letter.
Bucket names must contain 3-63 characters. Names containing dots can contain up to 222 characters, but each dot-separated component can be no longer than 63 characters.
Bucket names cannot be represented as an IP address in dotted-decimal notation (for example, 192.168.5.4).
Bucket names cannot begin with the "goog" prefix.
Bucket names cannot contain "google" or close misspellings, such as "g00gle".

This generates `noobaaBucket*****`
The `B` is not valid by the convention of Google.

pkg/system/phase4_configuring.go
```diff
@@ -597,6 +612,78 @@ func (r *Reconciler) prepareAzureBackingStore() error {
...
+	if err := r.createGCPBucketForBackingStore(gcpclient, projectID, bucketName); err != nil {
+		r.Logger.Info(err)
```
jeniawhite, line 656: Same here, if we fail on creation we shouldn't continue.

pkg/system/phase4_configuring.go
rajatsing: done

pkg/system/phase4_configuring.go
rajatsing: return is already in place at Line 645.

pkg/system/phase4_configuring.go
rajatsing: done

pkg/system/phase4_configuring.go
rajatsing: done!

pkg/system/phase4_configuring.go
rajatsing: Done!

Issue #364, opened 2020-07-12
Deleted or missing pods in pv-pool should be (re)created and keep existing PVCs


jackyalbo:
Currently when a used accidently delete a pvpool pod - we will stop the deletion with the finalizer,
But no more handling will be being made.
We need to handle such scenario better 
- creating missing pods will first try to connect to an existing pvcs
- maybe revert the deletion in some way

guymguym:
I don't think we need to revert the pod deletion in any way. 
Just recreate it if needed by the pv-pool spec.
The pod probably doesn't even need a finalizer, only the PVCs do.

PR #393, opened 2020-08-18
Change pod delete handling for pvpool

jackyalbo:
Few updates on how we handle the pods of the pv-pools:
- removed finalizers in order to not get the drain process getting stuck (during OCP upgrade) BZ1867762
- on spec changes in existing pods - check if the pods need to be deleted and then delete it and create a new one with the right spec - fix issue we had with changing the imagePullSecret BZ1868646
- fixing a known gap of ours #364 

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -7,6 +7,7 @@ import (
 	math "math"
 	"net/url"
 	"os"
+	"reflect"
 	"regexp"
 	"strings"
 	"time"
@@ -78,6 +79,7 @@ type Reconciler struct {
 	Secret           *corev1.Secret
 	PodAgentTemplate *corev1.Pod
 	PvcAgentTemplate *corev1.PersistentVolumeClaim
+	ServiceAccount   *corev1.ServiceAccount
 
 	SystemInfo             *nb.SystemInfo
 	ExternalConnectionInfo *nb.ExternalConnectionInfo
@@ -113,6 +115,7 @@ func NewReconciler(
 		BackingStore:     util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_backingstore_cr_yaml).(*nbv1.BackingStore),
 		NooBaa:           util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_noobaa_cr_yaml).(*nbv1.NooBaa),
 		Secret:           util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret),
+		ServiceAccount:   util.KubeObject(bundle.File_deploy_service_account_yaml).(*corev1.ServiceAccount),
 		PodAgentTemplate: util.KubeObject(bundle.File_deploy_internal_pod_agent_yaml).(*corev1.Pod),
 		PvcAgentTemplate: util.KubeObject(bundle.File_deploy_internal_pvc_agent_yaml).(*corev1.PersistentVolumeClaim),
 	}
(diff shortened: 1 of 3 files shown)


pkg/backingstore/reconciler.go
```diff
@@ -920,32 +927,56 @@ func (r *Reconciler) ReconcilePool() error {
 
 func (r *Reconciler) reconcilePvPool() error {
 	podsList := &corev1.PodList{}
+	pvcsList := &corev1.PersistentVolumeClaimList{}
 	util.KubeList(podsList, client.InNamespace(options.Namespace), client.MatchingLabels{"pool": r.BackingStore.Name})
+	util.KubeList(pvcsList, client.InNamespace(options.Namespace), client.MatchingLabels{"pool": r.BackingStore.Name})
+	if len(podsList.Items) < len(pvcsList.Items) {
+		r.reconcileDeletedPods(podsList, pvcsList)
```
dannyzaken, line 934: there is pod creation in both `reconcileDeletedPods` and `reconcileMissingPods`, and it's also not so clear from the name what's the purpose of each function.
I suggest making the flow clearer by ordering it differently: 
1. create PVCs
2. create pods
3. reconcile existing pods

something like that:
```go
	podsList := &corev1.PodList{}
	pvcsList := &corev1.PersistentVolumeClaimList{}
	util.KubeList(podsList, client.InNamespace(options.Namespace), client.MatchingLabels{"pool": r.BackingStore.Name})
	util.KubeList(pvcsList, client.InNamespace(options.Namespace), client.MatchingLabels{"pool": r.BackingStore.Name})

	if len(pvcsList.Items) < r.BackingStore.Spec.PVPool.NumVolumes {
                // only create new PVCs and not pods
		r.reconcilePVCS(podsList)
	}

        util.KubeList(pvcsList, client.InNamespace(options.Namespace), client.MatchingLabels{"pool": r.BackingStore.Name})
	if len(podsList.Items) < len(pvcsList.Items) {
                // create pods for orphan PVCs
		r.reconcileMissingPods(podsList, pvcsList)
	}

	return r.reconcileExistingPods(podsList)
}

```

pkg/backingstore/reconciler.go
```diff
@@ -920,32 +927,56 @@ func (r *Reconciler) ReconcilePool() error {
...
-			val, ok := os.LookupEnv(name)
-			if (envVar == nil && ok) || (envVar != nil && (!ok || envVar.Value != val)) {
-				r.Logger.Warnf("Pod's env variable %s does not match operator's env variable with the same name", name)
-			}
+		// check if pod need to be updated and deleted
+		if r.needUpdate(&pod) {
+			util.KubeDelete(&pod)
 		}
 
 		if !r.isPodinNoobaa(&pod) {
```
dannyzaken, line 982: should be in an `else` statement for the previous `if`

pkg/backingstore/reconciler.go
```diff
@@ -989,24 +1002,47 @@ func (r *Reconciler) reconcileExistingPods(podsList *corev1.PodList) error {
...
+	noobaaSecret := r.NooBaa.Spec.ImagePullSecret
+	if noobaaSecret == nil {
+		sa := util.KubeObject(bundle.File_deploy_service_account_yaml).(*corev1.ServiceAccount)
+		sa.Name = pod.Spec.ServiceAccountName
+		sa.Namespace = options.Namespace
+		if util.KubeCheck(sa) && !reflect.DeepEqual(sa.ImagePullSecrets, pod.Spec.ImagePullSecrets) {
+			r.Logger.Warnf("Change in Image Pull Secrets detected: SA(%v) Spec(%v)", sa.ImagePullSecrets, pod.Spec.ImagePullSecrets)
+			return true
+		}
+	} else if noobaaSecret.Name != podSecrets[0].Name {
```
dannyzaken, line 1029: maybe check if podSecrets contain noobaaSecret, would be safer

dannyzaken (approved):
just one comment

pkg/backingstore/reconciler.go
jackyalbo: as we currently replaces the whole podSecrets with an array of one secret - I think I will leave it with that - without adding extra risky complications: 
```
r.PodAgentTemplate.Spec.ImagePullSecrets =
```
if we will change to adding in the future then we will handle

Issue #304, opened 2020-05-14
Feature request/low hanging fruit: Allow changing the MongoDB URL via the NooBaa CR


lallinger-arbeit:
Since noobaa-db and noobaa-core are now separated it would make sense to have the possibility to redirect the MongoDB transactions to a external MongoDB instance.

Currently the operator sets the environment variable MONGODB_URL according to the name of the corresponding NooBaa CR. If the NooBaa CRD would be extended by a field where you can set this environment variable to a custom URL it would easily be possible to use an external MongoDB instance.
Also when this field is set to a custom MongoDB URL it would be possible to not deploy the noobaa-db statefulset and force noobaa-core to initialize the external MongoDB , if it is empty.

At the moment it is not possible to change MONGODB_URL manually via kubectl/oc set env because the operator immediately resets it.

guymguym:
Hi @lallinger-arbeit 
That's a great suggestion and I also agree it makes sense.
Can you describe some more of why/when you prefer to use an external DB? 
It would be very useful for us to understand more on this use-case and which solutions would benefit from it.
Thanks!

lallinger-arbeit:
Hi @guymguym 
In a develop environment, where it does not matter if any data is lost, the current deployment of the noobaa-db via statefulset is sufficient.

But in a production environment this can lead to problems. For example when all workload from cluster is moved to another cluster it is not easily possible to backup and restore PVCs without any downtime. Because of this we would like that noobaa uses a MongoDB which is outside of the cluster and under our full control, so we can also easily create backups of the MongoDB.

Currently we are already using an external MongoDB for noobaa. For this we use a script to override the MONGODB_URL env inside the noobaa-core and noobaa-endpoint pods. This is working fine, but it is a quite ugly hack.

Also it would be good if the connection between noobaa-core/noobaa-endpoint and the MongoDB would use a username and password for added security.

I think every solution which wants to move noobaa from one cluster to an other without downtime and data loss would benefit from this. Also the MongoDB can then be shared with other applications which would result in better usage of resources.

lallinger-arbeit:
We now pointed NoooBaa to a remote MongoDB replicaset. For this we used the already existing environment variable MONGO_RS_URL of noobaa-core/endpoint.
Setting this environment variable and the configuration for the used CA and cert (/data/mongo/ssl/) manually works fine. There is only one major downside: 
[...]

PR #395, opened 2020-08-20
Adding mongodb-url flag that can change the url of mongo, enabling to connect to a remote mongodb

liranmauda:
- Adding mongodb-url flag that can change the URL of mongo, enabling to connect to a remote mongodb
- Skipping create of db pod when given an external db URL

Expecting URL to start with:  `mongodb://` or `mongodb+srv://`

Fixes: #304

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -298,7 +298,11 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 			c.Env[j].Value = r.SetDesiredAgentProfile(c.Env[j].Value)
 
 		case "MONGODB_URL":
-			c.Env[j].Value = "mongodb://" + r.NooBaaMongoDB.Name + "-0." + r.NooBaaMongoDB.Spec.ServiceName + "/nbcore"
+			if r.NooBaa.Spec.MongoDbURL != "" {
+				c.Env[j].Value = r.NooBaa.Spec.MongoDbURL
+			} else {
+				c.Env[j].Value = "mongodb://" + r.NooBaaMongoDB.Name + "-0." + r.NooBaaMongoDB.Spec.ServiceName + "/nbcore"
+			}
 
 		case "POSTGRES_HOST":
 			c.Env[j].Value = r.NooBaaPostgresDB.Name + "-0." + r.NooBaaPostgresDB.Spec.ServiceName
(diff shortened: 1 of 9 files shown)


liranmauda:
Moving to draft.
After talking with @lallinger-arbeit and @QuintinBecker we might want to drop this effort

lallinger-arbeit:
@liranmauda could you do me a favour and rebase this? we are still using this and currently i am always rebasing this myself, but it gets harder for me to do it correctly with more and more changes...

liranmauda:
> @liranmauda could you do me a favour and rebase this? we are still using this and currently i am always rebasing this myself, but it gets harder for me to do it correctly with more and more changes...

Are you using it? 
I will rebase.

lallinger-arbeit:
> Are you using it?
yes, with the HA setup described [here](https://github.com/noobaa/noobaa-operator/issues/481)
> I will rebase.
thank you very much!

liranmauda:
Done, Hopefully I didn't miss anything :) 
@lallinger-arbeit Can you test it?

lallinger-arbeit:
Perfect, thanks! I will test it immediately!

lallinger-arbeit:
@liranmauda seems to be not right yet, noobaa-core does not get the set mongoDbURL, but the endpoints do:
grafik
grafik

lallinger-arbeit:
The attempts field now shows up as desired. Thinking about it i think it would be more straightforward if the field would be called `Retries` instead of `Attempts`. As attempts=0 would mean the trigger shouldn't be run... If you rename it to retries i would suggest the following tooltip: "Number of retries Noobaa will make if the trigger fails"

liranmauda:
pushed a fix.
Changing to "ready for review"

pkg/system/phase2_creating.go
```diff
@@ -87,8 +94,17 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
 		return err
 	}
-	if err := r.ReconcileDB(); err != nil {
-		return err
+	// create the mongo db only if mongo db url is not given.
+	if r.NooBaa.Spec.MongoDbURL == "" {
+		if err := r.UpgradeSplitDB(); err != nil {
+			return err
+		}
+		if err := r.ReconcileDB(); err != nil {
```
dannyzaken, line 102: ReconcileDB also reconciles postgres. you should only refer to `MongoDbURL` when dealing with mongo. I think for now you can verify in phase1 that if `MongoDbURL` is passed then the `DB_TYPE` must be mongo. later on, we might need to handle the migration from external mongo to postgres, where the db-type is postgres and there is a mongoURL

dannyzaken (changes requested):

lallinger-arbeit:
> 
> 
> pushed a fix.

working as inteded now, thanks @liranmauda :)

dannyzaken (approved):

PR #397, opened 2020-08-24
Set default max number of noobaa endpoints to 2 (autoscale)

nb-ohad:


nimrod-becker (approved):

guymguym:
@nb-ohad @nimrod-becker can you add some description on the rational of making this change in defaults? For the sake of who was not in the loop. Thanks

nb-ohad:
@guymguym The rational is QE test coverage on top of OCS standard deployment. 
I am not sure this is interesting here as it is an upstream related decision

nimrod-becker:
@guymguym ratio of endpoint to the default resources of core and db. We observed that 4 endpoints is too much (causes crashes) and 3 might be cutting it too close.
It's always changeable but defaults should be synced to the resources we set by default to memory of the core/db

PR #398, opened 2020-08-24
operator changes for postgres-db

dannyzaken:
* changed upstream postgres image to centos/postgresql-12-centos7
* DB service changes
* Pass postgres DB info as envs to noobaa-core


liranmauda (approved):

liranmauda:
If we don't need init container for postgres, Do we need it for mongo?

PR #406, opened 2020-09-01
Fixes for issues with retrieving agent_config when the server has issues while creating a pvpool

jackyalbo:
fix for overlooking a missing AGENT_CONFIG in pvpool secret
recheck for agent_config in case it wasn't available

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -928,6 +928,16 @@ func (r *Reconciler) ReconcilePool() error {
 }
 
 func (r *Reconciler) reconcilePvPool() error {
+	if r.Secret.StringData["AGENT_CONFIG"] == "" {
+		res, err := r.NBClient.GetHostsPoolAgentConfigAPI(nb.GetHostsPoolAgentConfigParams{
+			Name: r.BackingStore.Name,
+		})
+		if err != nil {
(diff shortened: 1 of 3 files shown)


pkg/backingstore/reconciler.go
```diff
@@ -928,6 +928,16 @@ func (r *Reconciler) ReconcilePool() error {
 }
 
 func (r *Reconciler) reconcilePvPool() error {
+	if r.Secret.StringData["AGENT_CONFIG"] == "" {
+		res, err := r.NBClient.GetHostsPoolAgentConfigAPI(nb.GetHostsPoolAgentConfigParams{
+			Name: r.BackingStore.Name,
+		})
+		if err != nil {
```
dannyzaken, line 935: not sure it's necessary here, but make sure you don't need to handle different errors in a different way (persistent vs. temp error)

dannyzaken (approved):

pkg/backingstore/reconciler.go
jackyalbo: well I'm currently unaware of any persistent error here. you think about anything specific? thinking about rpc failures that should be temporal - not sure about others... especially when the pool suppose to exist if we got here and it won't be any other type of pool other than hosts

PR #417, opened 2020-09-09
Adding Root secret - dbg

romayalon:

### Explain the changes
1. Added new secret that would be the Root Key for the new encryption feature. choosing random bytes is only temporary implementation.

1. 

### Testing Instructions:
1.

### DIFF
--- a/pkg/system/reconciler.go
+++ b/pkg/system/reconciler.go
@@ -74,6 +74,7 @@ type Reconciler struct {
 	SecretOp            *corev1.Secret
 	SecretAdmin         *corev1.Secret
 	SecretEndpoints     *corev1.Secret
+	SecretRootMasterKey *corev1.Secret
 	AWSCloudCreds       *cloudcredsv1.CredentialsRequest
 	AzureCloudCreds     *cloudcredsv1.CredentialsRequest
 	AzureContainerCreds *corev1.Secret
@@ -123,6 +124,7 @@ func NewReconciler(
 		SecretOp:            util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret),
 		SecretAdmin:         util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret),
 		SecretEndpoints:     util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret),
+		SecretRootMasterKey: util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret),
 		AzureContainerCreds: util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret),
 		GCPBucketCreds:      util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret),
 		AWSCloudCreds:       util.KubeObject(bundle.File_deploy_internal_cloud_creds_aws_cr_yaml).(*cloudcredsv1.CredentialsRequest),
@@ -154,6 +156,7 @@ func NewReconciler(
 	r.SecretOp.Namespace = r.Request.Namespace
 	r.SecretAdmin.Namespace = r.Request.Namespace
 	r.SecretEndpoints.Namespace = r.Request.Namespace
+	r.SecretRootMasterKey.Namespace = r.Request.Namespace
 	r.AzureContainerCreds.Namespace = r.Request.Namespace
 	r.GCPBucketCreds.Namespace = r.Request.Namespace
 	r.AWSCloudCreds.Namespace = r.Request.Namespace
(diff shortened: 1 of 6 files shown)


pkg/system/reconciler.go
```diff
@@ -205,6 +209,12 @@ func NewReconciler(
 	r.SecretServer.StringData["jwt"] = util.RandomBase64(16)
 	r.SecretServer.StringData["server_secret"] = util.RandomHex(4)
 
+	// set noobaa root master key secret
+	r.NooBaaRootSecret.Data["cipher_iv"] = util.RandomBytes(16)
+	r.NooBaaRootSecret.Data["cipher_key"] = util.RandomBytes(32)
```
guymguym, line 214: So I understand that you are using .Data instead of .StringData to store the raw random binary data, and avoid double base64 encoding, because kubernetes already stores it the secret encoded in base64.

One concern that I have with this is if it merges the two maps or just replaces Data when we define StringData... 

The other concern I have is that setting raw binary data in the secret may lead to unprintable StringData...

My instinct is to prefer to use double base64 in this case - and even make the key names represent it clearly that we are communicating b64 encoded data.

```suggestion
	r.NooBaaRootSecret.StringData["cipher_iv_b64"] = util.RandomBase64(16)
	r.NooBaaRootSecret.StringData["cipher_key_b64"] = util.RandomBase64(32)
```

pkg/system/reconciler.go
```diff
@@ -233,6 +243,7 @@ func (r *Reconciler) CheckAll() {
 	util.KubeCheckOptional(r.AWSCloudCreds)
 	util.KubeCheckOptional(r.AzureCloudCreds)
 	util.KubeCheckOptional(r.AzureContainerCreds)
+	util.KubeCheck(r.NooBaaRootSecret)
```
guymguym, line 246: push it above all the optionals, actually best move it with the other secrets.

pkg/system/reconciler.go
```diff
@@ -77,6 +77,7 @@ type Reconciler struct {
 	AWSCloudCreds       *cloudcredsv1.CredentialsRequest
 	AzureCloudCreds     *cloudcredsv1.CredentialsRequest
 	AzureContainerCreds *corev1.Secret
+	NooBaaRootSecret    *corev1.Secret
```
guymguym, line 80: I think the name can be like the other secrets, also move this line to be with the other secrets...
```suggestion
	SecretRootMasterKey    *corev1.Secret
```

pkg/system/reconciler.go
```diff
@@ -177,6 +180,7 @@ func NewReconciler(
 	r.SecretAdmin.Name = r.Request.Name + "-admin"
 	r.SecretEndpoints.Name = r.Request.Name + "-endpoints"
 	r.AzureContainerCreds.Name = r.Request.Name + "-azure-container-creds"
+	r.NooBaaRootSecret.Name = "noobaa-root-master-key"
```
guymguym, line 183:
```suggestion
	r.NooBaaRootSecret.Name = r.Request.Name + "-root-master-key"
```

pkg/system/reconciler.go
```diff
@@ -205,6 +209,12 @@ func NewReconciler(
 	r.SecretServer.StringData["jwt"] = util.RandomBase64(16)
 	r.SecretServer.StringData["server_secret"] = util.RandomHex(4)
 
+	// set noobaa root master key secret
+	r.NooBaaRootSecret.Data["cipher_iv"] = util.RandomBytes(16)
+	r.NooBaaRootSecret.Data["cipher_key"] = util.RandomBytes(32)
+	r.NooBaaRootSecret.StringData["cipher_type"] = "aes-256-gcm"
+	r.NooBaaRootSecret.StringData["description"] = "Root Master Key"
+	r.NooBaaRootSecret.StringData["disabled"] = "false"
```
guymguym, line 217: So this means we need multiple master-keys in this secret in order to do rotation.
Either we can write in the secret a single key with a json payload, or we can make some prefix notation to specify which item we define like:
```
	r.NooBaaRootSecret.StringData["0.cipher_type"] = "aes-256-gcm"
	r.NooBaaRootSecret.StringData["0.description"] = "Root Master Key 2020"
	r.NooBaaRootSecret.StringData["0.disabled"] = "true"

	r.NooBaaRootSecret.StringData["1.cipher_type"] = "aes-256-gcm"
	r.NooBaaRootSecret.StringData["1.description"] = "Root Master Key 2021"
	r.NooBaaRootSecret.StringData["1.disabled"] = "false"
```

guymguym (changes requested):

pkg/system/reconciler.go
romayalon: np, fixed.
also, moved other fields beside cipher_key to the core part (IV removed completely - not needed for the first encryption of the system master key) and updated the endpoint deployment and statefulset-core to set the cipher_key_b64 to env variable.

pkg/system/reconciler.go
romayalon: I'll handle rotation in the next PR so for now I think the current solution is ok

romayalon:
@guymguym 
please re review the recent changes :)

jeniawhite (approved):

PR #424, opened 2020-09-16
Reconcile multiple service monitors with multiple scrape paths

nb-ohad:
For each pod behind the `noobaa-mgmt` service we will scrape 3 paths:
1. `/metrics/web_server` - nodejs `web_server` process report + nooba reports
2. `/metrics/bg_workers` - nodejs `bg_workers` process report
3. `/metrics/hosted_agents`- nodejs `hosted_agents` process report

For each pod behind the `s3` service we will scrape one paths (on port 7001):
1. `/` - nodejs `endpoint` report

### DIFF
--- a/deploy/internal/service-mgmt.yaml
+++ b/deploy/internal/service-mgmt.yaml
@@ -4,6 +4,7 @@ metadata:
   name: SYSNAME-mgmt
   labels:
     app: noobaa
+    noobaa-mgmt-svc: "true"
   annotations:
     prometheus.io/scrape: "true"
     prometheus.io/scheme: http
--- a/deploy/internal/service-s3.yaml
+++ b/deploy/internal/service-s3.yaml
@@ -4,6 +4,7 @@ metadata:
   name: s3
   labels:
     app: noobaa
+    noobaa-s3-svc: "true"
   annotations:
     service.beta.openshift.io/serving-cert-secret-name: 'noobaa-s3-serving-cert'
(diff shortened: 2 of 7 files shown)


nb-ohad:
This PR is marked as [WIP] although the code is complete. The mark is there because the change was not tested yet no an OCS cluster

deploy/internal/service-mgmt.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: SYSNAME-mgmt
   labels:
     app: noobaa
+    noobaa-mgmt: "true"
```
dannyzaken, line 7: maybe use a different key for the label as we already use `noobaa-mgmt` for the pod labels. will be less confusing.  maybe `moobaa-mgmt-svc: "true"`

deploy/internal/service-s3.yaml
```diff
@@ -4,6 +4,7 @@ metadata:
   name: s3
   labels:
     app: noobaa
+    noobaa-s3: "true"
```
dannyzaken, line 7: same. maybe change to `noobaa-s3-svc: "true"`

dannyzaken: if you change the labels don't forget to change the selectors using them

nb-ohad: done

deploy/internal/service-mgmt.yaml
nb-ohad: done

dannyzaken (approved):

PR #428, opened 2020-09-29
added golangci-lint integration to noobaa-operator repo

dannyzaken:
golangci-lint is a linters aggregator for go

* added golangci-lint to GitHub workflows 
* fixed issues reported by golangci-lint (with default configuration)
* added golangci-lint target to `makefile`


nimrod-becker (approved):

liranmauda (approved):

guymguym (commented):
@dannyzaken I am a bit concerned with all those cases where you added return err, and that this is combined into a work on lint fixes, which makes it harder to notice. I didn't yet try to understand if the new error handling makes more sense than the previous..

PR #429, opened 2020-09-29
Set noobaa-db pod's termination grace period to 1min (60sec)

nb-ohad:
This will give mongodb enough time to response to a TERMSIG, sent to the container, and shutdown gracefully, before being forcefuly killed using KILLSIG


nimrod-becker:
LGTM but would like more point of views

nimrod-becker (approved):

dannyzaken (commented):
was there a bug that was found that it should fix?
The default is 30 seconds, did we see it isn't enough?

PR #430, opened 2020-09-30
Adding redirect insecureEdgeTerminationPolicy to management route

jackyalbo:
Fixed BZ1829055


nimrod-becker (approved):

nb-ohad:
@jackyalbo What about the S3 route?

jackyalbo:
@nb-ohad not sure we should do that. The issue was about reaching the mgmt without stating the https protocol in the url. I think for s3 redirecting http to https won't be much more comfortable to the user.

PR #433, opened 2020-10-05
noobaa user design doc

jackyalbo:


jackyalbo:
Few things I'm thinking of about here, if we are thinking of an approach some what similar to bucket access request
1. Maybe to remove the allow_bucket_create - this users will not be able to create new buckets just to use them.
Creating new buckets will only be allowed by creating an obc/br - and a user can ask access to it in allowed_buckets
2. If we allow user to create buckets - Maybe we shouldn't ask for the default pool/backingstore - and all buckets will be created on top of the default bucket class - another option will be to ask for bucketclass and no backingstoe like we did in obc.
3. allowed_buckets will hold only obcs and not any sort of other buckets.

PR #434, opened 2020-10-05
design for noobaa_user

jackyalbo:


jackyalbo:
Few things I'm thinking of about here, if we are thinking of an approach some what similar to bucket access request

Maybe to remove the allow_bucket_create - this users will not be able to create new buckets just to use them.
Creating new buckets will only be allowed by creating an obc/br - and a user can ask access to it in allowed_buckets
If we allow user to create buckets - Maybe we shouldn't ask for the default pool/backingstore - and all buckets will be created on top of the default bucket class - another option will be to ask for bucketclass and no backingstoe like we did in obc.
allowed_buckets will hold only obcs and not any sort of other buckets.

doc/noobaa-user-crd.md
```diff
@@ -0,0 +1,150 @@
+[NooBaa Operator](../README.md) /
+# NooBaaUser CRD
+
+NooBaaUser CRD is used in order to receive new user credantials for accessing different noobaa services (like S3 buckets for example)
```
liranmauda, line 4:
```suggestion
NooBaaUser CRD is used in order to receive new user credentials for accessing different noobaa services (like S3 buckets for example)
```

liranmauda (changes requested):

nb-ohad:
Terminology suggestion: 
A user in my mind is someone who logs in and this terminology is also used for OpenShift Users
In order to prevent confusion, I will suggest changing `NooBaaUser` to `NooBaaAccount`, `NooBaaCredentials`, or `NooBaaAccess`

Regarding Email: 
NooBaa accounts hold 2 different fields for name, an email which is unique and does not have to be of the email format if the account is only for s3 access and a display name, I suggest that the name field will fill the display name field and not the email field (the identifier) and that we should auto-generate that value for the email field

guymguym:
For reference on naming - COSI KEP () uses the name BucketAccess as the entity that provides credentials to the application.
However, this is not necessarily mapping 1-1 with an account or credentials or bucket, as the provider can create accounts/credentials using its own decision (say per cluster or per namespace or per bucket). So I would stick with `NooBaaAccount` - this is consistent with core and means exactly what we do.

jackyalbo:
Creating this PR - in order the check that my tests work - On my local I fail on something irrelevant to this PR - probably too slow.

liranmauda (approved):
test

doc/noobaa-account-crd.md
```diff
@@ -0,0 +1,151 @@
+[NooBaa Operator](../README.md) /
+# NooBaaAccount CRD
+
+NooBaaAccount CRD is used in order to receive new credentials set for accessing different noobaa services (like S3 buckets for example)
+
+This are the main fields which can be provided to this CRD:
+- AllowedBuckets - if fullPermission is true - all current and future buckets will be allowed full access by this account.
+- AllowedBuckets - if fullPermission is false - all the buckets in the permissionList will be allowed full access by this account.
+- AllowBucketCreate - the account will be able to create new buckets on top of the account's default backingstore
+- DefaultBackingStore - the name of the backingstore new buckets created by the account will be created on
```
romayalon, line 10: need update for default pool, and what about nsfs account config?

doc/noobaa-account-crd.md
```diff
@@ -0,0 +1,151 @@
...
+  labels:
+    app: noobaa
+  name: account1
+  namespace: noobaa
+spec:
+  allowedBuckets:
+    permissionList: 
+    - first.bucket
+  allowBucketCreate: true
+  defaultBackingStore: noobaa-default-backing-store
```
romayalon, line 96: update default backingstore to default pool in the examples too

pkg/noobaaaccount/reconciler.go
```diff
@@ -0,0 +1,467 @@
...
+		if err != nil {
+			return err
+		}
+		log.Infof("✅ Successfully updated account %q", r.NooBaaAccount.Name)
+	}
+
+	return nil
+}
+
+func (r *Reconciler) needUpdate() bool {
```
romayalon, line 408: don't we want it to be an update predicate in the account controller?

PR #444, opened 2020-10-15
Design of CRD expansion for namespace feature

romayalon:

Initial Design for NamespaceStore CRD and the expansion of bucketclass CRD for supporting namespace feature.

### DIFF
--- a/doc/bucket-class-crd.md
+++ b/doc/bucket-class-crd.md
@@ -1,18 +1,37 @@
 [NooBaa Operator](../README.md) /
 # BucketClass CRD
 
-NooBaaBucket CRD represents a class for buckets that defines policies for data placement and more.
+NooBaaBucket CRD represents a class for buckets that defines policies for data placement, namespace, and more.
 
 Data placement capabilities are built as a multi-layer structure, here are the layers bottom-up:
 - Spread Layer - list of backing-stores, aggregates the storage of multiple stores.
 - Mirroring Layer - list of spread-layers, async-mirroring to all mirrors, with locality optimization (will allocate on the closest region to the source endpoint), mirroring requires at least two backing-stores.
 - Tiering Layer - list of mirroring-layers, push cold data to next tier.
 
+Namespace policy:
+A namespace bucket-class will define a policy for namespace buckets.
+Namespace policy will require a type, the type's value can be one of the following: single, multi, cache.
+
+
+Namespace policy of type single will require the following configuration:
+  - Resource - a single namespace-store, defines the read and write target of the namespace bucket.
+
+Namespace policy of type multi will require the following configuration:
+- Read Resources - list of namespace-stores, defines the read targets of the namespace bucket.
+- Write Resource - a single namespace-store, defines the write target of the namespace bucket.
+
+Namespace policy of type cache will require the following configuration:
+  - Hub Resource - a single namespace-store, defines the read and write target of the namespace bucket.
+  - Hub Prefix - defines the prefix of the future cached data.
+  - TTL - defines the TTL of the cached data.
+
 Constraints:
 A backing-store name may appear in more than one bucket-class but may not appear more than once in a single bucket-class.
 The operator cli currently only supports a single tier placement-policy for a bucket-class. 
+A bucket class must have only one of Placement-policy or Namespace-policy.
 YAML must be used to create a bucket-class with a placement-policy that has multiple tiers.
 Placement-policy is case sensitive and should be of value (Mirror|Spread).
+Namespace-policy is case sensitive.
 
 For more information on using bucket-classes from S3 see [S3 Account](s3-account.md).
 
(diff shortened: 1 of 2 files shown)


doc/bucket-class-crd.md
```diff
@@ -1,18 +1,29 @@
 [NooBaa Operator](../README.md) /
 # BucketClass CRD
 
-NooBaaBucket CRD represents a class for buckets that defines policies for data placement and more.
+NooBaaBucket CRD represents a class for buckets that defines policies for data placement, namespace and more.
```
liranmauda, line 4:
```suggestion
NooBaaBucket CRD represents a class for buckets that defines policies for data placement, namespace, and more.
```

doc/namespace-store-crd.md
```diff
@@ -0,0 +1,218 @@
...
+      name: namespace-store-azure-blob-bs
+      namespace: noobaa
+    targetBlobContainer: CONTAINER
+  type: azure-blob
+```
+
+#### Credentials change
+
+In case the credentials of a Namespace-store need to be updated due to a periodic security policy or concern, the appropriate secret should be updated by the user, and the operator will be responsible for watching changes in those secrets and propagating the new credential update to the NooBaa system server.
+
```
liranmauda, line 125:
```suggestion
In case the credentials of a Namespace-store need to be updated due to periodic security policy or concern, the appropriate secret should be updated by the user, and the operator will be responsible for watching changes in those secrets and propagating the new credential update to the NooBaa system server.

```

doc/namespace-store-crd.md
```diff
@@ -0,0 +1,218 @@
...
+
+# Read Status
+
+Here is an example healthy status (see below example of non-healthy status):
```
liranmauda, line 129:
```suggestion
Here is an example of healthy status (see below an example of non-healthy status):
```

doc/namespace-store-crd.md
```diff
@@ -0,0 +1,218 @@
...
+# Delete
+
+Namespace-Store are used for data persistency, therefore there is a cleanup process before they can be deleted.
+The operator will use the `finalizer` pattern as explained in the link below, and set a finalizer on every Namespace-store to mark that external cleanup is needed before it can be delete:
+
+https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers
+
+After marking a Namespace-store for deletion, the operator will notify the NooBaa server on the deletion which will enter a *decommissioning* state, in which NooBaa will attempt to rebuild the data to a new Namespace-store location. Once the decomissioning process completes the operator will remove the finalizer and allow the CR to be deleted.
+
+There are cases where the decommissioning cannot complete due to inability to read the data from the Namespace-store that is already not serving - for example if the target bucket was already deleted or the credentials were invalidated or there is no network from the system to the Namespace-store service. In such cases the system status will be used to report these issues and suggest manual resolution for example:
```
liranmauda, line 178:
```suggestion
Namespace-Store is used for data persistency, therefore there is a cleanup process before they can be deleted.
The operator will use the `finalizer` pattern as explained in the link below, and set a finalizer on every Namespace-store to mark that external cleanup is needed before it can be deleted:

https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#finalizers

After marking a Namespace-store for deletion, the operator will notify the NooBaa server on the deletion which will enter a *decommissioning* state, in which NooBaa will attempt to rebuild the data to a new Namespace-store location. Once the decommissioning process completes the operator will remove the finalizer and allow the CR to be deleted.

There are cases where the decommissioning cannot complete due to the inability to read the data from the Namespace-store that is already not serving - for example, if the target bucket was already deleted or the credentials were invalidated or there is no network from the system to the Namespace-store service. In such cases the system status will be used to report these issues and suggest a manual resolution for example:
```

doc/bucket-class-crd.md
```diff
@@ -177,3 +188,46 @@ spec:
...
+  namespace: noobaa
+spec:
+  namespacePolicy:
+    readResources:
+    - bs1
+    - bs2
+    writeResource: bs1 
+    cache:
+      prefix: 'a'
+      ttl: 6000
```
guymguym, line 232: suggestion - design it like we do for backing-stores and namespace-stores:

```
  namespacePolicy:
    type: cache
    cache:
      hub: ns1
      prefix: aaa/
      ttl: 6000
```

```
  namespacePolicy:
    type: merge
    merge:
      writeResource: ns1
      readResources: 
      - ns1
      - ns2
```

```
  namespacePolicy:
    type: s3
    s3:
      resource: ns1
```

guymguym: updated the above suggestion ^

doc/bucket-class-crd.md
romayalon: thanks!

doc/bucket-class-crd.md
romayalon: thanks for the suggestion, updated.

nimrod-becker (approved):

PR #446, opened 2020-10-20
In IBM Cloud configure IBM COS as default backing store

nkkashyap:
Signed-off-by: Neeraj K Kashyap <nkashyap@in.ibm.com>

This PR enables the configuration of IBM COS as backing store for NooBaa in the IBM Cloud (IKS /ROKS).
Currently, Cloud Credential Operator is not supporting IBM Cloud creds. The code change assumes a secret `ibm-cloud-cos-creds`, under `kube-system`, will be created by Admin or by the IBM Cloud at the time of cluster creation
```
apiVersion: v1
kind: Secret
metadata:
  name: ibm-cloud-cos-creds
  namespace: kube-system
type: Opaque
data:
  IBM_COS_ACCESS_KEY_ID: RlM2I4ODQ=
  IBM_COS_SECRET_ACCESS_KEY: NjNjNmEzZTNk
```

The code change has been tested in IBM Cloud
```
$ oc get backingstore
NAME                           TYPE      PHASE   AGE
noobaa-default-backing-store   ibm-cos   Ready   23m
```

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -23,11 +23,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-const (
-	ibmCredSecret   = "ibm-cloud-cos-creds"
-	ibmCredSecretNS = "kube-system"
-)
-
 // ReconcilePhaseCreating runs the reconcile phase
 func (r *Reconciler) ReconcilePhaseCreating() error {
 
@@ -596,14 +591,12 @@ func (r *Reconciler) ReconcileGCPCredentials() error {
 	return err
 }
 
-// ReconcileIBMCredentials create dummy request
+// ReconcileIBMCredentials sets IsIBMCloud to indicate operator is running in IBM Cloud
 func (r *Reconciler) ReconcileIBMCredentials() error {
 	// Currently IBM Cloud is not supported by cloud credential operator
-	// In IBM Cloud, the HMAC keys will be provided through K8S Secret under kube-system namespace
-	r.Logger.Info("Running in IBM Cloud. Expecting Secret: <ibm-cloud-cos-creds> under NS: <kube-system>")
-	r.IBMCloudCreds.Spec.SecretRef.Name = ibmCredSecret
-	r.IBMCloudCreds.Spec.SecretRef.Namespace = ibmCredSecretNS
-	r.IBMCloudCreds.UID = "dummy-uid"
+	// In IBM Cloud, the COS Creds will be provided through Secret.
+	r.Logger.Info("Running in IBM Cloud")
+	r.IsIBMCloud = true
 	return nil
 }
 
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -23,6 +23,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	ibmCredSecret   = "ibm-cloud-cos-creds"
+	ibmCredSecretNS = "kube-system"
+)
+
 // ReconcilePhaseCreating runs the reconcile phase
 func (r *Reconciler) ReconcilePhaseCreating() error {
 
@@ -419,6 +424,7 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 // the bucket name allowed for the credentials. nil is returned if cloud credentials are not supported
 func (r *Reconciler) ReconcileBackingStoreCredentials() error {
 	// Skip if joining another NooBaa
+	r.Logger.Info("Reconciling Backing Store Credentials")
 	if r.JoinSecret != nil {
 		return nil
 	}
@@ -431,6 +437,9 @@ func (r *Reconciler) ReconcileBackingStoreCredentials() error {
 	}
 	if util.IsGCPPlatform() {
 		return r.ReconcileGCPCredentials()
+	}
+	if util.IsIBMPlatform() {
+		return r.ReconcileIBMCredentials()
 	}
 	return r.ReconcileRGWCredentials()
 }
@@ -585,6 +594,17 @@ func (r *Reconciler) ReconcileGCPCredentials() error {
 		return nil
 	}
 	return err
+}
+
+// ReconcileIBMCredentials create dummy request
+func (r *Reconciler) ReconcileIBMCredentials() error {
+	// Currently IBM Cloud is not supported by cloud credential operator
+	// In IBM Cloud, the HMAC keys will be provided through K8S Secret under kube-system namespace
+	r.Logger.Info("Running in IBM Cloud. Expecting Secret: <ibm-cloud-cos-creds> under NS: <kube-system>")
+	r.IBMCloudCreds.Spec.SecretRef.Name = ibmCredSecret
+	r.IBMCloudCreds.Spec.SecretRef.Namespace = ibmCredSecretNS
+	r.IBMCloudCreds.UID = "dummy-uid"
+	return nil
 }
 
 // SetDesiredAgentProfile updates the value of the AGENT_PROFILE env
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -33,8 +33,10 @@ import (
 )
 
 const (
-	ibmEndpoint = "https://s3.direct.%s.cloud-object-storage.appdomain.cloud"
-	ibmLocation = "%s-standard"
+	ibmEndpoint         = "https://s3.direct.%s.cloud-object-storage.appdomain.cloud"
+	ibmLocation         = "%s-standard"
+	ibmCOSCred          = "ibm-cloud-cos-creds"
+	ibmCOSCredDefaultNS = "kube-system"
 )
 
 type gcpAuthJSON struct {
@@ -465,8 +467,8 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
 		if err := r.prepareGCPBackingStore(); err != nil {
 			return err
 		}
-	} else if r.IBMCloudCreds.UID != "" {
-		log.Infof("CredentialsRequest %q created. Creating default backing store on IBM objectstore", r.IBMCloudCreds.Name)
+	} else if r.IsIBMCloud {
+		log.Infof("Running in IBM Cloud. Creating default backing store on IBM objectstore")
 		if err := r.prepareIBMBackingStore(); err != nil {
 			return err
 		}
@@ -697,16 +699,21 @@ func (r *Reconciler) prepareIBMBackingStore() error {
 
 	cloudCredsSecret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      r.IBMCloudCreds.Spec.SecretRef.Name,
-			Namespace: r.IBMCloudCreds.Spec.SecretRef.Namespace,
+			Name:      r.IBMCloudCOSCreds.Name,
+			Namespace: r.IBMCloudCOSCreds.Namespace,
 		},
 	}
 
 	util.KubeCheck(cloudCredsSecret)
 	if cloudCredsSecret.UID == "" {
-		// Secret not found
-		r.Logger.Errorf("Cloud credentials secret %q is not ready yet", r.IBMCloudCreds.Spec.SecretRef.Name)
-		return fmt.Errorf("Cloud credentials secret %q is not ready yet", r.IBMCloudCreds.Spec.SecretRef.Name)
+		// Secret not found, check the default ns
+		cloudCredsSecret.Namespace = ibmCOSCredDefaultNS
+		util.KubeCheck(cloudCredsSecret)
+		if cloudCredsSecret.UID == "" {
+			r.Logger.Errorf("Cloud credentials secret %q is not ready yet", r.IBMCloudCOSCreds.Name)
+			return fmt.Errorf("Cloud credentials secret %q is not ready yet", r.IBMCloudCOSCreds.Name)
+		}
+		r.IBMCloudCOSCreds.Namespace = ibmCOSCredDefaultNS
 	}
 
 	if val, ok := cloudCredsSecret.StringData["IBM_COS_Endpoint"]; ok {
@@ -715,6 +722,7 @@ func (r *Reconciler) prepareIBMBackingStore() error {
 		r.Logger.Infof("Endpoint provided in secret: %q", endpoint)
 		if val, ok := cloudCredsSecret.StringData["IBM_COS_Location"]; ok {
 			location = val
+			r.Logger.Infof("Location provided in secret: %q", location)
 		}
 	} else {
 		// Endpoint not provided in the secret, construct one based on the cluster's region
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -32,6 +32,11 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 )
 
+const (
+	ibmEndpoint = "https://s3.direct.%s.cloud-object-storage.appdomain.cloud"
+	ibmLocation = "%s-standard"
+)
+
 type gcpAuthJSON struct {
 	ProjectID string `json:"project_id"`
 }
@@ -440,24 +445,29 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
 		return nil
 	}
 
-	if r.CephObjectStoreUser.UID != "" {
-		log.Infof("CephObjectStoreUser %q created.  creating default backing store on ceph objectstore", r.CephObjectStoreUser.Name)
+	if r.CephObjectstoreUser.UID != "" {
+		log.Infof("CephObjectstoreUser %q created. Creating default backing store on ceph objectstore", r.CephObjectstoreUser.Name)
 		if err := r.prepareCephBackingStore(); err != nil {
 			return err
 		}
 	} else if r.AWSCloudCreds.UID != "" {
-		log.Infof("CredentialsRequest %q created.  creating default backing store on AWS objectstore", r.AWSCloudCreds.Name)
+		log.Infof("CredentialsRequest %q created. Creating default backing store on AWS objectstore", r.AWSCloudCreds.Name)
 		if err := r.prepareAWSBackingStore(); err != nil {
 			return err
 		}
 	} else if r.AzureCloudCreds.UID != "" {
-		log.Infof("CredentialsRequest %q created.  creating default backing store on Azure objectstore", r.AzureCloudCreds.Name)
+		log.Infof("CredentialsRequest %q created. Creating default backing store on Azure objectstore", r.AzureCloudCreds.Name)
 		if err := r.prepareAzureBackingStore(); err != nil {
 			return err
 		}
 	} else if r.GCPCloudCreds.UID != "" {
 		log.Infof("CredentialsRequest %q created.  creating default backing store on GCP objectstore", r.GCPCloudCreds.Name)
 		if err := r.prepareGCPBackingStore(); err != nil {
+			return err
+		}
+	} else if r.IBMCloudCreds.UID != "" {
+		log.Infof("CredentialsRequest %q created. Creating default backing store on IBM objectstore", r.IBMCloudCreds.Name)
+		if err := r.prepareIBMBackingStore(); err != nil {
 			return err
 		}
 	} else {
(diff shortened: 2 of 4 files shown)


romayalon:
Hey @nkkashyap,
thanks for your contribution!
Have you also tested the creation of the target bucket on IBM cloud and creating an OBC on top of the default backingstore and reading/writing to the OBC?

nkkashyap:
Hi @romayalon 
I have tested and verified the OBC usecase
Here are the details
```
$ oc get backingstore
NAME                           TYPE      PHASE   AGE
noobaa-default-backing-store   ibm-cos   Ready   4d1h
```
```
$ oc create -f - <<EOF
apiVersion: noobaa.io/v1alpha1
kind: BucketClass
metadata:
  name: ibmcos-bucket-class
  labels:
    app: noobaa
spec:
  placementPolicy:
    tiers:
    - backingStores:
      - noobaa-default-backing-store
EOF
```
```
 oc get bucketclass
NAME                          PLACEMENT                                                        PHASE   AGE
ibmcos-bucket-class           map[tiers:[map[backingStores:[noobaa-default-backing-store]]]]   Ready   4s
noobaa-default-bucket-class   map[tiers:[map[backingStores:[noobaa-default-backing-store]]]]   Ready   4d2h
```

```
$ oc create -f - <<EOF 
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ibmcos-noobaa.noobaa.io
provisioner: openshift-storage.noobaa.io/obc
parameters:
  bucketclass: ibmcos-bucket-class
reclaimPolicy: Delete
EOF
```
```
$ oc get sc ibmcos-noobaa.noobaa.io
NAME                      PROVISIONER                       RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
ibmcos-noobaa.noobaa.io   openshift-storage.noobaa.io/obc   Delete          Immediate           false                  2m26s
```

Create a OBC under `ibmcos` NS
```
$ oc new-project ibmcos 

$ oc create -f - <<EOF
apiVersion: objectbucket.io/v1alpha1
kind: ObjectBucketClaim
metadata:
  name: ibmcos01
spec:
  storageClassName: ibmcos-noobaa.noobaa.io
  bucketName: ibmcos01.test
  additionalConfig: {}
EOF

$ oc get ob
NAME                  STORAGE-CLASS             CLAIM-NAMESPACE   CLAIM-NAME   RECLAIM-POLICY   PHASE   AGE
obc-ibmcos-ibmcos01   ibmcos-noobaa.noobaa.io   ibmcos            ibmcos01     Delete           Bound   2m40s
```
Create Test POD
```
$ cat pod-devtest.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: pod-devtest-01
  namespace: ibmcos
spec:
  containers:
  - name: app
    envFrom:
    - secretRef:
        name: ibmcos01
    - configMapRef:
        name: ibmcos01
    image: banst/awscli
    imagePullPolicy: Always
    command:
    - sh
    - "-c"
    - |
      while true; do sleep 60; done;
```
```
$ oc -n ibmcos create -f ./pod-devtest.yaml 

$ oc -n ibmcos exec -it pod-devtest-01 -- sh
/ # pip install awscli-plugin-endpoint;
/ # aws configure set plugins.endpoint awscli_plugin_endpoint;
/ # aws configure set s3.endpoint_url https://s3.openshift-storage.svc
/ # aws configure set ca_bundle /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt;
/ # aws s3 ls s3://
2020-10-25 13:54:16 ibmcos01.test
/ # aws s3 ls s3://ibmcos01.test/
/ # aws s3 cp --recursive /etc s3://ibmcos01.test/
upload: etc/alpine-release to s3://ibmcos01.test/alpine-release 
/ # aws s3 ls s3://ibmcos01.test/
                           PRE apk/
                           PRE ca-certificates/
                           PRE crontabs/
                           PRE logrotate.d/
                           PRE modprobe.d/
                           PRE network/
                           PRE profile.d/
                           PRE ssl/
                           PRE sysctl.d/
                           PRE terminfo/
```

nkkashyap:
From local system
```
$ export ENDPOINT="https://s3-openshift-storage.perf-ocs-nk02-dee38fa0a303a62a413a3201238c532b-0000.us-south.containers.appdomain.cloud"
$ export AWS_ACCESS_KEY_ID="XYZ"
$ export AWS_SECRET_ACCESS_KEY="1234"
$ ~/Library/Python/3.7/bin/aws --endpoint-url=${ENDPOINT} s3 ls s3://
2020-10-25 19:39:44 ibmcos01.test
$ ~/Library/Python/3.7/bin/aws --endpoint-url=${ENDPOINT} s3 ls s3://ibmcos01.test
                           PRE apk/
                           PRE ca-certificates/
                           PRE crontabs/
                           PRE logrotate.d/
                           PRE modprobe.d/
                           PRE network/
                           PRE profile.d/
                           PRE ssl/
                           PRE sysctl.d/
```

pkg/system/reconciler.go
```diff
@@ -128,6 +129,7 @@ func NewReconciler(
 		AWSCloudCreds:       util.KubeObject(bundle.File_deploy_internal_cloud_creds_aws_cr_yaml).(*cloudcredsv1.CredentialsRequest),
 		AzureCloudCreds:     util.KubeObject(bundle.File_deploy_internal_cloud_creds_azure_cr_yaml).(*cloudcredsv1.CredentialsRequest),
 		GCPCloudCreds:       util.KubeObject(bundle.File_deploy_internal_cloud_creds_gcp_cr_yaml).(*cloudcredsv1.CredentialsRequest),
+		IBMCloudCreds:       util.KubeObject(bundle.File_deploy_internal_cloud_creds_aws_cr_yaml).(*cloudcredsv1.CredentialsRequest),
```
jeniawhite, line 132: I do not see a need for this secret.
Even if you eventually decide to create a secret, it shouldn't be an AWS credential request (cloud credentials operator doesn't support IBM COS).
You can just query and use the secret `ibm-cloud-cos-creds` under `kube-system` directly, also link the BackingStore to it.
Just make sure that it exists and created by the OCP.

pkg/system/phase4_configuring.go
```diff
@@ -677,6 +687,104 @@ func (r *Reconciler) prepareGCPBackingStore() error {
...
+		r.Logger.Errorf("Cloud credentials secret %q is not ready yet", r.IBMCloudCreds.Spec.SecretRef.Name)
+		return fmt.Errorf("Cloud credentials secret %q is not ready yet", r.IBMCloudCreds.Spec.SecretRef.Name)
+	}
+
+	if val, ok := cloudCredsSecret.StringData["IBM_COS_Endpoint"]; ok {
+		// Use the endpoint provided in the secret
+		endpoint = val
+		r.Logger.Infof("Endpoint provided in secret: %q", endpoint)
+		if val, ok := cloudCredsSecret.StringData["IBM_COS_Location"]; ok {
+			location = val
```
jeniawhite, line 717: Any reason not to add logging for `location` as well?
I would maybe even log them later on after the assignment.

jeniawhite (changes requested):

pkg/system/reconciler.go
guymguym: @jeniawhite What? How can we access kube-system? Why would we?

jeniawhite: @guymguym 
The cloud credentials operator does not support IBM COS.
The code pulls the credentials from `kube-system` namespace secret `ibm-cloud-cos-creds`.
This secret gets created by OCP when deployed over an IBM cluster environment.
Seems like doing a ref to that secret is allowed (I assume since it looks like it worked in the comments).

guymguym: I see. So it relies on the cluster-permission we have to read and write secrets
here - https://github.com/noobaa/noobaa-operator/blob/681687b5340d16626e4bfd7720b144eb7c99eaf5/deploy/cluster_role.yaml#L24-L31

But the fact that we added this permission was because OBC provisioner did not have any central controller that would be entrusted with the cluster-wide secrets permission. But going forward COSI which will replace OBC will not require us to have cluster-wide secret access, which I perceive as "not our business". 

Isn't OCP going to expose this somehow with cloud-credentials-operator, or any other separate operator that will do that for IBM cloud?

Adding @nimrod-becker

nimrod-becker: Cloud Creds Op won't be minting IBM cloud creds.

There is a talk about moving to a whole different model of security, in which each pod has its own identify and thus can access whatever services it needs. Thought it's not relevant for now, in order for the default to work until the new solution is implemented we need this special handling/

guymguym:
@nkkashyap Hey, Many thanks for your work. I appreciate it! :)

I wanted to ask, not sure if was asked in the previous comments, but is it possible that this secret be created in the target namespace for OCS/noobaa (by default this is `openshift-storage` for OCS on Openshift). I wouldn't want to rely on having a cluster-wide access to secrets.

pkg/system/reconciler.go
guymguym: Got it. I just added a new comment with a question is if it's possible that this secret be created in our namespace instead of on a kube-system namespace.

nimrod-becker: I think its a chicken and egg thing. The secret needs to be created during deployment of OCP when there is still no NS for OCS ...

guymguym: Our operator won't care if the secret will be created a bit after it started.
It can even be a long time after, it should only wakeup and reconcile.
So at least theoretically it can be imported to the `openshift-storage` namespace only later on when OCS installs/configures.

nimrod-becker: I Don't know if the creds can change or not, if they can just a copy won't be good enough

nkkashyap: Hi @jeniawhite @guymguym @nimrod-becker 
Thanks for your review comments. 
I have changed the code to look for `ibm-cloud-cos-creds` under the target namespace, and also under the default `openshift-storage` namespace. 
Also updated the code not to use AWS credential request for IBM Cloud.

pkg/system/phase4_configuring.go
nkkashyap: Update the code to log `location` as well.

nkkashyap:
> @nkkashyap Hey, Many thanks for your work. I appreciate it! :)
> 
> I wanted to ask, not sure if was asked in the previous comments, but is it possible that this secret be created in the target namespace for OCS/noobaa (by default this is `openshift-storage` for OCS on Openshift). I wouldn't want to rely on having a cluster-wide access to secrets.

Hi @guymguym 
In ROKS (managed OCP in IBM Cloud), the required secret `ibm-cloud-cos-creds` will be created by provider just after OCP cluster creation and before OCS deployment.  `noonbaa` is getting deployed under `openshift-storage` NS by the OCS on ROKS. The IBM Cloud provider can be update to create secret under  `openshift-storage` instead of `kube-system`.

pkg/system/reconciler.go
```diff
@@ -128,6 +130,7 @@ func NewReconciler(
 		AWSCloudCreds:       util.KubeObject(bundle.File_deploy_internal_cloud_creds_aws_cr_yaml).(*cloudcredsv1.CredentialsRequest),
 		AzureCloudCreds:     util.KubeObject(bundle.File_deploy_internal_cloud_creds_azure_cr_yaml).(*cloudcredsv1.CredentialsRequest),
 		GCPCloudCreds:       util.KubeObject(bundle.File_deploy_internal_cloud_creds_gcp_cr_yaml).(*cloudcredsv1.CredentialsRequest),
+		IBMCloudCOSCreds:    util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret),
```
jeniawhite, line 133: Why do you need this secret?
You do not reconcile or update it in any way.

pkg/system/phase4_configuring.go
```diff
@@ -677,6 +689,106 @@ func (r *Reconciler) prepareGCPBackingStore() error {
...
 
+func (r *Reconciler) prepareIBMBackingStore() error {
+	r.Logger.Info("Preparing backing store in IBM Cloud")
+
+	var (
+		endpoint string
+		location string
+	)
+
+	util.KubeCheck(r.IBMCloudCOSCreds)
```
jeniawhite, line 700: You can just get the secret using (example):
`
			secret := &corev1.Secret{
				TypeMeta: metav1.TypeMeta{Kind: "Secret"},
				ObjectMeta: metav1.ObjectMeta{
					Name:      Name,
					Namespace: Namespace,
				},
			}
if !util.KubeCheck(secret)
`
That way you will avoid adding secrets to the reconcile that you do not update or store anything in them.

pkg/system/phase4_configuring.go
```diff
@@ -677,6 +689,106 @@ func (r *Reconciler) prepareGCPBackingStore() error {
...
+			accessKeyID,
+			secretAccessKey,
+			"",
+		),
+		Region: &location,
+	}
+	if err := r.createS3BucketForBackingStore(s3Config, bucketName); err != nil {
+		return err
+	}
+	r.Logger.Infof("Creadted bucket: %s", bucketName)
```
jeniawhite, line 776: Created*

pkg/system/phase4_configuring.go
```diff
@@ -677,6 +689,106 @@ func (r *Reconciler) prepareGCPBackingStore() error {
...
+
+	// create backing store
+	r.DefaultBackingStore.Spec.Type = nbv1.StoreTypeIBMCos
+	r.DefaultBackingStore.Spec.IBMCos = &nbv1.IBMCosSpec{
+		TargetBucket: bucketName,
+		Secret: corev1.SecretReference{
+			Name:      r.IBMCloudCOSCreds.Name,
```
jeniawhite, line 783: Just use the name and namespace from above. No need for the reconcile object.

pkg/system/reconciler.go
```diff
@@ -194,6 +198,8 @@ func NewReconciler(
 	r.GCPBucketCreds.Name = r.Request.Name + "-gcp-bucket-creds"
 	r.GCPCloudCreds.Name = r.Request.Name + "-gcp-cloud-creds"
 	r.GCPCloudCreds.Spec.SecretRef.Name = r.Request.Name + "-gcp-cloud-creds-secret"
+	r.IsIBMCloud = false
```
jeniawhite, line 201: This can also be removed and replaced by util function call until the cloud credentials operator will support IBM COS.

jeniawhite (changes requested):

pkg/system/phase4_configuring.go
```diff
@@ -32,6 +32,13 @@ import (
 	"github.com/aws/aws-sdk-go/service/s3"
 )
 
+const (
+	ibmEndpoint         = "https://s3.direct.%s.cloud-object-storage.appdomain.cloud"
+	ibmLocation         = "%s-standard"
+	ibmCOSCred          = "ibm-cloud-cos-creds"
+	ibmCOSCredDefaultNS = "openshift-storage"
```
guymguym, line 39: I think we should avoid defining a default namespace for these credentials.
It should always be loaded from the namespace of the NooBaa system so that it is consistent with the RBAC permissions we require.
For example - in case we deploy noobaa on another namespace, why should it default to trying openshift-storage namespace for credentials lookup - these credentials were not given to it, and it shouldn't access it.
My suggestion is that the operator will lookup this fixed name secret in the noobaa system namespace.

pkg/system/phase4_configuring.go
guymguym: You are right, but I woudn't do that.
I think the current model is that all backing stores that have credentials are loaded into the reconciler struct for ease of use between the different phases.
So I would keep this working the same as the AWS and other credentials work.

pkg/system/reconciler.go
guymguym: There is a util function, but it uses api calls to kube to determine that, so we run that in the reconcile loop.
I wonder if we should run that just once and then keep the result instead of all these wasteful calls every reconcile cycle...

guymguym: In any case I wouldn't call that util function in this ctor, right?

jeniawhite: I've meant to remove this value and call the util function that exists in the single place where we look at that value.
But if you are saying that we do not want to trigger another API call, then we can leave it like that.

pkg/system/phase4_configuring.go
```diff
@@ -460,6 +467,11 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
 		if err := r.prepareGCPBackingStore(); err != nil {
 			return err
 		}
+	} else if r.IsIBMCloud {
```
guymguym, line 470: Following up on this flag - why not use the same model as the other providers and just check if `r. IBMCloudCOSCreds.UID != ""` ?
Is the COS flow so different than the other cases?

pkg/system/phase2_creating.go
```diff
@@ -503,6 +507,15 @@ func (r *Reconciler) ReconcileGCPCredentials() error {
 	return err
 }
 
+// ReconcileIBMCredentials sets IsIBMCloud to indicate operator is running in IBM Cloud
+func (r *Reconciler) ReconcileIBMCredentials() error {
+	// Currently IBM Cloud is not supported by cloud credential operator
+	// In IBM Cloud, the COS Creds will be provided through Secret.
+	r.Logger.Info("Running in IBM Cloud")
+	r.IsIBMCloud = true
```
guymguym, line 515: Why aren't we loading the credentials secret here?

pkg/system/phase4_configuring.go
jeniawhite: The thing is that in other credentials we are actually using these secrets in order to pull the credentials from the operator and assign them to our secret that won't get reconciled (with the addition of our information).
In this case, we do not do anything with the secret that we've created in the reconciler, since we just look and REF to the secret that we've listed.
So this is why I've said that we can remove this secret, but if you are saying that you prefer not to do the Kube check as I've suggested above, then we can leave this.

guymguym (changes requested):
@sacharya See comments above. I think it would be good to setup in a similar manner to the other providers unless there is a different flow needed - in that case it would be great if you can explain. Thanks

pkg/system/phase2_creating.go
jeniawhite: You are right, we can actually load the credentials here instead of listing them in Phase4.
They should be ready here since we do not need to create cloud credentials CR and wait for it to reconcile.

pkg/system/reconciler.go
guymguym: ok so I think the other comment I wrote here - https://github.com/noobaa/noobaa-operator/pull/446#pullrequestreview-536328625 is more relevant now.

pkg/system/phase4_configuring.go
jeniawhite: In other clouds, we are check that we've created the cloud operator credentials request CR.
In this case, we aren't creating a CR, we can obviously check if the secret exists or not and assume that it won't exist on other deployments, or check that we've created the IBM secret in our Phase2 (if we want to stick with the current code that has a "useless" secret in the reconciler)

jeniawhite (changes requested):

pkg/system/phase4_configuring.go
guymguym: @jeniawhite lets expand what you are saying here - should we detect the IBM cloud platform and then create the default backing store directly with the fixed secret name, without even trying to load it first? In case the secret does not exist this means our default backing store will already be created this way and will wait for the secret to appear.
On the other hand, if we do want the system reconciler to check the secret exists before deciding what is the default backing store - saw we want to decide to create a different backing store, or some other logic in the scope of the system reconciler.

I actually don't mind which option is preferred, but I would like to streamline with the rest of the cloud platform flows. So I would still write it just like we do for AWS creds and if we decide to refactor later we can do all platforms the same.

pkg/system/phase4_configuring.go
nkkashyap: Fixed the typo

pkg/system/phase4_configuring.go
nkkashyap: Removed the default namespace for cred secrets

guymguym:
Thanks @nkkashyap . I won't be able to go over this in more detail, but I trust you and @jeniawhite can make sure that the operator will keep a simple and consistent reconcile path. Think about it from the point of view of the next provider we want to add looking at the other providers including the one we just added here. Thanks

nkkashyap:
Resolved the merge conflicts. Tested and verified the OBC usecase with latest code change
```
$ nb --operator-image='nkkashyap/noobaa-operator:5.7.0' --namespace='openshift-storage' install
INFO[0000] CLI version: 5.7.0                           
INFO[0000] noobaa-image: noobaa/noobaa-core:5.6.0-20201206 
INFO[0000] operator-image: nkkashyap/noobaa-operator:5.7.0 
INFO[0000] Namespace: openshift-storage                 
INFO[0000]                                              
INFO[0000] CRD Create:                                  
INFO[0013] ✅ Created: CustomResourceDefinition "noobaas.noobaa.io" 
INFO[0014] ✅ Created: CustomResourceDefinition "backingstores.noobaa.io" 
INFO[0017] ✅ Created: CustomResourceDefinition "namespacestores.noobaa.io" 
INFO[0018] ✅ Created: CustomResourceDefinition "bucketclasses.noobaa.io" 
INFO[0019] ✅ Created: CustomResourceDefinition "objectbucketclaims.objectbucket.io" 
INFO[0020] ✅ Created: CustomResourceDefinition "objectbuckets.objectbucket.io" 
INFO[0020]                                              
INFO[0020] Operator Install:                            
INFO[0020] ✅ Already Exists: Namespace "openshift-storage" 
INFO[0021] ✅ Created: ServiceAccount "noobaa"           
INFO[0022] ✅ Created: Role "noobaa"                     
INFO[0024] ✅ Created: RoleBinding "noobaa"              
INFO[0024] ✅ Created: ClusterRole "openshift-storage.noobaa.io" 
INFO[0025] ✅ Created: ClusterRoleBinding "openshift-storage.noobaa.io" 
INFO[0026] ✅ Created: Deployment "noobaa-operator"      
INFO[0026]                                              
INFO[0026] System Create:                               
INFO[0027] ✅ Already Exists: Namespace "openshift-storage" 
INFO[0035] ✅ Created: NooBaa "noobaa"                   
```

```
$ oc create -f ./ibm-cloud-cos-creds.yaml 
secret/ibm-cloud-cos-creds created
```

```
$ oc get BackingStore
NAME                           TYPE      PHASE   AGE
noobaa-default-backing-store   ibm-cos   Ready   115s
```

```
$ oc create -f - <<EOF
apiVersion: noobaa.io/v1alpha1
kind: BucketClass
metadata:
  name: ibmcos-bucket-class
  labels:
    app: noobaa
spec:
  placementPolicy:
    tiers:
    - backingStores:
      - noobaa-default-backing-store
EOF
$ oc get bucketclass
NAME                          PLACEMENT                                                        NAMESPACEPOLICY   PHASE   AGE
ibmcos-bucket-class           map[tiers:[map[backingStores:[noobaa-default-backing-store]]]]                     Ready   36s
noobaa-default-bucket-class   map[tiers:[map[backingStores:[noobaa-default-backing-store]]]]                     Ready   14m
```
```
$ oc create -f - <<EOF 
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: ibmcos-noobaa.noobaa.io
provisioner: openshift-storage.noobaa.io/obc
parameters:
  bucketclass: ibmcos-bucket-class
reclaimPolicy: Delete
EOF

$ oc get sc ibmcos-noobaa.noobaa.io
NAME                      PROVISIONER                       RECLAIMPOLICY   VOLUMEBINDINGMODE   ALLOWVOLUMEEXPANSION   AGE
ibmcos-noobaa.noobaa.io   openshift-storage.noobaa.io/obc   Delete          Immediate           false                  15s
```
```
$ oc create -f - <<EOF
apiVersion: objectbucket.io/v1alpha1
kind: ObjectBucketClaim
metadata:
  name: ibmcos21
spec:
  storageClassName: ibmcos-noobaa.noobaa.io
  bucketName: ibmcos01.test
  additionalConfig: {}
EOF

$ oc get ob
NAME                  STORAGE-CLASS             CLAIM-NAMESPACE   CLAIM-NAME   RECLAIM-POLICY   PHASE   AGE
obc-ibmcos-ibmcos21   ibmcos-noobaa.noobaa.io                                  Delete           Bound   22s
```

```
$ oc get configmap,secrets
NAME                 DATA   AGE
configmap/ibmcos21   5      7m23s
NAME                              TYPE                                  DATA   AGE
secret/ibmcos21                   Opaque                                2      7m25s

$ oc create -f ./pod-devtest.yaml

$ oc -n ibmcos exec -it pod-devtest-01 -- sh
/ # pip install awscli-plugin-endpoint;
/ # aws configure set plugins.endpoint awscli_plugin_endpoint;
/ # aws configure set s3.endpoint_url https://s3.openshift-storage.svc
/ # ls /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt
/ # aws configure set ca_bundle /var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt;
/ # aws s3 ls s3://
2020-12-09 02:52:58 ibmcos01.test
/ # aws s3 ls s3://ibmcos01.test/
/ # aws s3 cp --recursive /etc s3://ibmcos01.test/
/ # aws s3 ls s3://ibmcos01.test/
                           PRE apk/
                           PRE ca-certificates/
                           PRE crontabs/
```

jeniawhite (approved):

nkkashyap:
Manual steps to create `ibm-cloud-cos-creds`
```
Here are the steps. 
1. Login to IBM Cloud
$ ibmcloud login --sso -a https://cloud.ibm.com -r us-south -g default -u <IBMid>

2. Create a IBM COS Service Instance
$ ibmcloud resource service-instance-create noobaa-stor cloud-object-storage standard global

3. Create access keys
$ ibmcloud resource service-key-create cos-cred-rw Writer --instance-name noobaa-stor --parameters '{ "HMAC": true}'

4. Create K8S secret under “openshift-storage” namespace
$ oc -n 'openshift-storage' create secret generic 'ibm-cloud-cos-creds' --type=Opaque --from-literal=IBM_COS_ACCESS_KEY_ID=<access_key_id> --from-literal=IBM_COS_SECRET_ACCESS_KEY=<secret_access_key>
```

PR #448, opened 2020-10-28
Refactor Ceph default backingstore code against latest rook changes

nb-ohad:
- Update rook dependency from v1.1.2 to v1.4.6
- Remove dependency on an `rgw-endpoint` label on the noobaa CR
- Remove dependency on reading the endpoint from the CephObjectStoreUser
- Read the cephObjectStoreUser secret from cephObjectStoreUser status
- Add proper "waiting" for both CephObjectStoreUser and the secret pointed by it
- Fix camel casing mistake with the cephObjectStoreUser variable


pkg/system/phase2_creating.go
```diff
@@ -352,43 +352,41 @@ func (r *Reconciler) ReconcileBackingStoreCredentials() error {
...
-			r.Logger.Warn("did not find an rgw-endpoint label on the noobaa CR")
-			return nil
-		}
+	// Try to list the ceph object stores.
+	cephObjectStoreList := &cephv1.CephObjectStoreList{}
+	if !util.KubeList(cephObjectStoreList, &client.ListOptions{Namespace: options.Namespace}) {
+		r.Logger.Info("failed to list ceph objectstore to use as backing store")
+		return nil
+	}
+	if len(cephObjectStoreList.Items) > 0 {
```
romayalon, line 373: I think you mean equal to 0 and not larger than 0.

nb-ohad: Thanks, fixed it

romayalon (approved):

dannyzaken (approved):

PR #451, opened 2020-11-04
Created diagnostic tar file directly to the specified dir

romayalon:

Changes: 
1. Created the diagnostics tar file directly to specified dir.

Problem: 
using the `nb diagnose --dir` didn't work as expected inside the must-gather pod.
Renaming the tar file to the wanted directory didn't work inside of the must-gather pod and raised an error that implies that the source directory and the target directory are not on the same file systems.

### DIFF
--- a/pkg/diagnose/diagnose.go
+++ b/pkg/diagnose/diagnose.go
@@ -144,6 +144,15 @@ func (c *Collector) CollectPodLogs(corePodSelector labels.Selector) {
 // ExportDiagnostics info
 func (c *Collector) ExportDiagnostics(destDir string) {
 	targetFile := fmt.Sprintf("%s.tar.gz", c.folderName)
+	if destDir != "" {
+		if _, err := os.Stat(destDir); os.IsNotExist(err) {
+			err := os.MkdirAll(destDir, os.ModePerm)
+			if err != nil {
+				c.log.Fatalf(`❌ Could not create directory %s, reason: %s`, destDir, err)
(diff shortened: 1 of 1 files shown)


nimrod-becker (approved):

jeniawhite (approved):

pkg/diagnose/diagnose.go
```diff
@@ -144,6 +144,15 @@ func (c *Collector) CollectPodLogs(corePodSelector labels.Selector) {
...
 func (c *Collector) ExportDiagnostics(destDir string) {
 	targetFile := fmt.Sprintf("%s.tar.gz", c.folderName)
+	if destDir != "" {
+		if _, err := os.Stat(destDir); os.IsNotExist(err) {
+			err := os.MkdirAll(destDir, os.ModePerm)
+			if err != nil {
+				c.log.Fatalf(`❌ Could not create directory %s, reason: %s`, destDir, err)
+			}
+		}
+		targetFile = fmt.Sprintf("%s/%s", destDir, targetFile)
```
jeniawhite, line 154: Make sure that writing the TAR from one file system to the second one works.

jeniawhite (approved):

pkg/diagnose/diagnose.go
romayalon: Pulkit is checking it on the same scenario, but there is no call to os.rename in the Tar function.

PR #453, opened 2020-11-05
Add secret for postgres creds

jackyalbo:
1. Add secret for postgres creds: user/password
2. connect to both db sts and core sts so they can communicate
3. changed postgres user to noobaa - as postgres user insisted in creating the default db - postgres and not nbcore
4. fixed issues with labels and selectors - to enable db service to really work

### DIFF
--- a/pkg/system/reconciler.go
+++ b/pkg/system/reconciler.go
@@ -71,6 +71,7 @@ type Reconciler struct {
 	ServiceS3           *corev1.Service
 	ServiceDb           *corev1.Service
 	SecretServer        *corev1.Secret
+	SecretDB            *corev1.Secret
 	SecretOp            *corev1.Secret
 	SecretAdmin         *corev1.Secret
 	SecretEndpoints     *corev1.Secret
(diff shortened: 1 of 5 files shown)


pkg/system/reconciler.go
```diff
@@ -223,6 +227,9 @@ func NewReconciler(
 	r.SecretServer.StringData["jwt"] = util.RandomBase64(16)
 	r.SecretServer.StringData["server_secret"] = util.RandomHex(4)
 
+	r.SecretDB.StringData["user"] = "noobaa"
+	r.SecretDB.StringData["password"] = util.RandomBase64(10)
```
dannyzaken, line 231: if you generate the password here won't it change the password in the secret when NewReconciler is called?

jackyalbo: @dannyzaken I don't think it suppose to change anything - like in jwt and server_secret - correct me if Im wrong here...

jackyalbo: I mean it will change in memory but not the secret itself

dannyzaken: you're right. the `kubeCheck` will load the existing secret if present

dannyzaken (approved):

PR #457, opened 2020-11-17
Moving mongo url into a secret

liranmauda:
### Changes:
- Removed the value from the statefulset-core
- Added to the sercretDb a url key
- Getting the mongo db url from a secret

### Not In Scope:
- We are not exposing the db secret in the crd yet, That we will do in another PR 

### DIFF
--- a/pkg/system/reconciler.go
+++ b/pkg/system/reconciler.go
@@ -229,6 +229,7 @@ func NewReconciler(
 
 	r.SecretDB.StringData["user"] = "noobaa"
 	r.SecretDB.StringData["password"] = util.RandomBase64(10)
+	r.SecretDB.StringData["url"] = "mongodb://" + r.NooBaaMongoDB.Name + "-0." + r.NooBaaMongoDB.Spec.ServiceName + "/nbcore"
 
 	// set noobaa root master key secret
 	r.SecretRootMasterKey.StringData["cipher_key_b64"] = util.RandomBase64(32)
@@ -242,8 +243,8 @@ func (r *Reconciler) CheckAll() {
 	util.KubeCheck(r.CoreApp)
 	util.KubeCheck(r.ServiceMgmt)
 	util.KubeCheck(r.ServiceS3)
+	util.KubeCheck(r.SecretDB)
 	if r.NooBaa.Spec.DBType == "postgres" {
-		util.KubeCheck(r.SecretDB)
 		util.KubeCheck(r.NooBaaPostgresDB)
 	} else {
 		util.KubeCheck(r.NooBaaMongoDB)
--- a/pkg/system/reconciler.go
+++ b/pkg/system/reconciler.go
@@ -71,7 +71,8 @@ type Reconciler struct {
 	ServiceS3           *corev1.Service
 	ServiceDb           *corev1.Service
 	SecretServer        *corev1.Secret
-	SecretDB            *corev1.Secret
+	SecretMongoDB       *corev1.Secret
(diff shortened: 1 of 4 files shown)


liranmauda:
@dannyzaken 
We can divide the secret DB into `Postgres secret DB` and `mongo secret DB`.
Currently, when we have only secret DB not divided, we populate it with data for Postgres and for mongo. 
What do you think?

dannyzaken:
@liranmauda maybe we should have separate secrets since during upgrade and migration we should deploy both DBs

jackyalbo:
I don't really like this differences between mongo and postgres. If we decide that url is inside a secret for some reason than it suppose to be for both. I also don't think it suppose to be the same secre: a secret that we get from the user and a secret we are creating for saving our own created creds. Also not sure we are suppose to create the secret for him. he should create a secret and send it to us in some way.

dannyzaken:
@jackyalbo I agree that the secret should contain all connection details, but there will always be differences. 
this PR is only handling the internal DB secret that is passed to noobaa-core. the external secret that should be created by the user will be an optional secret to connect to an external DB

jackyalbo:
@dannyzaken I talked with @liranmauda and know I understand better that we want an internal secret without connection to the external one, but I do think we need to put all the connection details in this secret, maybe also the db name - to make it as close to mongo's url as possible. what do you think?

guymguym:
@dannyzaken maybe we can check how the mongodb & postgres operators set their credentials and other access information into kubernetes secrets? Maybe aligning to their way is preferred?

dannyzaken:
@guymguym I think this is more relevant to the external secret passed by the user, isn't it?
in any case, for mongo at least, I think this is the relevant page: https://github.com/mongodb/mongodb-kubernetes-operator/blob/master/docs/users.md - it looks like they only store the password in a secret and the user name is passed in the MongoDB CR. they also have a different way to pass TLS certificates.
as for Postgres - there isn't any formal operator and I guess that each operator has its own thing. didn't check it really

liranmauda:
For Mongo operator, We might need also `publicApiKey` and `user` 
* [mongodb.com/kubernetes-operator/](https://docs.mongodb.com/kubernetes-operator/master/tutorial/create-operator-credentials/index.html)

But I think for starters we can use only a URL as it can contain credentials in the form of `mongodb://user:password@mongoserver.com/...`

* https://github.com/noobaa/noobaa-operator/issues/304#issuecomment-683690960

pkg/system/reconciler.go
```diff
@@ -189,7 +192,8 @@ func NewReconciler(
 	r.ServiceS3.Name = "s3"
 	r.ServiceDb.Name = r.Request.Name + "-db"
 	r.SecretServer.Name = r.Request.Name + "-server"
-	r.SecretDB.Name = r.Request.Name + "-db"
+	r.SecretMongoDB.Name = r.Request.Name + "-db"
+	r.SecretPostgresDB.Name = r.Request.Name + "-db"
```
liranmauda, line 196: Need to change the names to differentiate between the mongo secret to the postgres

liranmauda:
Moving to draft.
After talking with @lallinger-arbeit and @QuintinBecker we might want to drop this effort

lallinger-arbeit:
@liranmauda you should not move this to draft, we currently still plan to use it. Also when postgresql is available with noobaa we then would like to provide our own postgresql instance and specify our credentials via a secret..

nimrod-becker:
Closing this stale, PR.
As we discussed we also want to start moving away from MongoDB in upstream as well

PR #464, opened 2020-11-24
Namespace Bucketclass and NamespaceStore CRDS

romayalon:


deploy/crds/noobaa.io_bucketclasses_crd.yaml
```diff
@@ -8,6 +8,10 @@ spec:
     description: Placement
     name: Placement
     type: string
+  - JSONPath: .spec.namespacePolicy
```
dannyzaken, line 11: rebase and regenerate the CRDs. it should generate it as `apiextensions.k8s.io/v1` now

```
make gen-api
```

pkg/bucketclass/bucketclass.go
```diff
@@ -39,14 +41,96 @@ func Cmd() *cobra.Command {
...
+		CmdCreateNamespaceBucketclass(),
+		CmdCreateRegularBucketClass(),
+	)
+	return cmd
+}
+
+// CmdCreateRegularBucketClass returns a CLI command
+func CmdCreateRegularBucketClass() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "regular <bucket-class-name>",
```
dannyzaken, line 58: not sure `regular` is a good name. 
maybe `placement-bucketclass` to be consisten with the CRD?

jeniawhite (approved):

dannyzaken (commented):
some minor comments so far. still reviewing

dannyzaken (approved):

pkg/bucketclass/bucketclass.go
guymguym: @romayalon @dannyzaken 

So are we now defining "types" of bucket classes that the CLI creates?
I think the usage currently is a bit weird:
```
noobaa bucketclass create regular ...
noobaa bucketclass create namespace-bucketclass ...
```

For now, I think would suggest to keep a generic create command for bucket classes, and then define the policies to apply using flags such as --placementPolicy ... and --namespacePolicy ...
Actually getting to different "types" of bucketclasses might lead us to think about use cases more than policies, and then it will be like `noobaa bucketclass create hub-cache "my-cached-cloud-hub" --hub-connection ...` which will eventually set both placement policy and namespace policy.

romayalon: @guymguym I tried this command but it won't work because it's not one of the available commands (placement-bucketclass/namespace-buckeclass) actually I think that if I won't have this protection we might get into a situation of having both of the policies... 
I can change the naming to be namespace/placement policy if you feel it's better phrased.. WDYT?

deploy/crds/noobaa.io_bucketclasses_crd.yaml
romayalon: fixed

pkg/apis/noobaa/v1alpha1/namespacestore_types.go
```diff
@@ -0,0 +1,137 @@
...
+	// Items is the list of NamespaceStores.
+	Items []NamespaceStore `json:"items"`
+}
+
+// NamespaceStoreSpec defines the desired state of NamespaceStore
+// +k8s:openapi-gen=true
+type NamespaceStoreSpec struct {
+
+	// Type is an enum of supported types
+	Type StoreType `json:"type"`
```
guymguym, line 61: separate to NSType?

pkg/bucketclass/bucketclass.go
```diff
@@ -39,14 +41,96 @@ func Cmd() *cobra.Command {
...
+		CmdCreateNamespaceBucketclass(),
+		CmdCreatePlacementBucketClass(),
+	)
+	return cmd
+}
+
+// CmdCreatePlacementBucketClass returns a CLI command
+func CmdCreatePlacementBucketClass() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "placement-bucketclass <bucket-class-name>",
```
guymguym, line 58: For caching bucket classes we need both policies, so how to do it from CLI?

```
kubectl create -f bucketclass.yaml
noobaa backingstore create  --placementPolicy  --namespacePolicy
```

romayalon: right, added support for both policied for cache.

PR #465, opened 2020-11-25
Adding missing Env variables for core sts

jackyalbo:
Added a new check on missing env variable and adding them if they appear in the yaml

Doing this only for core sts for now just for you guys to take a look - we can copy the logic to other yamls if we need to - for now tagged as WIP until we figure this out

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -293,6 +293,8 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 				coreImageChanged = true
 				c.Image = r.NooBaa.Status.ActualImage
 			}
+			// adding the missing Env varibale from default container
+			util.MergeEnvArrays(&c.Env, &r.DefaultCoreApp.Env);
 			for j := range c.Env {
 				switch c.Env[j].Name {
 				case "AGENT_PROFILE":
@@ -341,7 +343,6 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 						}
 					}
 				}
-						
 			}
 
 			util.ReflectEnvVariable(&c.Env, "HTTP_PROXY")
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -1074,6 +1074,21 @@ func ReflectEnvVariable(env *[]corev1.EnvVar, name string) {
 	}
 }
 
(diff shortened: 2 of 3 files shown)


pkg/system/phase2_creating.go
```diff
@@ -341,7 +343,14 @@ func (r *Reconciler) SetDesiredCoreApp() error {
...
 				}
-						
+			}
+
+			// adding missing Env varibale from default container
+			for j := range r.DefaultCoreApp.Env {
+				if (!existingEnvs[r.DefaultCoreApp.Env[j].Name] && 
+					(r.DefaultCoreApp.Env[j].Value != "" || r.DefaultCoreApp.Env[j].ValueFrom != nil)) {
+					c.Env = append(c.Env, r.DefaultCoreApp.Env[j])
+				}
```
nb-ohad, line 353: I don't think that copy is enough as sometimes we leave envs values empty on purpose so the code will reconcile them with custom logic.  If I will just delete n env that has an empty value (in the template)  it will not get its reconciled value back.

I think a better approach is to take the list from the template then merge into it the current envs then run the reconcile code (the for with the switch statement) on the merged list

nb-ohad (changes requested):

pkg/system/phase2_creating.go
dannyzaken: @nb-ohad is right. just add a `mergeEnvs` function and call it before the envs `for`

nb-ohad: Just to clarify, you should merge only the missing envs from the  template  into the CR, then the reconcile env loop can stay the same

jackyalbo:
@dannyzaken @nb-ohad fixed to merge - please re-review

pkg/util/util.go
```diff
@@ -1074,6 +1074,21 @@ func ReflectEnvVariable(env *[]corev1.EnvVar, name string) {
...
+	for _, item := range *envA {
+		existingEnvs[item.Name] = true
+	}
+
+	for _, item := range *envB {
+			if !existingEnvs[item.Name] {
+				*envA = append(*envA, item)
+			}
+	}
+}
```
nb-ohad, line 1090: I think something with the indentation here is wrong

nb-ohad (approved):

dannyzaken (approved):

PR #468, opened 2020-11-25
[Github action] Add weekly build and push 

nimrod-becker:


.github/workflows/weekly-build.yml
```diff
@@ -0,0 +1,40 @@
...
+
+      - name: Login to DockerHub Registry
+        run: echo ${{ secrets.GHACTIONSDOCKERHUB }} | docker login -u ${{ secrets.GHACTIONSDOCKERHUBNAME }} --password-stdin
+
+      - name: Build & Push Docker Image
+        env:
+          DOCKERHUB_OWNER: ${{ secrets.GHACTIONSDOCKERHUBNAME }}
+        run: |
+            make image VERSION=latest
+            docker tag  ${{ steps.prep.outputs.operatortags }}
```
dannyzaken, line 37: you can skip the tag here by passing the version from prep instead of using latest:
```
make image VERSION={{ steps.prep.outputs.version }}
```

dannyzaken (approved):

.github/workflows/weekly-build.yml
nimrod-becker: but the tag includes the "master-" and version does not, I can maybe use the tags and skip tagging the image later

PR #483, opened 2020-12-08
Add disable aggregator field to NooBaa CR and export it as env variable

romayalon:


dannyzaken (approved):

guymguym:
Why add this to spec?? Is this a valid configuration? If this is just a workaround please keep it off the CR spec.

dannyzaken:
I agree with @guymguym, my hand was easy on the approve. we should control this setting in a different way and not change the CRD.

nb-ohad:
Adding to what Guy has said, a simple env var will do the trick

dannyzaken:
@nb-ohad is right. in that case we shouldn't change anything in the operator. setting the env on the core sts is enough

romayalon:
that's correct, it can be done by a single env var. thanks :) 
closing

PR #493, opened 2020-12-14
Upgrade namespace CRDs

romayalon:

### DIFF
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -16,6 +16,7 @@ import (
 
 	"github.com/marstr/randname"
 	nbv1 "github.com/noobaa/noobaa-operator/v2/pkg/apis/noobaa/v1alpha1"
+	"github.com/noobaa/noobaa-operator/v2/pkg/bundle"
 	"github.com/noobaa/noobaa-operator/v2/pkg/nb"
 	"github.com/noobaa/noobaa-operator/v2/pkg/options"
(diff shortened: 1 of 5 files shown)


pkg/system/phase4_configuring.go
```diff
@@ -1089,3 +1098,133 @@ func (r *Reconciler) ReconcileDeploymentEndpointStatus() error {
...
+
+			if !util.KubeCheck(secret) {
+				// Create secret
+				util.Panic(controllerutil.SetControllerReference(nsStore, secret, scheme.Scheme))
+				if !util.KubeCreateSkipExisting(secret) {
+					logrus.Errorf(`❌ Could not create Secret %q in Namespace %q (conflict)`, secret.Name, secret.Namespace)
+					continue
+				}
+			}
+			err = r.NBClient.SetNamespaceStoreInfo(nb.NamespaceStoreInfo{
```
dannyzaken, line 1220: not sure I understand the flow. once you create a NS store wouldn't its reconciler try to create a NS resource in noobaa? maybe I missed it, but I didn't see where it's handled

romayalon: no it shouldn't, we are talking about upgrade so it already exist as namespace resource before the upgrade.
when calling read system it already exist so it won't be called..

dannyzaken: 👍

dannyzaken (approved):

PR #521, opened 2021-01-20
Adding timeout to RPC calls

jackyalbo:
Fixing BZ1915261

it seems we didn't timeout RPC calls which led to a hanging requests which hold the whole reconcile flow. limiting now to 2 minutes like we do in the server.

### DIFF
--- a/pkg/nb/rpc.go
+++ b/pkg/nb/rpc.go
@@ -4,6 +4,7 @@ import (
 	"net/http"
 	"strings"
 	"sync"
+	"time"
 
 	util "github.com/noobaa/noobaa-operator/v2/pkg/util"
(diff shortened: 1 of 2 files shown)


pkg/nb/rpc.go
```diff
@@ -126,6 +129,7 @@ func NewRPC() *RPC {
 	return &RPC{
 		HTTPClient: http.Client{
 			Transport: util.InsecureHTTPTransport,
+			Timeout: RPCSendTimeout,
```
jeniawhite, line 132: We can start with 2 minutes and see how it goes.
This timeout covers all of the flow: connection, getting the response and reading its body.

jeniawhite (approved):

PR #526, opened 2021-01-24
noobaa cli enhancement - Create backingstore and namespacestore by specifying existing secret name

romayalon:

### DIFF
--- a/pkg/backingstore/backingstore.go
+++ b/pkg/backingstore/backingstore.go
@@ -82,6 +82,10 @@ func CmdCreateAWSS3() *cobra.Command {
 		`Secret key for authentication - the best practice is to **omit this flag**, in that case the CLI will prompt to prompt and read it securely from the terminal to avoid leaking secrets in the shell history`,
 	)
 	cmd.Flags().String(
+		"secret-name", "",
+		`The name of a secret for authentication - should have AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY properties`,
+	)
+	cmd.Flags().String(
 		"region", "",
 		"The AWS bucket region",
 	)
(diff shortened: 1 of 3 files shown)


pkg/namespacestore/namespacestore.go
```diff
@@ -265,11 +282,20 @@ func createCommon(cmd *cobra.Command, args []string, storeType nbv1.NSType, popu
 func RunCreateAWSS3(cmd *cobra.Command, args []string) {
 	createCommon(cmd, args, nbv1.NSStoreTypeAWSS3, func(namespaceStore *nbv1.NamespaceStore, secret *corev1.Secret) {
 		targetBucket := util.GetFlagStringOrPrompt(cmd, "target-bucket")
-		accessKey := util.GetFlagStringOrPromptPassword(cmd, "access-key")
-		secretKey := util.GetFlagStringOrPromptPassword(cmd, "secret-key")
 		region, _ := cmd.Flags().GetString("region")
-		secret.StringData["AWS_ACCESS_KEY_ID"] = accessKey
-		secret.StringData["AWS_SECRET_ACCESS_KEY"] = secretKey
+		secretName, _ := cmd.Flags().GetString("secret-name")
```
jeniawhite, line 286: Looks like this whole code can be in a function
The only difference is the keys, not the logic

pkg/backingstore/backingstore.go
```diff
@@ -323,11 +344,19 @@ func createCommon(cmd *cobra.Command, args []string, storeType nbv1.StoreType, p
 func RunCreateAWSS3(cmd *cobra.Command, args []string) {
 	createCommon(cmd, args, nbv1.StoreTypeAWSS3, func(backStore *nbv1.BackingStore, secret *corev1.Secret) {
 		targetBucket := util.GetFlagStringOrPrompt(cmd, "target-bucket")
-		accessKey := util.GetFlagStringOrPromptPassword(cmd, "access-key")
-		secretKey := util.GetFlagStringOrPromptPassword(cmd, "secret-key")
 		region, _ := cmd.Flags().GetString("region")
-		secret.StringData["AWS_ACCESS_KEY_ID"] = accessKey
-		secret.StringData["AWS_SECRET_ACCESS_KEY"] = secretKey
+		secretName, _ := cmd.Flags().GetString("secret-name")
```
jeniawhite, line 348: Same thing as namespacestore

jeniawhite (approved):

pkg/namespacestore/namespacestore.go
romayalon: let's keep it as it is so the flags will be handled at the same place and won't need to pass 4 keys for each call.

pkg/backingstore/backingstore.go
romayalon: same

dannyzaken (approved):

PR #527, opened 2021-01-25
CLI | Wait for System Phase is "Ready" on status

liranmauda:
### Explain the changes:

When running status and the system is not ready, we will not print an empty "System Status"
We will wait for "System Phase is Ready",  and then we will then print the System Status.

### Testing Instructions:
Have a working noobaa and do:
1. Run `kubectl edit noobaa noobaa`
2. Edit the image and replace it with a new core image 
3. Wait for the system to go down, then run `noobaa status` using the CLI.

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -109,6 +109,11 @@ func RunStatus(cmd *cobra.Command, args []string) {
 	log.Printf("Operator Status:")
 	operator.RunStatus(cmd, args)
 	log.Printf("")
+	log.Printf("System Wait Ready:")
+	if system.WaitReady() {
+		log.Printf("")
+		log.Printf("")
+	}
 	log.Printf("System Status:")
 	system.RunStatus(cmd, args)
 


pkg/install/install.go
```diff
@@ -109,6 +109,11 @@ func RunStatus(cmd *cobra.Command, args []string) {
 	log.Printf("Operator Status:")
 	operator.RunStatus(cmd, args)
 	log.Printf("")
+	log.Printf("System Wait Ready:")
+	if system.WaitReady() {
```
dannyzaken, line 113: just nitpicking here, but what happens when there is no noobaa CR or CRD installed?

liranmauda: It Woll get into an endless loop of:
```
⏳ Failed to get system: no matches for kind "NooBaa" in version "noobaa.io/v1alpha1"
```
But that is no different from a situation where it hangs on connecting or configuring, which can happen on install also. 
We can add to the `WaitReady` timeout if we want.

dannyzaken (approved):

PR #528, opened 2021-01-27
Add owner filter predicate

romayalon:

### DIFF
--- a/pkg/controller/noobaa/noobaa_controller.go
+++ b/pkg/controller/noobaa/noobaa_controller.go
@@ -57,6 +57,12 @@ func Add(mgr manager.Manager) error {
 	// Predicate that allow us to log event that are being queued
 	logEventsPredicate := util.LogEventsPredicate{}
 
+	// Predicate that filter events that noobaa is not their owner
+	filterForOwnerPredicate := util.FilterForOwner{
+		OwnerType: &nbv1.NooBaa{},
+		Scheme:    mgr.GetScheme(),
+	}
+
 	// Predicate that allows events that only change spec, labels or finalizers will log any allowed events
 	// This will stop infinite reconciles that triggered by status or irrelevant metadata changes
 	noobaaPredicate := util.ComposePredicates(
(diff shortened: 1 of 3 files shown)


pkg/util/predicates.go
```diff
@@ -71,6 +73,50 @@ func (p FinalizersChangedPredicate) Update(e event.UpdateEvent) bool {
...
+func hasCorrectOwner(arr []v1.OwnerReference, ownerSubstrings []string) bool {
+	for _, b := range arr {
+		for _, sub := range ownerSubstrings {
+			if strings.Contains(b.Name, sub) {
+				return true
+			}
+		}
+	}
+	return false
+}
```
nb-ohad, line 91: Wouldn't it be easier to have it as a `FilterForOwner` method? you will have access to the `OwnerSubstrings` via object `r.OwnerSubstrings` and will not pollute the package scope

pkg/util/predicates.go
```diff
@@ -71,6 +73,50 @@ func (p FinalizersChangedPredicate) Update(e event.UpdateEvent) bool {
...
+func (p FilterForOwner) Delete(e event.DeleteEvent) bool {
+	eventOwners := e.Meta.GetOwnerReferences()
+	return hasCorrectOwner(eventOwners, p.OwnerSubstrings)
+}
+
+// Update implements the update event trap for FilterForOwner
+func (p FilterForOwner) Update(e event.UpdateEvent) bool {
+	newEventOwners := e.MetaNew.GetOwnerReferences()
+	oldEventOwners := e.MetaOld.GetOwnerReferences()
+	return hasCorrectOwner(newEventOwners, p.OwnerSubstrings) || hasCorrectOwner(oldEventOwners, p.OwnerSubstrings)
```
nb-ohad, line 110: Why do you care about the old owner in this case? Why does it matter?

pkg/util/predicates.go
romayalon: thanks, fixed.

pkg/util/predicates.go
romayalon: right, they will be filtered out anyway by the ownerHandler, deleted.

pkg/controller/noobaa/noobaa_controller.go
```diff
@@ -57,6 +57,15 @@ func Add(mgr manager.Manager) error {
...
 	logEventsPredicate := util.LogEventsPredicate{}
 
+	// Predicate that filter events that noobaa is not their owner
+	filterForOwnerPredicate := util.FilterForOwner{
+		OwnerSubstrings: []string{"noobaa"},
+	}
+	// Predicate that filter pod events that one of noobaa statefulset/deployments is not their owner
+	podfilterForOwnerPredicate := util.FilterForOwner{
+		OwnerSubstrings: []string{"noobaa-operator", "noobaa-core", "noobaa-db", "noobaa-endpoint"},
+	}
```
nb-ohad, line 67: Is there a benefit of using a shared instance of the predicate, it seems that each resource type has a diff Owner and I think that it will be more precise to use an inline instantiation of the predicate with the proper owner on each watch

pkg/util/predicates.go
```diff
@@ -71,6 +73,49 @@ func (p FinalizersChangedPredicate) Update(e event.UpdateEvent) bool {
 		!reflect.DeepEqual(e.MetaOld.GetFinalizers(), e.MetaNew.GetFinalizers())
 }
 
+// FilterForOwner will only allow events that owned by noobaa
+type FilterForOwner struct {
+	OwnerSubstrings []string
```
nb-ohad, line 78: Why do you need an array where the list can implicitly be created by stacking more than one predicate on the watch? In most cases, you just use a single string. 
As an added bonus, moving to a single string will make the code simpler

nb-ohad (changes requested):

pkg/controller/noobaa/noobaa_controller.go
romayalon: deleted it according to our conversation so it's irelevant.

nb-ohad (approved):

PR #534, opened 2021-02-04
Added govcloud regions to valid AWSRegions

edcrosbys:
AWS govcloud regions were missing from mapValidAWSRegions variable. Added us-gov-east-1 and us-gov-west-1.


edcrosbys:
fixes #535

PR #549, opened 2021-02-16
Fixing stuck in pending OBC NS

jeniawhite:
Fixed stuck in pending OBC NS that couldn't create buckets in DS on internal.
Applied default bucket class to OBC NS with the creation of the tiering policy for the OBC NS.
TODO: Need to see what happens on OBC NS Cache because cannot configure both tiering and namespace in the bucket class, currently using default as well.
Fixes https://bugzilla.redhat.com/show_bug.cgi?id=1927330

### DIFF
--- a/pkg/obc/provisioner.go
+++ b/pkg/obc/provisioner.go
@@ -99,7 +99,7 @@ func (p *Provisioner) Provision(bucketOptions *obAPI.BucketOptions) (*nbv1.Objec
 	}
 	// TODO: we need to better handle the case that a bucket was created, but Provision failed
 	// right now we will fail on create bucket when Provision is called the second time
-	err = r.CreateBucket()
+	err = r.CreateBucket(p, bucketOptions)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/obc/provisioner.go
```diff
@@ -381,6 +372,72 @@ func (r *BucketRequest) CreateBucket() error {
...
+			Namespace: p.Namespace,
+		},
+	}
+
+	if !util.KubeCheck(bucketClass) {
+		msg := fmt.Sprintf("GetDefaultBucketClass BucketClass %q not found in provisioner namespace %q", bucketClassName, p.Namespace)
+		return nil, fmt.Errorf(msg)
+	}
+
+	if bucketClass.Status.Phase != nbv1.BucketClassPhaseReady {
```
romayalon, line 399: what's the reason for this ready bucketclass check? I don't think we enforce this on other creations of OBCs.

romayalon (approved):

PR #552, opened 2021-02-21
cli diagnose to collect endpoint logs

jackyalbo:


pkg/diagnose/diagnose.go
```diff
@@ -70,6 +70,9 @@ func RunCollect(cmd *cobra.Command, args []string) {
 	operatorPodSelector, _ := labels.Parse("noobaa-operator=deployment")
 	c.CollectPodLogs(operatorPodSelector)
 
+	endpointPodSelector, _ := labels.Parse("noobaa-s3=" + options.SystemName)
```
nimrod-becker, line 73: will it catch all endpoints (auto-scale...)

jackyalbo: checked a system with 2 endpoints - both had this label

nimrod-becker (approved):

PR #576, opened 2021-03-10
Creating Noobaa SCC when installing using cli

jackyalbo:
Will create Noobaa SCC if not found and SCC CRD exist
Added option to uninstall SCC by running: ```noobaa uninstall --cleanup ```

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -1,6 +1,7 @@
 package operator
 
 import (
+	"fmt"
 	"os"
 	"strings"
 
@@ -8,6 +9,7 @@ import (
 	"github.com/noobaa/noobaa-operator/v2/pkg/options"
 	"github.com/noobaa/noobaa-operator/v2/pkg/util"
 
+	secv1 "github.com/openshift/api/security/v1"
 	"github.com/spf13/cobra"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -94,6 +96,7 @@ func RunInstall(cmd *cobra.Command, args []string) {
 	util.KubeCreateSkipExisting(c.RoleBinding)
 	util.KubeCreateSkipExisting(c.ClusterRole)
 	util.KubeCreateSkipExisting(c.ClusterRoleBinding)
+	util.KubeCreateOptional(c.SecurityContextConstraints)
 	noDeploy, _ := cmd.Flags().GetBool("no-deploy")
 	if !noDeploy {
 		util.KubeCreateSkipExisting(c.Deployment)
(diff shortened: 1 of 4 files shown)


deploy/security_context_constraints.yaml
```diff
@@ -0,0 +1,36 @@
+apiVersion: security.openshift.io/v1
```
guymguym, line 1: I would rename the file to scc.yaml - I wouldn't expect to find it with this full name and might miss it

pkg/install/install.go
```diff
@@ -34,6 +34,7 @@ func CmdUninstall() *cobra.Command {
 	}
 	cmd.Flags().Bool("cleanup", false, "Enable deletion of Namespace and CRD's")
 	cmd.Flags().Bool("cleanup_data", false, "Clean object buckets")
+	cmd.Flags().Bool("cleanscc", false, "Enable deletion of the NooBaa SCC")
```
guymguym, line 37: why not include in cleanup? it cleans other clusterwide entities

pkg/operator/operator.go
```diff
@@ -210,6 +221,10 @@ func LoadOperatorConf(cmd *cobra.Command) *Conf {
 		c.Deployment.Spec.Template.Spec.ImagePullSecrets =
 			[]corev1.LocalObjectReference{{Name: options.ImagePullSecret}}
 	}
+
+	c.SecurityContextConstraints.Name = c.SA.Name
+	c.SecurityContextConstraints.Namespace = options.Namespace
```
guymguym, line 226: namespace? for SCC? seems irrelevant and might cause errors.

pkg/operator/operator.go
```diff
@@ -210,6 +221,10 @@ func LoadOperatorConf(cmd *cobra.Command) *Conf {
 		c.Deployment.Spec.Template.Spec.ImagePullSecrets =
 			[]corev1.LocalObjectReference{{Name: options.ImagePullSecret}}
 	}
+
+	c.SecurityContextConstraints.Name = c.SA.Name
```
guymguym, line 225: why do we need to set the name dynamically and not in the yaml?

pkg/operator/operator.go
```diff
@@ -94,6 +96,7 @@ func RunInstall(cmd *cobra.Command, args []string) {
 	util.KubeCreateSkipExisting(c.RoleBinding)
 	util.KubeCreateSkipExisting(c.ClusterRole)
 	util.KubeCreateSkipExisting(c.ClusterRoleBinding)
+	util.KubeCreateSkipExisting(c.SecurityContextConstraints)
```
guymguym, line 99: Actually is this where you wanted to add it at all? This only runs when calling `noobaa install` and is not running during reconcile at all. This is fine if we just want to support CLI install of the SCC, but this means that we still do not handle it instead of OCS operator.

guymguym: Also this will fail on non openshift clusters, where the SCC CRD is not available.
See how we create other openshift CR's like Routes.

pkg/install/install.go
jackyalbo: IDK just for a shared cluster like gogo - I accidentally deleted the SCC there

pkg/operator/operator.go
jackyalbo: We want to install during cli install and that's it for now - reconciles will still be only in OCS operator
Tried in minikube and it worked fine. Just warned that the SCC is misssing: 
```INFO[0002] ❌ CRD Missing: SecurityContextConstraints "noobaa"```

pkg/operator/operator.go
jackyalbo: right, I can remove both

pkg/operator/operator.go
jackyalbo: moved to cleanup

dannyzaken (approved):

pkg/operator/operator.go
guymguym: this looks like an error to me. that's why we used the ReconcileObjectOptional -
https://github.com/noobaa/noobaa-operator/blob/47283a1777a0a42567de44a608771de7981ca0aa/pkg/system/reconciler.go#L522
and the CheckOptional -
https://github.com/noobaa/noobaa-operator/blob/47283a1777a0a42567de44a608771de7981ca0aa/pkg/util/util.go#L366
to avoid these things looking like errors to the user.

jackyalbo: Added KubeCreateOptional

PR #578, opened 2021-03-11
delete the noobaa-db pod before starting migration process

dannyzaken:

Fixes BZ https://bugzilla.redhat.com/show_bug.cgi?id=1922113

* when we upgrade ocs-operator to 4.7, we apply changes to noobaa SCC from ocs-operator. these changes cause noobaa-db pod to restart
* The SCC changes require some new code to run in noobaa init container, but when noobaa-db-0 starts it is still using the old noobaa-core image as the init container.
* Since the new code is missing, the init container is getting into a crash loop
* Once OCS operator updates the noobaa CR with the new noobaa core image, noobaa-operator updates the noobaa-db STS. For some reason, when the init container is in a crash loop, this update does not cause the pod to restart with the new image (not sure if it's a Statefulset bug or as designed. opened an [issue](https://github.com/kubernetes/kubernetes/issues/100094))
* To fix the issue - deleting the noobaa-db pod when starting the upgrade and migration process. this flow is only relevant when dbtype=postgres and mongo STS exists 
* Deleting the mongo STS when the migration is done

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -20,6 +20,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
@@ -845,25 +846,35 @@ func (r *Reconciler) UpgradeMigrateDB() error {
 	}
 
 	r.Logger.Infof("UpgradeMigrateDB: upgrade phase - %s", phase)
-	oldSts := &appsv1.StatefulSet{
+	mongoSts := &appsv1.StatefulSet{
 		TypeMeta: metav1.TypeMeta{Kind: "StatefulSet"},
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "noobaa-db",
 			Namespace: options.Namespace,
 		},
 	}
-	if !util.KubeCheck(oldSts) {
-		r.Logger.Info("Old (mongo) STS was not found. skipping migration")
-		phase = nbv1.UpgradePhaseNone
-	} else {
-		replicas := int32(1)
-		if phase == nbv1.UpgradePhaseClean {
-			r.Logger.Info("UpgradeMigrateDB:: Cleanning phase - setting old STS to 0 replicas")
-			replicas = int32(0)
+
+	mongoExists := util.KubeCheck(mongoSts)
+
+	switch phase {
+
+	case "":
+		if mongoExists {
+			r.Logger.Infof("UpgradeMigrateDB: setting phase to %s", nbv1.UpgradePhasePrepare)
+			phase = nbv1.UpgradePhasePrepare
+		} else {
+			// no mongo STS. skip migration
+			r.Logger.Info("Old (mongo) STS was not found. skipping migration")
+			phase = nbv1.UpgradePhaseNone
 		}
-		if err := r.ReconcileObject(oldSts, func() error { // remove old sts pods when finish migrating
-			oldSts.Spec.Replicas = &replicas
-			podSpec := &oldSts.Spec.Template.Spec
+
+	case nbv1.UpgradePhasePrepare:
+		r.Logger.Infof("UpgradeMigrateDB:: prepare phase")
+
+		// update mongo sts with the new noobaa-core image as the init container image
+		r.Logger.Infof("UpgradeMigrateDB:: updating mongo STS init container")
+		if err := r.ReconcileObject(mongoSts, func() error { // remove old sts pods when finish migrating
+			podSpec := &mongoSts.Spec.Template.Spec
 			podSpec.ServiceAccountName = "noobaa"
 			for i := range podSpec.InitContainers {
 				c := &podSpec.InitContainers[i]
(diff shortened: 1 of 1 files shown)


pkg/system/phase2_creating.go
```diff
@@ -889,6 +890,20 @@ func (r *Reconciler) UpgradeMigrateDB() error {
 			return err
 		}
 		if phase == "" {
+			// when starting - restart the db pod. This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1922113
+			r.Logger.Info("attempt to delete noobaa-db-0")
```
liranmauda, line 894: Isn't it getting the noobaa-db-0 and not deleting?

dannyzaken: I get the object first before deleting it

liranmauda: Yes, it is getting the object but the printing can mislead.
```suggestion
			r.Logger.Info("attempt to delete noobaa-db-0, getting the pod")
```

liranmauda (approved):

pkg/system/phase2_creating.go
```diff
@@ -889,6 +890,20 @@ func (r *Reconciler) UpgradeMigrateDB() error {
 			return err
 		}
 		if phase == "" {
+			// when starting - restart the db pod. This is a fix for https://bugzilla.redhat.com/show_bug.cgi?id=1922113
```
jackyalbo, line 893: Why not inside of the Prepare it's part of that phase  no? I don't think we should do it here

pkg/system/phase2_creating.go
```diff
@@ -889,6 +890,20 @@ func (r *Reconciler) UpgradeMigrateDB() error {
...
+			r.Logger.Info("attempt to delete noobaa-db-0")
+			dbPod := &corev1.Pod{}
+			err := r.Client.Get(r.Ctx, types.NamespacedName{Namespace: options.Namespace, Name: "noobaa-db-0"}, dbPod)
+			if err != nil {
+				r.Logger.Errorf("got error when trying to get noobaa-db-0 pod - %v", err)
+				return err
+			}
+			r.Logger.Info("deleting noobaa-db-0 pod")
+			err = r.Client.Delete(r.Ctx, dbPod)
```
jackyalbo, line 902: Maybe check for the image first before the deleting to avoid unneeded  wait

jackyalbo (changes requested):

pkg/system/phase2_creating.go
```diff
@@ -845,97 +846,139 @@ func (r *Reconciler) UpgradeMigrateDB() error {
...
-		}); err != nil {
-			r.Logger.Errorf("got error on endpoints deployment reconcile %v", err)
-			return err
-		}
-		if phase == "" {
+
+	mongoExists := util.KubeCheck(mongoSts)
+
+	// if mongo STS exists (or we are in the middle of cleanup phase) then perform migration from mongo to postgres
+	if mongoExists || phase == nbv1.UpgradePhaseClean {
```
jackyalbo, line 860: addition due to deleting  right? Maybe this check is unneeded? like check in "" if we have mongo set to UpgradePhasePrepare and if we don't set to  UpgradePhaseNone. what do you think?

pkg/system/phase2_creating.go
```diff
@@ -845,97 +846,139 @@ func (r *Reconciler) UpgradeMigrateDB() error {
...
+			} else {
+				r.Logger.Infof("UpgradeMigrateDB:: migration not finished yet")
+				return fmt.Errorf("job didn't finish yet")
+			}
+
+		case nbv1.UpgradePhaseClean:
+			r.Logger.Infof("UpgradeMigrateDB:: cleanup phase")
+
+			r.Logger.Infof("UpgradeMigrateDB:: deleting mongodb STS")
+			err := r.Client.Delete(r.Ctx, mongoSts)
```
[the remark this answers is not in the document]
dannyzaken: no. the PVC is not deleted

pkg/system/phase2_creating.go
```diff
@@ -950,3 +993,12 @@ func (r *Reconciler) SetDesiredJobUpgradeDB() error {
 	r.setDesiredCoreEnv(&r.UpgradeJob.Spec.Template.Spec.Containers[0])
 	return nil
 }
+
+// SetEndpointsDeploymentReplicas updates the number of replicas on the endpoints deployment
+func (r *Reconciler) SetEndpointsDeploymentReplicas(replicas int32) error {
+	r.Logger.Infof("UpgradeMigrateDB:: setting endpoints replica count to %d", replicas)
+	return r.ReconcileObject(r.DeploymentEndpoint, func() error { 
+		r.DeploymentEndpoint.Spec.Replicas = &replicas
```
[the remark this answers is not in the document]
dannyzaken: yes. it disables the HPA

pkg/system/phase2_creating.go
```diff
@@ -845,97 +846,139 @@ func (r *Reconciler) UpgradeMigrateDB() error {
...
-				if len(oldDbPodList.Items) == 0 {
-					phase = nbv1.UpgradePhaseFinished
-					r.Logger.Infof("UpgradeMigrateDB:: mongo pod terminated. setting phase to %s", nbv1.UpgradePhaseFinished)
-				} else {
-					r.Logger.Infof("UpgradeMigrateDB:: mongo pod is still running. waiting for termination")
-					return fmt.Errorf("not all old pods are cleaned yet")
-				}
 			}
+
+			// remove endpoints and core pods. set replicas to 0
```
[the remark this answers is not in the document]
dannyzaken: you're right. I thought to move it here but didn't want to shake the existing code too much. 
I'll fix the comment

pkg/system/phase2_creating.go
dannyzaken: in the cleaning phase, we still want to enter the switch after the STS is deleted

jackyalbo: Yeah I understood that, so wouldn't it be cleaner if we move it inside?

jackyalbo (approved):

pkg/system/phase2_creating.go
dannyzaken: ok. got you. I'll change it

PR #579, opened 2021-03-11
NSFS namespacestore implementation

romayalon:


pkg/namespacestore/namespacestore.go
```diff
@@ -174,6 +175,24 @@ func CmdCreateAzureBlob() *cobra.Command {
...
 
+// CmdCreateNSFS returns a CLI command
+func CmdCreateNSFS() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "nsfs <namespace-store-name>",
+		Short: "Create nsfs namespace store",
+		Run:   RunCreateNSFS,
+	}
+	cmd.Flags().String(
+		"fs-backend", "",
```
jeniawhite, line 186: I assume this is only in the initial PR?
Because I think that there should be strict types for that, does it make sense?

pkg/namespacestore/namespacestore.go
```diff
@@ -394,6 +413,23 @@ func RunCreateAzureBlob(cmd *cobra.Command, args []string) {
 	})
 }
 
+// RunCreateNSFS runs a CLI command
+func RunCreateNSFS(cmd *cobra.Command, args []string) {
+	log := util.Logger()
+	createCommon(cmd, args, nbv1.NSStoreTypeNSFS, func(namespaceStore *nbv1.NamespaceStore, secret *corev1.Secret) {
+		fsPath := util.GetFlagStringOrPrompt(cmd, "fs-path")
+		fsBackend, _ := cmd.Flags().GetString("fs-backend")
```
jeniawhite, line 421: Will need to add checks for type of `fsBackend`

jeniawhite (approved):

pkg/namespacestore/namespacestore.go
romayalon: added an enum validation to the namespace store types, thank you.

pkg/namespacestore/namespacestore.go
romayalon: same, added enum validation instead of the cli.

PR #584, opened 2021-03-17
setting upgrade job restartPolicy to never

dannyzaken:
* setting `restartPolicy` to never will keep failed pods around so we can examine the failures
* setting `backoffLimit` to 4 (default is 6) to reduce the number of pods that can potentially stay in the NS
* one thing to notice is that failed runs will leave a pod (not running) with ERROR state. This can lead to support cases, but maybe these are cases we want to know about

related to BZ https://bugzilla.redhat.com/show_bug.cgi?id=1938557

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -25,6 +25,8 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const upgradeJobBackoffLimit = int32(4)
+
 // ReconcilePhaseCreating runs the reconcile phase
 func (r *Reconciler) ReconcilePhaseCreating() error {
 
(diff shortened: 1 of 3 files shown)


nimrod-becker (approved):

nimrod-becker (approved):

pkg/system/phase2_creating.go
```diff
@@ -987,8 +998,13 @@ func (r *Reconciler) UpgradeMigrateDB() error {
...
 // SetDesiredJobUpgradeDB updates the UpgradeJob as desired for reconciling
 func (r *Reconciler) SetDesiredJobUpgradeDB() error {
+	backoffLimit := upgradeJobBackoffLimit
 	r.UpgradeJob.Spec.Template.Spec.Containers[0].Image = r.NooBaa.Status.ActualImage
 	r.UpgradeJob.Spec.Template.Spec.Containers[0].Command = []string{"/noobaa_init_files/noobaa_init.sh", "db_migrate"}
+	// setting the restart policy to never to keep the pods around after failed migrations
+	// also reducing the backoff limit to avoid to many pods staying around in case of an issue
+	r.UpgradeJob.Spec.Template.Spec.RestartPolicy = corev1.RestartPolicyNever
+	r.UpgradeJob.Spec.BackoffLimit = &backoffLimit
 	r.setDesiredCoreEnv(&r.UpgradeJob.Spec.Template.Spec.Containers[0])
```
jackyalbo, line 1008: move this before the added code (line 1004) - so the two parts will be better divided

jackyalbo (changes requested):

jackyalbo (approved):

PR #592, opened 2021-03-29
fix for deleting namespacestore/backingstore reconciliation

romayalon:

1. keep reconciling deleting backingstores/namespacestore when their phase is rejected. fixes BZ #1940957.


dannyzaken (commented):
we are already have `NamespaceStorePhaseDeleting` and `BackingStorePhaseDeleting`. I think that we should stay in this phase and keep checking for the conditions instead of putting the backingstore\namespacestore into a rejected phase.

so I think when in deleting phase we should keep returning temporary error as long as the conditions are not met

romayalon:
@dannyzaken 
oh ok, I thought we must keep the rejected phase. 
changed the code to return temp error and now we will keep reconciling. 
thank you

PR #593, opened 2021-03-30
moved core STS reconciliation to run after migration

dannyzaken:

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1938557

* noobaa-core is reconciled before we start the migration process
* when changing DB type to postgres, the operator restart noobaa-core and postgres before the migration
* in some cases, noobaa-core starts before the migration starts and it creates a support account (here: https://github.com/noobaa/noobaa-core/blob/c40153e5e4389834ab597fef94ce60615cbeb18a/src/server/web_server.js#L66)
* changed the sequence to reconcile core after migration phase

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -94,9 +94,7 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 	if err := r.UpgradeSplitDB(); err != nil {
 		return err
 	}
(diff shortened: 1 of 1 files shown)


dannyzaken:
@jackyalbo It seems safe to change the order. can you check that I am not missing anything here?

jackyalbo (approved):
Just not sure if we need the core before we are creating the service. I would check. other than that LGTM

PR #598, opened 2021-04-04
specify mongo db image when reconciling old mongo sts

dannyzaken:

fixes https://bugzilla.redhat.com/show_bug.cgi?id=1945450

* when reconciling mongo STS in migrate flow we no longer have the mong image stored in the spec (dbImage is used for postgres image).
* explicitly set the mongo image when reconciling mongo STS since in some cases the STS could have been reconciled with the posters image.

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -887,6 +887,19 @@ func (r *Reconciler) UpgradeMigrateDB() error {
 					c.Image = r.NooBaa.Status.ActualImage
 				}
 			}
+			for i := range podSpec.Containers {
+				c := &podSpec.Containers[i]
+				if c.Name == "db" {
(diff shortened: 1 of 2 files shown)


pkg/system/phase2_creating.go
```diff
@@ -887,6 +887,14 @@ func (r *Reconciler) UpgradeMigrateDB() error {
 					c.Image = r.NooBaa.Status.ActualImage
 				}
 			}
+			for i := range podSpec.Containers {
+				c := &podSpec.Containers[i]
+				switch c.Name {
```
jackyalbo, line 892: Do we really need a switch here? Can't we do in simple if like in init-containers?
if c.Name == "db" {
	c.Image = options.DBMongoImage
}

jackyalbo (approved):

PR #611, opened 2021-04-21
Add initial manual build workflow

nimrod-becker:
Signed-off-by: Nimrod Becker <bnimrod@gmail.com>

### DIFF
--- a/.github/workflows/manual-build.yaml
+++ b/.github/workflows/manual-build.yaml
@@ -1,0 +1,24 @@
+name: Manual Build Dispatch
+on: 
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch to Build From'
+        required: true
+      tag:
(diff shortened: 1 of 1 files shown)


.github/workflows/manual-build.yaml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.event.inputs.branch }}
+
+      - name: Get Current Date
+        id: date
+        run: echo "::set-output name=date::$(date +'%Y%m%d')"
+	run: echo "hello world done"
```
liranmauda, line 24: Is this line intentional?

nimrod-becker: yes, its a hello world I need to verify working and then I can add actual logic

PR #618, opened 2021-04-22
changed noobaa-endpoint security context

dannyzaken:
changes for OCS-4.8. revert some of the changes done for NSFS
- changed service account to noobaa
- removed the requested securityContext

fixes bz https://bugzilla.redhat.com/show_bug.cgi?id=1952344

### DIFF
--- a/deploy/internal/deployment-endpoint.yaml
+++ b/deploy/internal/deployment-endpoint.yaml
@@ -20,7 +20,7 @@ spec:
         noobaa-s3: noobaa
         app: noobaa
     spec:
-      serviceAccountName: noobaa-endpoint
+      serviceAccountName: noobaa
       volumes:
         - name: mgmt-secret
           secret:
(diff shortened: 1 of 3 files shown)


dannyzaken:
@jackyalbo this should remove the need for noobaa-endpoint SCC, right?

jackyalbo:
> @jackyalbo this should remove the need for noobaa-endpoint SCC, right?

right. But did we backport also the whole cli additions? I guess we don't have to remove them. But it will create the noobaa-endpoint scc, the service account and all the rest - which nothing will use.

jackyalbo (approved):

dannyzaken:
> > @jackyalbo this should remove the need for noobaa-endpoint SCC, right?
> 
> right. But did we backport also the whole cli additions? I guess we don't have to remove them. But it will create the noobaa-endpoint scc, the service account and all the rest when using noobaa cli install - which nothing will use.

I think that's OK. most installations are not done from CLI

PR #626, opened 2021-05-03
reconciling mongo service if it exists regardless of DBType

dannyzaken:
- in some cases mongo service can be updated with wrong values when DBType is changed to postgres and the running operator version is 5.6
- fixes BZ https://bugzilla.redhat.com/show_bug.cgi?id=1955328

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -103,22 +103,34 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 		if err := r.ReconcileDB(); err != nil {
 			return err
 		}
-		if err := r.ReconcileObject(r.ServiceDb, r.SetDesiredServiceDB); err != nil {
-			return err
+
+		if r.NooBaa.Spec.DBType == "postgres" {
+			if err := r.ReconcileObject(r.ServiceDbPg, r.SetDesiredServiceDBForPostgres); err != nil {
+				return err
+			}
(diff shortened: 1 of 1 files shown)


pkg/system/phase2_creating.go
```diff
@@ -103,22 +103,34 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
+			if err := r.ReconcileObject(r.ServiceDbPg, r.SetDesiredServiceDBForPostgres); err != nil {
+				return err
+			}
+			// fix for https://bugzilla.redhat.com/show_bug.cgi?id=1955328
+			// if DBType=postgres was passed in version 5.6 (OCS 4.6) the operator reconciled
+			// the mongo service with postgres values. see here:
+			// https://github.com/noobaa/noobaa-operator/blob/112c510650612b1a6b88582cf41c53b30068161c/pkg/system/phase2_creating.go#L121-L126
+			// to fix that, reconcile mongo service as well if it exists
+			if util.KubeCheck(r.ServiceDb) {
+				r.Logger.Infof("DZDZ - found existing mongo db service [%q] will reconcile", r.ServiceDb.Name)
```
liranmauda, line 117: DZDZ :)

liranmauda (approved):

pkg/system/phase2_creating.go
dannyzaken: thanks. fixed

jackyalbo (approved):

Issue #303, opened 2020-05-11
CLI `noobaa install yaml` - command to dump the install yamls


guymguym:
Pretty simple to implement and this will be useful for integrating with other yaml based tools that can install noobaa - such as helm.

PR #642, opened 2021-05-26
CLI noobaa install yaml - command to dump the install yamls #303

baum:
### Testing Instructions:
Run the following on clean k8s cluster:
- `noobaa --mini=true install yaml  1> install.yaml`
- `kubectl create -f install.yaml`


liranmauda (approved):
LGTM,
Just one mark, we need to remember to remove it on the DS

nimrod-becker:
@liranmauda true

dannyzaken (approved):

dannyzaken:
@liranmauda why do we need to remove it DS?

liranmauda:
> @liranmauda why do we need to remove it DS?

@dannyzaken  The DS CLI does not support `install` and `uninstall`.
As we hide the `install` and `uninstall` CLI command we would like to avoid dumping the yamls (which will ease the installation of noobaa using `kubectl create -f install.yaml`)

With that said, as this command only dumps the yamls, do you see value in keeping it in the DS CLI (for analyzing maybe)?

@dannyzaken @guymguym @baum WDYT?

nimrod-becker:
We can keep it, I don't see the harm in it

Issue #484, opened 2020-12-09
NSFS: generic bug: noobaa install not reporting erroneous input for options 


bssrikanth:
This issue is tagged with NSFS but bug reported here is generic.

currently noobaa install does not report any errors if a user wrongly inputs options with  – vs. --. Ex: When user wrongly inputs –pv-pool-default-storage-class='<storage class>' instead of --pv-pool-default-storage-class='<storage class>', the install proceeds without reporting any errors and completes. But after the install `noobaa-default-backing-store` pod remains in Pending state showing there is no storage provision-er detected for pvc. 

If user has used wrong syntax unknowingly then there is lot of time spent on seeing why the pvc was not able to create before tracing it back to the incorrect syntax used.

```
# noobaa install --image-pull-secret='myregistrykey' --db-storage-class='<storage class>' –pv-pool-default-storage-class='<storage class>'
INFO[0000] CLI version: 5.5.0-nsfs                      
INFO[0000] noobaa-image: noobaa/noobaa-core:5.5.0-nsfs  
INFO[0000] operator-image: noobaa/noobaa-operator:5.5.0-nsfs 
INFO[0000] Namespace: noobaa                            
INFO[0000]                                              
INFO[0000] CRD Create:                                  
INFO[0000] ✅ Created: CustomResourceDefinition "noobaas.noobaa.io" 
INFO[0000] ✅ Created: CustomResourceDefinition "backingstores.noobaa.io" 
INFO[0000] ✅ Created: CustomResourceDefinition "bucketclasses.noobaa.io" 
INFO[0000] ✅ Created: CustomResourceDefinition "objectbucketclaims.objectbucket.io" 
INFO[0000] ✅ Created: CustomResourceDefinition "objectbuckets.objectbucket.io" 
INFO[0000]                                              
INFO[0000] Operator Install:                            
INFO[0000] ✅ Already Exists: Namespace "noobaa"         
INFO[0000] ✅ Created: ServiceAccount "noobaa"           
INFO[0000] ✅ Created: Role "noobaa"                     
INFO[0000] ✅ Created: RoleBinding "noobaa"              
INFO[0000] ✅ Created: ClusterRole "noobaa.noobaa.io"    
INFO[0000] ✅ Created: ClusterRoleBinding "noobaa.noobaa.io" 
INFO[0000] ✅ Created: Deployment "noobaa-operator"      
INFO[0000]                                              
INFO[0000] System Create:                               
[...]

PR #646, opened 2021-06-03
Generate an error if a user wrongly inputs options with –

baum:
Issue #484
=========

Execute pre-run check and generate an error if there are elements in the command flag args list, 
which indicate wrong inputs options. The general rule is that all noobaa commands would generate
an error on non-flag arguments due to the PersistentPreRunE implementation at the root Cmd.
Children of the root command will inherit and execute PersistentPreRunE.

Defined a shared Pass function at options, i.e. this implementation never returns
an error for the commands that require non-flag argument, such as for instance
-> noobaa bucket create <bucket-name> [flags] [options]...
Those commands should define their PersistentPreRunE as Pass function.

Tests:
====

Rejecting non-flag arguments
------------------------------

```
➜  noobaa install --image-pull-secret='myregistrykey' --db-storage-class='<storage class>' –pv-pool-default-storage-class='<storage class>' 
Error: unknown args: [–pv-pool-default-storage-class=<storage class>]
➜  noobaa install yaml  –pv-pool-default-storage-class='<storage class>'
Error: unknown args: [–pv-pool-default-storage-class=<storage class>]
```

Accepting non-flag arguments
------------------------------
```
➜  noobaa bucket create                  
FATA[0000] Missing expected arguments: <bucket-name> 

Usage:
  noobaa bucket create <bucket-name> [flags] [options]

➜  noobaa bucket create second.bucket
INFO[0000] ✅ Exists: NooBaa "noobaa"                    
INFO[0000] ✅ Exists: Service "noobaa-mgmt"
...
```

### DIFF
--- a/pkg/backingstore/backingstore.go
+++ b/pkg/backingstore/backingstore.go
@@ -48,8 +48,9 @@ func Cmd() *cobra.Command {
 // CmdCreate returns a CLI command
 func CmdCreate() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "create",
-		Short: "Create backing store",
+		Use:               "create",
+		Short:             "Create backing store",
+		PersistentPreRunE: options.PersistentPreRunEPass,
 	}
 	cmd.AddCommand(
 		CmdCreateAWSS3(),
@@ -229,9 +230,10 @@ func CmdCreatePVPool() *cobra.Command {
 // CmdDelete returns a CLI command
 func CmdDelete() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "delete <backing-store-name>",
-		Short: "Delete backing store",
-		Run:   RunDelete,
+		Use:               "delete <backing-store-name>",
+		Short:             "Delete backing store",
+		Run:               RunDelete,
+		PersistentPreRunE: options.PersistentPreRunEPass,
 	}
 	return cmd
 }
@@ -239,9 +241,10 @@ func CmdDelete() *cobra.Command {
 // CmdStatus returns a CLI command
 func CmdStatus() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "status <backing-store-name>",
-		Short: "Status backing store",
-		Run:   RunStatus,
+		Use:               "status <backing-store-name>",
+		Short:             "Status backing store",
+		Run:               RunStatus,
+		PersistentPreRunE: options.PersistentPreRunEPass,
 	}
 	return cmd
 }
@@ -249,9 +252,10 @@ func CmdStatus() *cobra.Command {
 // CmdRunRemovePendingPods returns a CLI command
 func CmdRunRemovePendingPods() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "remove_pending <backing-store-name>",
-		Short: "Deletes all the pending pods that failed to connect to server",
-		Run:   RunRemovePendingPods,
+		Use:               "remove_pending <backing-store-name>",
+		Short:             "Deletes all the pending pods that failed to connect to server",
+		Run:               RunRemovePendingPods,
+		PersistentPreRunE: options.PersistentPreRunEPass,
 	}
 	return cmd
 }
@@ -269,10 +273,11 @@ func CmdList() *cobra.Command {
 // CmdReconcile returns a CLI command
 func CmdReconcile() *cobra.Command {
 	cmd := &cobra.Command{
-		Hidden: true,
-		Use:    "reconcile",
-		Short:  "Runs a reconcile attempt like noobaa-operator",
-		Run:    RunReconcile,
+		Hidden:            true,
+		Use:               "reconcile",
+		Short:             "Runs a reconcile attempt like noobaa-operator",
+		Run:               RunReconcile,
+		PersistentPreRunE: options.PersistentPreRunEPass,
 	}
 	return cmd
 }
@@ -819,8 +824,8 @@ func RunReconcile(cmd *cobra.Command, args []string) {
 	}
 	backingStoreName := args[0]
 	klient := util.KubeClient()
-	intervalSec := time.Duration(3)
-	util.Panic(wait.PollImmediateInfinite(intervalSec*time.Second, func() (bool, error) {
+	intervalDuration := time.Duration(3)
+	util.Panic(wait.PollImmediateInfinite(intervalDuration*time.Second, func() (bool, error) {
 		req := reconcile.Request{
 			NamespacedName: types.NamespacedName{
 				Namespace: options.Namespace,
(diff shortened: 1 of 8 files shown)


pkg/install/install.go
```diff
@@ -14,12 +15,24 @@ import (
 	"github.com/spf13/cobra"
 )
 
+// PreRunEInstall generate an error if a user wrongly inputs options with –
+// See https://github.com/noobaa/noobaa-operator/issues/484
+func PreRunEInstall(cmd *cobra.Command, args []string) error {
+	flagArgs := cmd.Flags().Args()
+	if len(flagArgs) > 0 {
```
dannyzaken, line 22: I'm less familiar with cobra, but isn't there a more generic solution for that problem? this is probably an issue that happens in other commands as well. for example this will not handle `noobaa install yaml –pv-pool-default-storage-class='<storage class>'`

baum: @dannyzaken initially tried to define the PreRunE() filter on the root command, however it breaks commands which do expect non-flag arguments, like bucket name in  `noobaa bucket create <bucket-name> [flags] [options]`

it is also possible to define the PreRunE() fitter on the root command and then exclude the commands which expect the arguments.

WDYT?

baum: @dannyzaken , Let's another way. Defined that all noobaa commands should generate an error on non-flag arguments at the root Cmd, children of this command will inherit and execute [PersistentPreRunE](https://github.com/spf13/cobra/blob/4590150168e93f4b017c6e33469e26590ba839df/command.go#L107-L108):

https://github.com/baum/noobaa-operator/blob/bb9102de598eb83f45ca0eb91f58665c2b53ea83/pkg/cli/cli.go#L85-L93

Defined a shared Pass function at options, i.e. this implementation never returns an error:

https://github.com/baum/noobaa-operator/blob/bb9102de598eb83f45ca0eb91f58665c2b53ea83/pkg/options/options.go#L116-L122

Found 23 commands which require non-flag arguments, those commands set their PersistentPreRunE to the passing function, see for instance 'bucket create':
https://github.com/baum/noobaa-operator/blob/bb9102de598eb83f45ca0eb91f58665c2b53ea83/pkg/bucket/bucket.go#L36

So now:
```
➜  noobaa install yaml  –pv-pool-default-storage-class='<storage class>'
Error: unknown args: [–pv-pool-default-storage-class=<storage class>]
```

On the other hand commands requiring non-flag args are happy:
```
➜  noobaa bucket create                  
FATA[0000] Missing expected arguments: <bucket-name> 

Usage:
  noobaa bucket create <bucket-name> [flags] [options]

➜  noobaa bucket create second.bucket
INFO[0000] ✅ Exists: NooBaa "noobaa"                    
INFO[0000] ✅ Exists: Service "noobaa-mgmt"              
INFO[0000] ✅ Exists: Secret "noobaa-operator"           
INFO[0000] ✅ Exists: Secret "noobaa-admin"              
....
```

How does it look to you?

dannyzaken: I think it's a good solution, but it means that anyone that is implementing a new command requiring non-flag args will need to be aware of this change, right?
we also need to make sure we didn't miss any of the commands that require non-flag args

baum:
@dannyzaken, you're correct, commands requiring non-flag arguments need to define the PersistentPreRunE method in such model. it was relatively easy to locate those commands since they reference the args list at the beginning of their Run method.  

cobra commands form a tree, and "persistent" pre-post hooks could be inherited. it would be nice if we could mark whole subtrees - that would make our lives easier. however, some tree leaves do not match their sub-tree behavior.  for instance, in the "noobaa bucket" sub-tree there are four subcommands: 
-  create      Create a NooBaa bucket
-  delete      Delete a NooBaa bucket
-  list        List NooBaa buckets
-  status      Show the status of a NooBaa bucket 

and most require a non-flag argument, however `list` does not.

so I struggle to find a better strategy other than defining a default behavior, ie. allow or deny and then mark those commands that do not match.  i do not have a preference either way.

WDYT?

pkg/install/install.go
baum: @dannyzaken, you're correct, commands requiring non-flag arguments need to define the PersistentPreRunE method in such a model, however, the implementation is shared. it was relatively easy to locate those commands requiring non-flag arguments since they reference the args list at the beginning of their Run method.  

cobra commands form a tree, and "persistent" pre-post hooks are inherited. it would be nice if we could mark whole subtrees - that would make our lives easier. however, some tree leaves do not match their sub-tree behavior.  for instance, in the "noobaa bucket" sub-tree there are four subcommands: 
-  create      Create a NooBaa bucket
-  delete      Delete a NooBaa bucket
-  list        List NooBaa buckets
-  status      Show the status of a NooBaa bucket 

and most require a non-flag argument, however `list` does not.

so I struggle to find a better strategy other than defining a default behavior, ie. allow or deny and then mark those commands that do not match.  i do not have a preference either way.

WDYT?

dannyzaken (approved):

guymguym:
If we really want args validation - why not use the cobra NoArgs() option for any command that should not have extra args?
There are more options here that we can use to define args validations - https://github.com/spf13/cobra#positional-and-custom-arguments.

I don't like how the current solution uses PersistentPreRunE to specify the args validations and looks unclear to read.

pkg/backingstore/backingstore.go
```diff
@@ -819,8 +824,8 @@ func RunReconcile(cmd *cobra.Command, args []string) {
 	}
 	backingStoreName := args[0]
 	klient := util.KubeClient()
-	intervalSec := time.Duration(3)
-	util.Panic(wait.PollImmediateInfinite(intervalSec*time.Second, func() (bool, error) {
+	intervalDuration := time.Duration(3)
```
guymguym, line 827: Why change to a less descriptive name? It is now less clear that this duration represents seconds.

baum: Hey Guy. I've changed the variable name since I got a lint warning from VS code, that the duration variable name should not include the specific time unit name like seconds.  Duration is converted into specific time units, seconds in this case,  by multiplication by `time.Second` scale factor... Does it make sense?

guymguym: I see, I consider this a matter of style and not a lint warning, but whatever.
So perhaps it would require to separate like this:
```
intervalSec := 3
intervalDuration := time.Second * time.Duration(intervalSec)
```
and then you have both an int representing the number of seconds which can be easily printed for us humans,
and also the duration type for the wait call.

baum:
> If we really want args validation - why not use the cobra NoArgs() option for any command that should not have extra args?
> There are more options here that we can use to define args validations - https://github.com/spf13/cobra#positional-and-custom-arguments.
> 
> I don't like how the current solution uses PersistentPreRunE to specify the args validations and looks unclear to read.

@guymguym,  your suggestion looks elegant, we could also use  ExactArgs(1) instead of  code like https://github.com/baum/noobaa-operator/blob/14e5c0a9dbfc6d44d8f2674d7e862b53f01d2f32/pkg/bucket/bucket.go#L72-L74

WDYT?

baum:
@guymguym @dannyzaken 
@liranmauda  reported this implementation breaks `noobaa api account_api create_account`

Submitted a fix reverting this change and using cobra args validations - https://github.com/spf13/cobra#positional-and-custom-arguments

See PR #658

PR #653, opened 2021-06-14
Add the debug-level flag functionality as install option

liranmauda:
### Functionality
- Add NOOBAA_LOG_LEVEL as env variable in deployment-endpoint.yaml and statefulset-core.yaml and into noobaa CR
- Adding the debug-level flag to the options
- Add the debug-level flag functionality in the install command

### Gap
- Need to add a CLI command that changes the NOOBAA_LOG_LEVEL in noobaa CR for running clusters

### Testing
- Run `noobaa install --debug-level <number>` and see that the appropriate logs are being printed

### DIFF
--- a/pkg/apis/noobaa/v1alpha1/noobaa_types.go
+++ b/pkg/apis/noobaa/v1alpha1/noobaa_types.go
@@ -101,6 +101,10 @@ type NooBaaSpec struct {
 	// +optional
 	MongoDbURL string `json:"mongoDbURL,omitempty"`
 
+	// DebugLevel (optional) sets the debug level
+	// +optional
+	DebugLevel int `json:"debugLevel,omitempty"`
+
 	// PVPoolDefaultStorageClass (optional) overrides the default cluster StorageClass for the pv-pool volumes.
 	// This affects where the system stores data chunks (encrypted).
 	// Updates to this field will only affect new pv-pools,
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
@@ -341,6 +342,9 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 			} else {
 				c.Env[j].Value = "mongodb://" + r.NooBaaMongoDB.Name + "-0." + r.NooBaaMongoDB.Spec.ServiceName + "/nbcore"
 			}
+
+		case "NOOBAA_LOG_LEVEL":
+				c.Env[j].Value = strconv.Itoa(r.NooBaa.Spec.DebugLevel)
 
 		case "POSTGRES_HOST":
 			c.Env[j].Value = r.NooBaaPostgresDB.Name + "-0." + r.NooBaaPostgresDB.Spec.ServiceName
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -318,6 +318,8 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 					if r.JoinSecret == nil {
 						c.Env[j].Value = r.MongoConnectionString
 					}
+				case "NOOBAA_LOG_LEVEL":
+					c.Env[j].Value = strconv.Itoa(r.NooBaa.Spec.DebugLevel)
 				case "LOCAL_MD_SERVER":
 					if r.JoinSecret == nil {
 						c.Env[j].Value = "true"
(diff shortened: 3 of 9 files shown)


pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -101,6 +101,10 @@ type NooBaaSpec struct {
 	// +optional
 	MongoDbURL string `json:"mongoDbURL,omitempty"`
 
+	// DebugLevel (optional) sets the debug level
+	// +optional
+	DebugLevel string `json:"debugLevel,omitempty"`
```
guymguym, line 106: why string and not a number?

guymguym: The fact that env variables are string only doesn't mean that CRD fields should make their schema fields to be all strings. Instead I would say that env vars should be converted to and from strings when needed.

liranmauda: 👍

pkg/system/phase2_creating.go
```diff
@@ -342,6 +347,11 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 				c.Env[j].Value = "mongodb://" + r.NooBaaMongoDB.Name + "-0." + r.NooBaaMongoDB.Spec.ServiceName + "/nbcore"
 			}
 
+		case "NOOBAA_LOG_LEVEL":
+			if r.NooBaa.Spec.DebugLevel != 0 {
```
guymguym, line 351: what about the else case? if you never remove it then there is not way to disable this setting once it was done...
check out the other optional spec fields that have both a set and unset of an env.

pkg/system/phase4_configuring.go
```diff
@@ -318,6 +318,10 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 					if r.JoinSecret == nil {
 						c.Env[j].Value = r.MongoConnectionString
 					}
+				case "NOOBAA_LOG_LEVEL":
+					if r.JoinSecret == nil {
```
guymguym, line 322: JoinSecret?? copy-pasta...

pkg/system/phase4_configuring.go
```diff
@@ -318,6 +318,10 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
...
+						c.Env[j].Value = strconv.Itoa(r.DebugLevel)
```
guymguym, line 323: Again - also here we are missing the else case to unset the env if needed.

guymguym (changes requested):

pkg/system/phase2_creating.go
liranmauda: Removed the if completely as we have a default in the yaml (0)

pkg/system/phase4_configuring.go
liranmauda: Removed the if completely as we have a default in the yaml (0)

pkg/system/system.go
```diff
@@ -122,6 +122,8 @@ func LoadSystemDefaults() *nbv1.NooBaa {
 	sys.Finalizers = []string{nbv1.GracefulFinalizer}
 	dbType := options.DBType
 	sys.Spec.DBType = nbv1.DBTypes(dbType)
+	debugLevel := options.DebugLevel
+	sys.Spec.DebugLevel = debugLevel
```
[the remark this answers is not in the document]
liranmauda: True, Removed the later

pkg/system/phase4_configuring.go
```diff
@@ -318,6 +318,8 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 					if r.JoinSecret == nil {
 						c.Env[j].Value = r.MongoConnectionString
 					}
+				case "NOOBAA_LOG_LEVEL":
+					c.Env[j].Value = strconv.Itoa(r.NooBaa.Spec.DebugLevel)
```
guymguym, line 322: but what if the spec doesn't include DebugLevel?

guymguym: you can't rely on the CLI setting it because the NooBaa CR can be created by yaml...

guymguym: Oh but perhaps this is OK because the variable is int, and when empty it is omited, but still returned as 0 in golang when empty.

pkg/system/system.go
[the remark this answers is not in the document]
guymguym: OK, but why are you using this temp variable? Why not just set `sys.Spec.DebugLevel = options.DebugLevel` ?

guymguym (approved):

PR #661, opened 2021-06-16
Disable huge_pages in Postgres config

baum:
Issue:
  https://bugzilla.redhat.com/show_bug.cgi?id=1946792

Test:
  Verify in the postgress DB container:

   - existence and content of `/opt/app-root/src/postgresql-cfg/noobaa-postgres.conf` file:
```
bash-4.2$ pwd
/opt/app-root/src
bash-4.2$ cat postgresql-cfg/noobaa-postgres.conf
# disable huge_pages trial
# see https://bugzilla.redhat.com/show_bug.cgi?id=1946792#c18
huge_pages = off

```

- `openshift-custom-postgresql.conf` includes the
    `noobaa-postgres.conf` file

```
bash-4.2$ grep noobaa-postgres /var/lib/pgsql/openshift-custom-postgresql.conf
include '/opt/app-root/src/postgresql-cfg/noobaa-postgres.conf'
```

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
 	"github.com/noobaa/noobaa-operator/v5/pkg/options"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
 	cloudcredsv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
@@ -194,16 +195,20 @@ func (r *Reconciler) SetDesiredServiceDBForPostgres() error {
 // SetDesiredNooBaaDB updates the NooBaaDB as desired for reconciling
 func (r *Reconciler) SetDesiredNooBaaDB() error {
 	var NooBaaDB *appsv1.StatefulSet = nil
+	var NooBaaDBTemplate *appsv1.StatefulSet = nil
+
 	if r.NooBaa.Spec.DBType == "postgres" {
 		NooBaaDB = r.NooBaaPostgresDB
 		NooBaaDB.Spec.Template.Labels["noobaa-db"] = "postgres"
 		NooBaaDB.Spec.Selector.MatchLabels["noobaa-db"] = "postgres"
 		NooBaaDB.Spec.ServiceName = r.ServiceDbPg.Name
+		NooBaaDBTemplate = util.KubeObject(bundle.File_deploy_internal_statefulset_postgres_db_yaml).(*appsv1.StatefulSet)
 	} else {
 		NooBaaDB = r.NooBaaMongoDB
 		NooBaaDB.Spec.Template.Labels["noobaa-db"] = r.Request.Name
 		NooBaaDB.Spec.Selector.MatchLabels["noobaa-db"] = r.Request.Name
 		NooBaaDB.Spec.ServiceName = r.ServiceDb.Name
+		NooBaaDBTemplate = util.KubeObject(bundle.File_deploy_internal_statefulset_db_yaml).(*appsv1.StatefulSet)
 	}
 
 	podSpec := &NooBaaDB.Spec.Template.Spec
@@ -295,6 +300,12 @@ func (r *Reconciler) SetDesiredNooBaaDB() error {
 		}
 
 	} else {
+		// upgrade path add new resources,
+		// merge the volumes & volumeMounts from the template
+		util.MergeVolumeList( &NooBaaDB.Spec.Template.Spec.Volumes, &NooBaaDBTemplate.Spec.Template.Spec.Volumes)
+		for i := range NooBaaDB.Spec.Template.Spec.Containers {
+			util.MergeVolumeMountList(&NooBaaDB.Spec.Template.Spec.Containers[i].VolumeMounts, &NooBaaDBTemplate.Spec.Template.Spec.Containers[i].VolumeMounts)
+		}
 
 		// when already exists we check that there is no update requested to the volumes
 		// otherwise we report that volume update is unsupported
@@ -746,6 +757,14 @@ func (r *Reconciler) ReconcileRootSecret() error {
 func (r *Reconciler) ReconcileDB() error {
 	var err error
 	if r.NooBaa.Spec.DBType == "postgres" {
+		// this config map is required by the NooBaaPostgresDB StatefulSet,
+		// if the configMap was not created at this step, NooBaaPostgresDB
+		// would fail to start.
+		r.Own(r.PostgresDBConf)
+		if !util.KubeCreateSkipExisting(r.PostgresDBConf) {
+			return fmt.Errorf("could not create Postgres DB configMap %q in Namespace %q", r.PostgresDBConf.Name, r.PostgresDBConf.Namespace)
+		}
+
 		err = r.ReconcileObject(r.NooBaaPostgresDB, r.SetDesiredNooBaaDB)
 		// Making sure that previous CRs without the value will deploy MongoDB
 	} else if r.NooBaa.Spec.DBType == "" || r.NooBaa.Spec.DBType == "mongodb" {
--- a/pkg/system/reconciler.go
+++ b/pkg/system/reconciler.go
@@ -69,6 +69,7 @@ type Reconciler struct {
 	CoreApp                   *appsv1.StatefulSet
 	DefaultCoreApp            *corev1.Container
 	NooBaaMongoDB             *appsv1.StatefulSet
+	PostgresDBConf            *corev1.ConfigMap
 	NooBaaPostgresDB          *appsv1.StatefulSet
 	ServiceMgmt               *corev1.Service
 	ServiceS3                 *corev1.Service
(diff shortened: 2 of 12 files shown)


pkg/system/phase2_creating.go
```diff
@@ -746,6 +746,8 @@ func (r *Reconciler) ReconcileRootSecret() error {
 func (r *Reconciler) ReconcileDB() error {
 	var err error
 	if r.NooBaa.Spec.DBType == "postgres" {
+		r.PostgresDBConf.Namespace = r.NooBaaPostgresDB.Namespace
+		util.KubeCreateSkipExisting(r.PostgresDBConf)
```
guymguym, line 750: what about error checking? check out other examples using KubeCreateSkipExisting

baum: is it better?

https://github.com/baum/noobaa-operator/blob/60bf90429750c74a8e8f85568b607116c03edc01/pkg/system/phase2_creating.go#L751-L758

baum: also moved Postgres configMap Namespace assignment to the reconciler constructor:

https://github.com/noobaa/noobaa-operator/compare/0fcb889..3c3fe2f

liranmauda (approved):
LGTM

pkg/system/phase2_creating.go
```diff
@@ -746,6 +746,15 @@ func (r *Reconciler) ReconcileRootSecret() error {
 func (r *Reconciler) ReconcileDB() error {
 	var err error
 	if r.NooBaa.Spec.DBType == "postgres" {
+		util.KubeCreateSkipExisting(r.PostgresDBConf)
+
+		// this config map is required by the NooBaaPostgresDB StatefulSet,
+		// if the configMap was not created at this step, NooBaaPostgresDB
+		// would fail to start.
+		if !util.KubeCheck(r.PostgresDBConf) {
```
romayalon, line 754: I think that Guy meant that you will check errors of the actual `KubeCreateSkipExisting` output

pkg/system/reconciler.go
```diff
@@ -160,6 +162,7 @@ func NewReconciler(
 	r.ServiceAccount.Namespace = r.Request.Namespace
 	r.CoreApp.Namespace = r.Request.Namespace
 	r.NooBaaMongoDB.Namespace = r.Request.Namespace
+	r.PostgresDBConf.Namespace = r.Request.Namespace
```
romayalon, line 165: add it also to checkAll function in this file

pkg/system/phase2_creating.go
baum: @romayalon Existence of the configMap is required for the Postgres DB to start successfully. KubeCreateSkipExisting() returns false also in the case  "resource already exists", making examing this function return value inconclusive.  Makes more sense to test for the configMap existence. WDYT?

pkg/system/phase2_creating.go
guymguym: That would work but I think maybe it's worth to split util.KubeCreateSkipExisting into two a bit different functions:
- util.KubeCreateSkipExisting - almost like KubeCreateSkipExisting is today, only change is if already exists then return true (instead of false)
- util.KubeCreateFailExisting - this will do exactly what KubeCreateSkipExisting does today (!) yes the name is simply wrong because it returns false when already exists, just like with any other error.

This will make more sense for cases like yours where the we want to create it once and if it exists that's ok:
```golang
if !util.KubeCreateSkipExisting(r.PostgresDBConf) {
    return fmt.Errorf(...)
}
```

And for all the current cases that use KubeCreateSkipExisting we can choose which option fits better.
In most CLI calls it doesn't really matter because no body checked what the function returns.
But for cases that did check the return value, we want to see what was the poets intention...
WDYT?

pkg/system/phase2_creating.go
guymguym: you're right, the function name is misleading, see my comment above.

pkg/system/reconciler.go
guymguym: but should it be conditional there in checkAll depending on the dbtype?

romayalon: yes, there is existing condition there for dbtype

pkg/system/phase2_creating.go
romayalon: why not use KubeCreateOptional? the optional section is on a missing CRD

guymguym: because that's not the same case.
KubeCreateOptional is for CRDs that are optional in the cluster.
In these cases we want an error checking for the existence of the CR itself.

pkg/system/phase2_creating.go
baum: @romayalon, thank you! Sounds good.

See https://github.com/baum/noobaa-operator/blob/39c5a4ea3cad97d10aef872aa3c86ca68cbc6966/pkg/system/reconciler.go#L270

pkg/system/reconciler.go
baum: @romayalon, thank you! Sounds good.

See https://github.com/baum/noobaa-operator/blob/39c5a4ea3cad97d10aef872aa3c86ca68cbc6966/pkg/system/reconciler.go#L270

pkg/system/phase2_creating.go
baum: @guymguym sounds good. renamed symbol KubeCreateSkipExisting -> KubeCreateFailExisting and extracted kubeCreateSkipOrFailExisting() and implemented both skip and fail paths in terms of this new extracted function.

See https://github.com/baum/noobaa-operator/blob/39c5a4ea3cad97d10aef872aa3c86ca68cbc6966/pkg/util/util.go#L206-L263

Now the flow of new configMap creation looks solid 😀:

https://github.com/baum/noobaa-operator/blob/39c5a4ea3cad97d10aef872aa3c86ca68cbc6966/pkg/system/phase2_creating.go#L749-L755

pkg/operator/operator.go
```diff
@@ -90,20 +90,20 @@ func CmdRun() *cobra.Command {
...
-	util.KubeCreateSkipExisting(c.NS)
-	util.KubeCreateSkipExisting(c.SA)
-	util.KubeCreateSkipExisting(c.SAEndpoint)
-	util.KubeCreateSkipExisting(c.Role)
-	util.KubeCreateSkipExisting(c.RoleEndpoint)
-	util.KubeCreateSkipExisting(c.RoleBinding)
-	util.KubeCreateSkipExisting(c.RoleBindingEndpoint)
-	util.KubeCreateSkipExisting(c.ClusterRole)
-	util.KubeCreateSkipExisting(c.ClusterRoleBinding)
+	util.KubeCreateFailExisting(c.NS)
```
guymguym, line 93: I don't think these should change. keep it KubeCreateSkipExisting.

guymguym: I mean that for all the calls in the CLI that don't even check the return value.

pkg/system/reconciler.go
romayalon: oh I see it's missing also in the // Set Names section,
and maybe finding a better name other than huge-pages-config is preferred since it might be used for other configurations in the future

pkg/operator/operator.go
baum: @guymguym 👍 reverting unconditional invocations.
See https://github.com/noobaa/noobaa-operator/compare/39c5a4e..3ec69ef

pkg/system/phase2_creating.go
```diff
@@ -746,6 +746,13 @@ func (r *Reconciler) ReconcileRootSecret() error {
 func (r *Reconciler) ReconcileDB() error {
 	var err error
 	if r.NooBaa.Spec.DBType == "postgres" {
+		// this config map is required by the NooBaaPostgresDB StatefulSet,
+		// if the configMap was not created at this step, NooBaaPostgresDB
+		// would fail to start.
+		if !util.KubeCreateSkipExisting(r.PostgresDBConf) {
```
guymguym, line 752: Maybe I missed it, but did you call r.Own(r.PostgresDBConf) ?
You need this to make the configmap be deleted when the noobaa CR is deleted.

pkg/system/reconciler.go
baum: @romayalon,  👍 tweaked YAML level definitions, to get rid of huge-pages references on name level, looks tidier now, and allows future extensions. 

 https://github.com/noobaa/noobaa-operator/compare/3ec69ef..5d1098a

however set names change is probably not required, since the configMap name is static/constant and also referenced by the stateful state, as a volume.

WDYT?

liranmauda: @baum That was what I was thinking also, I saw that there is no .name but figured out it is not needed. 
@romayalon Are we missing something?

pkg/system/phase2_creating.go
baum: @guymguym 👍

https://github.com/noobaa/noobaa-operator/compare/5d1098a..42d9cbd

pkg/system/reconciler.go
romayalon: @baum @liranmauda it brings up the question of why even creating this Yaml file and not use an empty one that already exists and populate it before creation? is there any special reason for it?

baum: @romayalon @liranmauda

The configMap name is defined here:

https://github.com/baum/noobaa-operator/blob/42d9cbd07179a6afbb54e0c6d2714fc4f2532b8d/deploy/internal/configmap-postgres-db.yaml#L4

Referenced by the stateful set here:

https://github.com/baum/noobaa-operator/blob/42d9cbd07179a6afbb54e0c6d2714fc4f2532b8d/deploy/internal/statefulset-postgres-db.yaml#L70

No golang logic is required, all on YAML level.

Thank you for your review!

pkg/system/phase2_creating.go
guymguym: Thanks, you don't need to provide commit links on every change, you can just push and update the PR itself.

dannyzaken (commented):
@baum - to handle upgrades you'll need to add the configMap volume and mount in `SetDesiredNooBaaDB`

https://github.com/baum/noobaa-operator/blob/2a1e79af0a6938d6abfc2bcb0832e3fa7c008081/pkg/system/phase2_creating.go#L195-L195

baum:
@dannyzaken 10x, see the new volume, volume mounts merge logic on upgrade path

https://github.com/noobaa/noobaa-operator/compare/2a1e79a..2cbbc83

dannyzaken (approved):

PR #672, opened 2021-06-24
NooBaa Component Rescheduling

baum:
Improve NooBaa pods recovery in case of a node failure. Find noobaa pods
running on the failing node and force delete them to speed up
rescheduling in healthy node.

### DIFF
--- a/.github/workflows/run_hac_test.yml
+++ b/.github/workflows/run_hac_test.yml
@@ -1,0 +1,35 @@
+name: HAC on KIND cluster Test
+
+# Run on each new PR and each new push to existing PR
+on: [push, pull_request]
+
+jobs:
+  run-hac-test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+
+      - name: Set environment variables
+        run: |
+          echo PATH=$PATH:$HOME/go/bin                                          >> $GITHUB_ENV
+          echo OPERATOR_IMAGE=localhost:5000/noobaa/noobaa-operator:integration >> $GITHUB_ENV
+
+      - name: Deploy Dependencies
+        run: |
+          bash .travis/install-5nodes-kind-cluster.sh
+          go get -v github.com/onsi/ginkgo/ginkgo
+          ginkgo version
+
+      - name: Build NooBaa
+        run: |
+          make cli
+          make image
+          docker tag noobaa/noobaa-operator:$(go run cmd/version/main.go) $OPERATOR_IMAGE
+          docker push $OPERATOR_IMAGE
+
+      - name: Install NooBaa
+        run: ./build/_output/bin/noobaa-operator-local --mini --operator-image=$OPERATOR_IMAGE install
+
+      - name: Run HAC test
+        run: make test-hac
--- a/doc/high-availability-controller.md
+++ b/doc/high-availability-controller.md
@@ -1,0 +1,68 @@
+### High Availability (HA) controller
+
+## Abstract
+
+Improve NooBaa pods recovery in the case of a node failure. Find NooBaa pods running on the failing node and force to delete them to speed up rescheduling in the healthy node.
+
+## Problem
+
+* In cases where a node is in `NotReady` state, it takes 5 minutes for deployment pods (kubernetes Deployment to failover on a different node.
+* Statefulset pods, such as noobaa-core and noobaa-db pods, are not restarting automatically until the old pod is explicitly force deleted.
+* For pods that are connected to a PV, such as noobaa-db, after the pod is force deleted, it takes more time (~8 minutes) for the PV to detached from the old pod so that the new pod can attach to the PV.
+
+## Solution
+
+To make the pods failover faster to a new node, noobaa-operator [watches](https://kubernetes.io/docs/reference/using-api/api-concepts/#efficient-detection-of-changes) the cluster node states. When a node transitions from `Ready` to `NotReady` status, the HA Controller looks for NooBaa pods on that node, these pods will be force deleted. Once deleted the pods will restart on a new `Ready` node.
+
+```
+                            +-----------------+
+                            |  HA Controller  |
+                            +--------+--------+
+                                     |
+                                     |
+                                     |
++--------+     +---------------------+------------------------+
+|  ETCD  +-----+                 API Server                   |
++--------+     +----+----------------+-----------------+------+
+                    |                |                 |
+                    |                |                 |
+                    |                |                 |
+               +----+-----+     +----+-----+      +----+-----+
+               |   Node   |     |   Node   |      |   Node   |
+               +----------+     +----------+      +----------+
+```
+
+## Implementation
+
+[High Availability (HA) controller](https://github.com/noobaa/noobaa-operator/pull/672) is a [controller](https://sdk.operatorframework.io/docs/building-operators/golang/references/client/) defining kubernetes [Nodes](https://pkg.go.dev/k8s.io/api/core/v1#Node) as the source of its [events](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/event).
+
+### Node failure flow
+
+1. Communication between [K8S API server](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) and [kubelet](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) running on a worker node is severed
+2. API server marks the worker node state as `NotReady`
+3. HA Controller (HAC) watching cluster nodes states, detect a worker node state transition, reconciliation is initiated.
+4. The HAC Reconciler lists NooBaa pods on the failing node and requests API server to delete those pods. The new pod state is committed into ETCD
+5. [The pod controller](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/) (Deployment, StatefulSet, etc) reacts to pod deletion and reschedules the pod on a healthy node
+
+### Node readiness condition
+
+Node is ready, if there is a `NodeReady` [node condition](https://pkg.go.dev/k8s.io/api/core/v1#NodeCondition) in node's status. A worker node becomes not ready if the connection between the worker and the master node was broken, the node rebooted, or any other communication error between the K8S API Server and the kubelet process.
+
+### Predicate
+
+[Predicates](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate) allow controllers to filter events before they are provided to EventHandlers. There are several kinds of events, such as CreateEvents, GenericEvent, DeleteEvent and UpdateEvent. Update event where old Node state is `Ready` and current state is `NotReady` indicates a node goes down event. All other events are filtered out.
+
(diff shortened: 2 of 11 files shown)


guymguym (changes requested):
Hey @baum 
This PR looks great. Awesome doc!
Since this feature is a new controller, I want to first discuss its goal and scope.

Instead of node-watcher, I think this should be named and purposed as High Availability (HA) controller. Think about it this way - a reconciler's purpose is well defined by its target effect, not its source triggers, because that's the nature of reconcile - to always apply the desired state, based on any sources needed for the job.

So I would redefine this controller as the HA-controller, which is what we are trying to impose, and describe the problems that are currently in the scope - for now just node-shutdown/disconnect (or other not ready states?). I would also add section that walks through the flow of events on node shutdown - with and without this controller running.

About the code examples in "the solution" section in the doc - it makes the doc colorful and welcoming indeed, but I think that the code should *appear* once in the source tree rather than copied to the doc which is bound to become obsolete. I would compress those technical details to a a few sentences or bullets that go through the steps in the reconcile flow.

baum:
@guymguym sounds good, updated per [your concerns](https://github.com/noobaa/noobaa-operator/pull/672#pullrequestreview-693457148)

pkg/hac/hac.go
```diff
@@ -0,0 +1,85 @@
...
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"github.com/noobaa/noobaa-operator/v5/pkg/options"
+	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+)
+
+// Name of this controller
+const Name = "high-availability-controller"
```
romayalon, line 22: why Name defined here? isn't it just used when creating the controller?

pkg/hac/hac.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// looking for noobaa pods running on the failing node in the watched namespace
+	labelOption := client.MatchingLabels{"app": "noobaa"}
+	namespaceOption := client.InNamespace(options.Namespace)
+	nodeOption := client.MatchingFields{"spec.nodeName": hac.Node.Name}
+
+	// fetch the noobaa pods from the api server
+	podList := &corev1.PodList{}
+	if !util.KubeList(podList, labelOption, namespaceOption, nodeOption) {
+		return res, errors.Errorf("failed to list noobaa pods on the node %v in namespace %v", hac.Node.Name, options.Namespace)
```
romayalon, line 69: when failing don't you need to requeue ?

pkg/hac/hac.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// delete the found pods
+	for _, pod := range podList.Items {
+		var gracePeriod int64 = 0
+		deleteOpts := client.DeleteOptions{GracePeriodSeconds: &gracePeriod}
```
romayalon, line 75: since it's not changing, why not define it outside of the loop?

pkg/hac/hac.go
baum: Controller Name variable is referenced twice:
- in this file, line 43, in NewHAC() for logger configuration  
- in pkg/controller/ha/ha_controller.go, line 59, Add(), see https://github.com/baum/noobaa-operator/blob/37980cba41350d3c74e4f1589c558f6f9aab8a20/pkg/controller/ha/ha_controller.go#L59

pkg/hac/hac.go
baum: @romayalon 

> The Controller will requeue the Request to be processed again if an error is non-nil

From [here](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler), so returning non-nil error is enough.

pkg/hac/hac.go
baum: @romayalon moved outside the loop. 10x!

pkg/hac/hac.go
romayalon: right, I mean that in other reconcilers we are setting the res.RequeueAfter to 3 seconds, do you think 3 seconds is an appropriate period for this reconciler too?

baum: @romayalon  not sure RequeueAfter should be specified. In other reconcilers, some processes indeed happen, say deployments are scaling or volumes are provisioned, etc... In this controller, the intent is to address the specific event of Node state transition. So unless there is an error no re-queue is ever needed.   if there is an error (pods list or pod delete) no requirements on specific delay from this controller. WDYT?

romayalon: no specific delay means default delay, is the default delay is appropriate for this reconciler?

baum: why would not it be? 😊

baum: The default period is ASAP, honoring RateLimit, see [here](https://github.com/kubernetes-sigs/controller-runtime/blob/985e819db7af1cd17782ca7ad27644d0ca655054/pkg/internal/controller/controller.go#L300-L304). LGTM.

dannyzaken: @baum AFAIK the default requeue is done with exponential backoff. maybe to better control the reconcile and make it more consistent we should use a constant delay

baum:
Added     High Availability (HA) Controller integration test, see [here](https://github.com/noobaa/noobaa-operator/pull/672/commits/c814dbee188be76833f01b6e0c350a37eb01cd3d)

GitHub Action CI integration test using [KIND](https://kind.sigs.k8s.io/docs/user/quick-start/) cluster.
Simulate a node failure by killing a cluster node.
 Verify that NooBaa pods running on the failing node are deleted.

.github/workflows/run_hac_test.yml
```diff
@@ -0,0 +1,43 @@
+name: HAC on KIND cluster Test
+
+# Run on each new PR and each new push to existing PR
+on: [push, pull_request]
+
+env:
+  ACTIONS_ALLOW_UNSECURE_COMMANDS: "true"
+  GO111MODULE: "on"
+  GOPROXY: "https://proxy.golang.org"
+  KUBERNETES_VERSION: "v1.17.3"
```
liranmauda, line 10: Why Do we need this Env variable?

doc/high-availability-controller.md
```diff
@@ -0,0 +1,68 @@
+### High Availability (HA) controller
+
+## Abstract
+
+Improve NooBaa pods recovery in the case of a node failure. Find NooBaa pods running on the failing node and force to delete them to speed up rescheduling in healthy node.
```
liranmauda, line 5:
```suggestion
## Abstract

Improve NooBaa pods recovery in the case of a node failure. Find NooBaa pods running on the failing node and force to delete them to speed up rescheduling in the healthy node.
```

doc/high-availability-controller.md
```diff
@@ -0,0 +1,68 @@
...
+
+[High Availability (HA) controller](https://github.com/noobaa/noobaa-operator/pull/672) is a [controller](https://sdk.operatorframework.io/docs/building-operators/golang/references/client/) defining kubernetes [Nodes](https://pkg.go.dev/k8s.io/api/core/v1#Node) as the source of its [events](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/event).
+
+### Node failure flow
+
+1. Communication between [K8S API server](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) and [kubelet](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) running on a worker node is severed
+2. API server marks the worker node state as `NotReady`
+3. HA Controller (HAC) watching cluster nodes states, detects a worker node state transition, reconciliation is initiated.
+4. The HAC Reconciler lists NooBaa pods on the failing node and requests API server to delete those pods. The new pod state is commited into ETCD
+5. [The pod controller](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/) (Deployment, StatefulSet, etc) reacts to pod deletion and reschedules the pod on a healthy node
```
liranmauda, line 45:
```suggestion
### Node failure flow

1. Communication between [K8S API server](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) and [kubelet](https://kubernetes.io/docs/reference/command-line-tools-reference/kubelet/) running on a worker node is severed
2. API server marks the worker node state as `NotReady`
3. HA Controller (HAC) watching cluster nodes states, detect a worker node state transition, reconciliation is initiated.
4. The HAC Reconciler lists NooBaa pods on the failing node and requests API server to delete those pods. The new pod state is committed into ETCD
5. [The pod controller](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-controller-manager/) (Deployment, StatefulSet, etc) reacts to pod deletion and reschedules the pod on a healthy node
```

doc/high-availability-controller.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Node readiness condition
+
+Node is ready, if there is a `NodeReady` [node condition](https://pkg.go.dev/k8s.io/api/core/v1#NodeCondition) in node's status. A worker node becomes not ready if the connection between worker and master node was broken, node rebooted or any other communication error between the K8S API Server and the kubelet process.
```
liranmauda, line 49:
```suggestion
### Node readiness condition

Node is ready, if there is a `NodeReady` [node condition](https://pkg.go.dev/k8s.io/api/core/v1#NodeCondition) in node's status. A worker node becomes not ready if the connection between the worker and the master node was broken, the node rebooted, or any other communication error between the K8S API Server and the kubelet process.
```

doc/high-availability-controller.md
```diff
@@ -0,0 +1,68 @@
...
+
+### Predicate
+
+[Predicates](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/predicate) allow controllers to filter events before they are provided to EventHandlers. There are several kinds of events, such as CreateEvents, GenericEvent, DeleteEvent and UpdateEvent. Update event where old Node state is `Ready` and current state is `NotReady` indicates a node goes down event. All other events are filtered out.
+
+### Event handler
+
+`Reconcile()` is called when a node in the cluster transitions from `Ready` to `NotReady` state. High Availability (HA) controller lists all NooBaa pods in the failing node filtering using pods label, namespace and name of the node conditions:
```
liranmauda, line 57:
```suggestion
`Reconcile()` is called when a node in the cluster transitions from `Ready` to `NotReady` state. High Availability (HA) controller lists all NooBaa pods in the failing node filtering using pods label, namespace, and name of the node conditions:
```

pkg/hac/hac.go
baum: @dannyzaken, by design, it is desired to handle the failure event immediately. Not sure how introducing artificial delay is beneficial here.

.github/workflows/run_hac_test.yml
baum: Probably not, removed. 10x, @liranmauda !

dannyzaken (approved):
LGTM

PR #675, opened 2021-06-27
Load env vars to noobaa core from config map

### LINKED ISSUE #426: Load custom user env from a fixed name configmap
Since a lot of customizations are passed into the containers as env variables, 
we want to allow the user (noobaa admin) to persist some environment settings in a configmap and the noobaa containers will load.

kfir-payne:
Signed-off-by: Kfir <kfirpayne@gmail.com>

related to https://github.com/noobaa/noobaa-operator/issues/426

### Explain the changes
1. Added `envFrom: configMapRef` and `annotations: ConfigMapMD5` to the core and endpoint deployment.
2. The `ConfigMapMD5` annotation referencing to an up-to-date hash of the data array of the `noobaa-config`.
3. Added validation on changes to the configmap, triggering a restart on the core and endpoint pods.
4. Reduced the time interval in the `wait.PollImmediateInfinite()` in the `util.KubeDelete()` to cope with the STS creating the core pod before the `wait.Poll` checking if the pod got deleted, reduced from 1 second to 10 milliseconds.

### Work Needed:
1. Decide which env vars from the endpoint we can move to the configmap.

1. Fixes #426

### Testing Instructions:
1.

### DIFF
--- a/deploy/internal/deployment-endpoint.yaml
+++ b/deploy/internal/deployment-endpoint.yaml
@@ -4,6 +4,8 @@ metadata:
   labels:
     app: noobaa
   name: noobaa-endpoint
+  annotations:
+    ConfigMapHash: ""
 spec:
   replicas: 1
   selector:
@@ -50,6 +52,16 @@ spec:
             - containerPort: 6001
             - containerPort: 6443
           env:
+            - name: NOOBAA_DISABLE_COMPRESSION
+              valueFrom:
+                configMapKeyRef:
+                  name: noobaa-config
+                  key: NOOBAA_DISABLE_COMPRESSION
+            - name: NOOBAA_LOG_LEVEL
+              valueFrom:
+                configMapKeyRef:
+                  name: noobaa-config
+                  key: NOOBAA_LOG_LEVEL
             - name: MGMT_ADDR
             - name: BG_ADDR
             - name: MD_ADDR
@@ -68,10 +80,6 @@ spec:
             - name: LOCAL_N2N_AGENT
             - name: JWT_SECRET
             - name: NOOBAA_ROOT_SECRET
-            - name: NOOBAA_LOG_LEVEL
-              value: "0"
-            - name: NOOBAA_DISABLE_COMPRESSION
-              value: "false"
             - name: NOOBAA_AUTH_TOKEN
               valueFrom:
                 secretKeyRef:
--- a/deploy/internal/statefulset-core.yaml
+++ b/deploy/internal/statefulset-core.yaml
@@ -4,6 +4,8 @@ metadata:
   name: noobaa-core
   labels:
     app: noobaa
+  annotations:
+    ConfigMapHash: ""
 spec:
   replicas: 1
   selector:
@@ -61,6 +63,21 @@ spec:
             - containerPort: 8446
             - containerPort: 60100
           env:
+            - name: NOOBAA_DISABLE_COMPRESSION
+              valueFrom:
+                configMapKeyRef:
+                  name: noobaa-config
+                  key: NOOBAA_DISABLE_COMPRESSION
+            - name: DISABLE_DEV_RANDOM_SEED
+              valueFrom:
+                configMapKeyRef:
+                  name: noobaa-config
+                  key: DISABLE_DEV_RANDOM_SEED
+            - name: NOOBAA_LOG_LEVEL
+              valueFrom:
+                configMapKeyRef:
+                  name: noobaa-config
+                  key: NOOBAA_LOG_LEVEL
             - name: MONGODB_URL
               value: "mongodb://noobaa-db-0.noobaa-db/nbcore"
             - name: POSTGRES_HOST
@@ -73,10 +90,6 @@ spec:
               value: mongodb
             - name: CONTAINER_PLATFORM
               value: KUBERNETES
-            - name: NOOBAA_LOG_LEVEL
-              value: "0"
-            - name: NOOBAA_DISABLE_COMPRESSION
-              value: "false"
             - name: JWT_SECRET
               valueFrom:
                 secretKeyRef:
@@ -90,8 +103,6 @@ spec:
             - name: NOOBAA_ROOT_SECRET
             - name: AGENT_PROFILE
               value: VALUE_AGENT_PROFILE
-            - name: DISABLE_DEV_RANDOM_SEED
-              value: "true"
             - name: OAUTH_AUTHORIZATION_ENDPOINT
               value: ""
             - name: OAUTH_TOKEN_ENDPOINT
@@ -100,8 +111,6 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.serviceAccountName
-            - name: container_dbg
-              value: "" # any non-empty value will set the container to dbg mode
             - name: CONTAINER_CPU_REQUEST
               valueFrom:
                 resourceFieldRef:
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
-	"strconv"
 	"strings"
 	"time"
 
@@ -72,6 +71,10 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 	// Skip if joining another NooBaa
 	if r.JoinSecret != nil {
 		return nil
+	}
+
+	if err := r.ReconcileObject(r.CoreAppConfig, r.SetDesiredCoreAppConfig); err != nil {
+		return err
 	}
 
 	// A failure to discover OAuth endpoints should not fail the entire reconcile phase.
@@ -346,8 +349,6 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 				c.Env[j].Value = "mongodb://" + r.NooBaaMongoDB.Name + "-0." + r.NooBaaMongoDB.Spec.ServiceName + "/nbcore"
 			}
 
-		case "NOOBAA_LOG_LEVEL":
-			c.Env[j].Value = strconv.Itoa(r.NooBaa.Spec.DebugLevel)
 
 		case "POSTGRES_HOST":
 			c.Env[j].Value = r.NooBaaPostgresDB.Name + "-0." + r.NooBaaPostgresDB.Spec.ServiceName
@@ -465,6 +466,8 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 		}
 
 	}
+
+	r.SetConfigMapAnnotation(r.CoreApp.ObjectMeta.Annotations)
 
 	phase := r.NooBaa.Status.UpgradePhase
 	replicas := int32(1)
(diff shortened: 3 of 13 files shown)


deploy/internal/deployment-endpoint.yaml
```diff
@@ -4,6 +4,8 @@ metadata:
   labels:
     app: noobaa
   name: noobaa-endpoint
+  annotations:
+    ConfigMapMD5: ""
```
liranmauda, line 8: Why we call it `ConfigMapMD5` if it uses sha256?

pkg/system/reconciler.go
```diff
@@ -67,6 +67,7 @@ type Reconciler struct {
 	NooBaa                    *nbv1.NooBaa
 	ServiceAccount            *corev1.ServiceAccount
 	CoreApp                   *appsv1.StatefulSet
+	CoreAppConfig			  *corev1.ConfigMap
```
liranmauda, line 70: Indentation

dannyzaken:
@kfir-payne following our discussion:
since `KubeDelete` performs polling on the resource deletion, it is less fitting for controlled pods deletion. I suggest adding another function like `KubeDeleteNoPolling` (or some other better name).
when reconciling the config map, if there is a mismatch in the SHA between STS\deployment and the cm - fix the annotation on the STS\deployment, and then if there is no deletion timestamp on the pod you delete the pod

deploy/internal/deployment-endpoint.yaml
kfir-payne: Changed that to `ConfigMapHash`

pkg/system/reconciler.go
kfir-payne: Thats weird, in the VSCode it looks indented. 
any idea why here it doesn't?

kfir-payne:
@dannyzaken I added the logic you described above, let me know if this is what you meant, or if you have any other comments.

deploy/internal/statefulset-core.yaml
```diff
@@ -60,6 +62,9 @@ spec:
             - containerPort: 8445
             - containerPort: 8446
             - containerPort: 60100
+          envFrom:
```
dannyzaken, line 65: same as endpoints deployment

deploy/internal/deployment-endpoint.yaml
```diff
@@ -49,6 +51,9 @@ spec:
           ports:
             - containerPort: 6001
             - containerPort: 6443
+          envFrom:
```
dannyzaken, line 54: This will load all of the envs from the config map. since we want to have a single configmap for all pods I think it's better to pick specific envs and use `valueFrom`

dannyzaken (changes requested):

pkg/system/phase2_creating.go
```diff
@@ -143,6 +143,10 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 		return err
 	}
 
+	if err := r.ReconcileObject(r.CoreAppConfig, r.SetDesiredCoreAppConfig); err != nil {
```
dannyzaken, line 146: maybe better to move that to the beginning since the core pod will not start before the configmap is created

dannyzaken: actually, I'm not sure you want to reconcile the configmap, since you will override any changes done by the user. you just need to create once it if it's not present

deploy/internal/deployment-endpoint.yaml
kfir-payne: correct, but any new env vars added to the config map will not be loaded to the pods.
I don't know if its a use case we want to cover, if not, we can change to `valueFrom`.

pkg/system/phase2_creating.go
kfir-payne: not sure I understand what you mean, why I will override user changes if I'm reconciling the config map, we call `controllerutil.CreateOrUpdate` in `reconcileObject`, moreover, I dont see any user changes to the cm being deleted in the current behaviour.
I do accept moving the cm reconcile call to be called before the core pod.

dannyzaken: now when I think about it again, I was mistaken before. you do want to reconcile, just not to override any existing values. 
the reason we need to reconcile, is to handle any additional envs that we will add to the config map in future versions.

kfir-payne: @dannyzaken pushed a change regarding your last comment, let me know if the fix I implemented answers on the requirement.

pkg/system/phase2_creating.go
```diff
@@ -1111,3 +1117,35 @@ func (r *Reconciler) SetEndpointsDeploymentReplicas(replicas int32) error {
...
+			updateNeeded = true
+		}
+	}
+	if updateNeeded {
+		if !util.KubeUpdate(r.CoreAppConfig) {
+			r.Logger.Errorf("Failed to load initial data into %v", r.CoreAppConfig.Name)
+		}
+	}
+	return nil
+}
```
dannyzaken, line 1142: there is no need to call `KubeUpdate` here. ReconcileObject will already update the object if needed. you just need to set the configmap with the desired values

kfir-payne:
@dannyzaken @liranmauda pushed fixes to what we discussed about, would like your review on the changes.

pkg/system/phase4_configuring.go
```diff
@@ -1401,3 +1405,43 @@ func (r *Reconciler) ReconcileNamespaceStores(namespaceResources []nb.NamespaceR
...
+		epPodList := &corev1.PodList{}
+		epPodSelector, _ := labels.Parse("noobaa-s3=" + r.Request.Name)
+		util.KubeList(epPodList, &client.ListOptions{Namespace: options.Namespace, LabelSelector: epPodSelector})
+		r.DeploymentEndpoint.ObjectMeta.Annotations["ConfigMapHash"] = sha256Hex
+
+		if !util.KubeUpdate(r.DeploymentEndpoint) {
+			r.Logger.Info("NooBaa %q failed to update Deployment Endpoint Config Map Hash", options.SystemName)
+		}
+		for _, pod := range epPodList.Items {
+			util.KubeDelete(&pod)
```
dannyzaken, line 1423: use `KubeDeleteNoPolling`

pkg/system/phase4_configuring.go
```diff
@@ -1401,3 +1405,43 @@ func (r *Reconciler) ReconcileNamespaceStores(namespaceResources []nb.NamespaceR
...
+	sha256Hex := util.GetCmDataHash(r.CoreAppConfig.Data)
+
+	if (r.DeploymentEndpoint.ObjectMeta.Annotations["ConfigMapHash"] != sha256Hex) {
+		epPodList := &corev1.PodList{}
+		epPodSelector, _ := labels.Parse("noobaa-s3=" + r.Request.Name)
+		util.KubeList(epPodList, &client.ListOptions{Namespace: options.Namespace, LabelSelector: epPodSelector})
+		r.DeploymentEndpoint.ObjectMeta.Annotations["ConfigMapHash"] = sha256Hex
+
+		if !util.KubeUpdate(r.DeploymentEndpoint) {
+			r.Logger.Info("NooBaa %q failed to update Deployment Endpoint Config Map Hash", options.SystemName)
```
dannyzaken, line 1420: should be r.Logger.Infof
```suggestion
			r.Logger.Infof("NooBaa %q failed to update Deployment Endpoint Config Map Hash", options.SystemName)
```

pkg/system/phase4_configuring.go
```diff
@@ -1401,3 +1405,43 @@ func (r *Reconciler) ReconcileNamespaceStores(namespaceResources []nb.NamespaceR
...
+
+	corePodList := &corev1.PodList{}
+	corePodSelector, _ := labels.Parse("noobaa-core=" + r.Request.Name)
+	util.KubeList(corePodList, &client.ListOptions{Namespace: options.Namespace, LabelSelector: corePodSelector})
+
+	if (r.CoreApp.ObjectMeta.Annotations["ConfigMapHash"] != sha256Hex){
+		r.CoreApp.ObjectMeta.Annotations["ConfigMapHash"] = sha256Hex
+		
+		if !util.KubeUpdate(r.CoreApp) {
+			r.Logger.Info("NooBaa %q failed to update Core STS Config Map Hash", options.SystemName)
```
dannyzaken, line 1435:
```suggestion
			r.Logger.Infof("NooBaa %q failed to update Core STS Config Map Hash", options.SystemName)
```

dannyzaken:
@kfir-payne didn't we say that debugLevel will be removed from the CRD? no reason to have it duplicated. 
as for the debug-level flag, if set you can just create the config map in the system `RunCreate`

kfir-payne:
@dannyzaken don't you think its more readable and simple to have just one place where the config-map get created and being set with default values?
I can see the redundancy but I think its easier to have a single place on setting the desired state of the config map.

dannyzaken:
@kfir-payne but then what is the behavior if the user sets one value of the debug level in the configmap and a different in the CR?

kfir-payne:
@dannyzaken I see, fixing it now.

dannyzaken (approved):

test/cli/test_cli_functions.sh
```diff
@@ -180,6 +180,34 @@ function noobaa_install {
...
+    local core_debug_level=$(kuberun silence exec noobaa-core-0 -- printenv NOOBAA_LOG_LEVEL)
+
+    kuberun silence patch configmap noobaa-config -p ${patch}
+
+    while [[ "${core_debug_level}" != "${cm_debug_level}" ]]
+    do
+        echo_time "💬  Waiting for NOOBAA_LOG_LEVEL core env var to match the noobaa-config"
+        sleep 10
+        core_debug_level=$(kuberun silence exec noobaa-core-0 -- printenv NOOBAA_LOG_LEVEL)
+    done < <(sleep 360)
```
liranmauda, line 200: Why do we want to sleep that long? 
+ I think it is more straightforward to do the sleep inside the loop:
```suggestion
    while [[ "${core_debug_level}" != "${cm_debug_level}" ]]
    do
        echo_time "💬  Waiting for NOOBAA_LOG_LEVEL core env var to match the noobaa-config"
        sleep 10
        core_debug_level=$(kuberun silence exec noobaa-core-0 -- printenv NOOBAA_LOG_LEVEL)
        sleep 360
    done 
```

kfir-payne: the `sleep 360` outside the loop is just a timeout so we won't get stuck waiting for the update, the way I see it in your suggestion we are checking one time if the update occurred and than sleeping 360 seconds.

liranmauda: I don't see how it act as a timeout:
```
$ count=1 ; while true ; do echo ${count} ; ((count++)) ; sleep 2 ; done < <(sleep 6)
1
2
3
4
5
6
7
8
9
10
11
12
```
So if you are missing some sleep inside the loop, you just keep checking, In your case you are probably succeeding, have the loop ends.

liranmauda: In your example: `done < <(sleep 360)` you are passing the `fd` as an input to the loop, but you are not doing anything with it

kfir-payne: you are right, fixed that with a timeout condition in the while loop.

liranmauda (approved):
LGTM

Issue #677, opened 2021-07-04
Namespacestore NSFS creation with sub-path "/"


jeniawhite:
### Environment info
- NooBaa Operator Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
1. We allow to create namespacestore with a non-relative path like "/"
2. The creation passes and even when I do a status on the namespacestore I'm getting that everything is valid
3. When I try to do `nb status` I can see that it gets stuck in Configuration stage
4. Inside the operator I can see the following errors in the reconcile:
time="2021-07-04T13:06:16Z" level=error msg="ReconcileObject: Error Deployment /apis/apps/v1/namespaces/jenia-nsfs/deployments/noobaa-endpoint Deployment.apps \"noobaa-endpoint\" is invalid: spec.template.spec.containers[0].volumeMounts.subPath: Invalid value: \"/\": must be a relative path" sys=jenia-nsfs/noobaa
time="2021-07-04T13:06:16Z" level=info msg="SetPhase: temporary error during phase \"Configuring\"" sys=jenia-nsfs/noobaa
time="2021-07-04T13:06:16Z" level=warning msg="⏳ Temporary Error: Deployment.apps \"noobaa-endpoint\" is invalid: spec.template.spec.containers[0].volumeMounts.subPath: Invalid value: \"/\": must be a relative path" sys=jenia-nsfs/noobaa
time="2021-07-04T13:06:16Z" level=info msg="UpdateStatus: Done generation 1" sys=jenia-nsfs/noobaa

### Expected behavior
1. We should probably block non relative sub-paths because the deployment doesn't allow such behavior

### Steps to reproduce
1. ```nb namespacestore create nsfs nsfs-nss --fs-backend='CEPH_FS' --pvc-name='nsfs-vol' --sub-path='/'```

PR #681, opened 2021-07-07
Added namespacestore validation

igorpick:
Link to issue:  
https://github.com/noobaa/noobaa-operator/issues/677

1. Added validation to the namespacestore reconciler and noobaa cli
  * Validate if Spec.NSFS.SubPath must be a relative path
  * Validate if Spec.NSFS.SubPath must not contain '..'
  * Validate if Spec.NSFS.PvcName must not be empty

2. Move all config validations from MakeExternalConnectionParams to the validator

3. Skip DesiredEndpointMounts when namespacestore is in rejected phase

4. Added the validator automation tests

Tests:
```
$noobaa namespacestore create nsfs fs1 --pvc-name='nsfs-vol' --fs-backend='GPFS' --sub-path='/'
INFO[0000] ✅ Exists: NooBaa "noobaa"
FATA[0000] ❌ SubPath / must be a relative path

Options:
      --fs-backend='': The file system backend type - CEPH_FS | GPFS | NFSv4
      --pvc-name='': The pvc name in which the file system resides
      --sub-path='': The path to a sub directory inside the pvc file system

Usage:
  noobaa namespacestore create nsfs <namespace-store-name> [flags] [options]

```

```
$kubectl apply -f namespacestore.yaml
namespacestore.noobaa.io/fs1 created

time="2021-07-07T09:26:27Z" level=info msg="SetPhase: Rejected" namespace=noobaa/fs1
time="2021-07-07T09:26:27Z" level=error msg="❌ Persistent Error: SubPath / must be a relative path" namespace=noobaa/fs1
time="2021-07-07T09:26:27Z" level=info msg="SetPhase: Rejected" namespace=noobaa/fs1
time="2021-07-07T09:26:27Z" level=error msg="❌ Persistent Error: SubPath / must be a relative path" namespace=noobaa/fs1
time="2021-07-07T09:26:34Z" level=warning msg="namespacestore fs1 is in rejected phase" source="SystemReconciler:setDesiredEndpointMounts" sys=noobaa/noobaa
time="2021-07-07T09:26:45Z" level=warning msg="namespacestore fs1 is in rejected phase" source="SystemReconciler:setDesiredEndpointMounts" sys=noobaa/noobaa
```

### DIFF
--- a/pkg/namespacestore/reconciler.go
+++ b/pkg/namespacestore/reconciler.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"net/url"
-	"regexp"
 	"time"
 
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
@@ -138,6 +137,7 @@ func (r *Reconciler) Reconcile() (reconcile.Result, error) {
 			return res, nil
 		}
 	}
+
 	system.CheckSystem(r.NooBaa)
 	var err error
 
@@ -260,6 +260,11 @@ func (r *Reconciler) ReconcilePhaseVerifying() error {
 		"noobaa operator started phase 1/3 - \"Verifying\"",
 	)
 
+	err := ValidateNamespaceStore(r.NamespaceStore)
+	if err != nil {
+		return err
+	}
+
 	if r.NooBaa.UID == "" {
 		return util.NewPersistentError("MissingSystem",
 			fmt.Sprintf("NooBaa system %q not found or deleted", r.NooBaa.Name))
@@ -359,16 +364,17 @@ func (r *Reconciler) ReconcileDeletion() error {
 					allowedBuckets.PermissionList = []string{}
 				}
 				err := r.NBClient.UpdateAccountS3Access(nb.UpdateAccountS3AccessParams{
-					Email:        account.Email,
-					S3Access:     account.HasS3Access,
-					DefaultResource:  &internalPoolName,
-					AllowBuckets: &allowedBuckets,
+					Email:           account.Email,
+					S3Access:        account.HasS3Access,
+					DefaultResource: &internalPoolName,
+					AllowBuckets:    &allowedBuckets,
 				})
 				if err != nil {
 					return err
 				}
 			}
 		}
+
 		err := r.NBClient.DeleteNamespaceResourceAPI(nb.DeleteNamespaceResourceParams{Name: r.NamespaceResourceinfo.Name})
 		if err != nil {
 			if rpcErr, isRPCErr := err.(*nb.RPCError); isRPCErr {
@@ -438,11 +444,11 @@ func (r *Reconciler) ReadSystemInfo() error {
 
 	// handling namespace fs resource
 	if r.NamespaceStore.Spec.Type == nbv1.NSStoreTypeNSFS {
-		fsRootPath := "/nsfs/" + r.NamespaceStore.Name 
+		fsRootPath := "/nsfs/" + r.NamespaceStore.Name
 		r.CreateNamespaceResourceParams = &nb.CreateNamespaceResourceParams{
 			Name: r.NamespaceStore.Name,
 			NSFSConfig: &nb.NSFSConfig{
-				FsBackend: r.NamespaceStore.Spec.NSFS.FsBackend,
+				FsBackend:  r.NamespaceStore.Spec.NSFS.FsBackend,
 				FsRootPath: fsRootPath,
 			},
 			NamespaceStore: &nb.NamespaceStoreInfo{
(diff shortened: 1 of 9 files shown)


nimrod-becker:
Should we add this to the test flows as well?

igorpick:
@nimrod-becker you are right, there should be a unit test. I will try to figure out how to add unit tests there.

igorpick:
> Should we add this to the test flows as well?

you are right, there should be a unit test. I will try to figure out how to add unit tests there.

nimrod-becker:
@igorpick we already have a test which creates OBC and BS and NS, I think you can just add a case to that test

pkg/namespacestore/validation.go
```diff
@@ -0,0 +1,36 @@
...
+import (
+	"fmt"
+	"strings"
+
+	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+)
+
+// ValidateNamespaceStore validates namespacestore configuration
+func ValidateNamespaceStore(nsStore *nbv1.NamespaceStore) error{
```
romayalon, line 12: validateNamespaceStore is a too much general name, you specifically check here validation for namespacestore of type NSFS

pkg/namespacestore/validation.go
```diff
@@ -0,0 +1,36 @@
...
+
+	//pvcName validation
+	if nsStore.Spec.NSFS != nil && nsStore.Spec.NSFS.PvcName == "" {
```
romayalon, line 15: instead of checking `nsStore.Spec.NSFS != nil` twice you can check it at the beginning of the function

pkg/namespacestore/validation.go
igorpick: you are right, there are only NSFS validation for now, but the method is a place to put there all additional namestore validations as well.

pkg/namespacestore/reconciler.go
```diff
@@ -201,6 +201,10 @@ func (r *Reconciler) Reconcile() (reconcile.Result, error) {
 // ReconcilePhases runs the reconcile flow and populates System.Status.
 func (r *Reconciler) ReconcilePhases() error {
 
+	if err := r.ReconcilePhaseValidation(); err != nil {
```
romayalon, line 204: ReconcilePhaseValidation implies that there is a new validation phase for namespacestore, but you didn't add this new phase. in my opinion, we don't need this phase and you can just call ValidateNamespaceStore in phase connecting

igorpick: I forgot to add the validation phase. thanks
the connection phase is too late, in my opinion, the validation phase should be on the top.

pkg/namespacestore/validation.go
romayalon: for other types of namespacestores the validation already exists and taking place inside the function MakeExternalConnectionParams.
I don't find it correct to have only one function in that file, you will probably want to have a switch case in the original function and create multiple validation functions for each type of namespacestore,
but anyway, please move all validations to the file if yo desire to have a validation file for all types

pkg/namespacestore/validation.go
```diff
@@ -0,0 +1,160 @@
...
+//Endpoint validation and sets default
+func validateEndPoint(endPointPointer *string) error {
+	endPoint := *endPointPointer
+
+	if endPoint == "" {
+		endPoint = defaultEndPointURI
+	}
+
+	if endPoint == "" {
+		return util.NewPersistentError("ConfigurationError", "endpoint url must be not empty")
```
dannyzaken, line 137: why do you return an error here?

pkg/namespacestore/validation.go
```diff
@@ -0,0 +1,160 @@
...
+
+	return nil
+}
+
+//SignatureVersion validation, must be empty or v2 or v4
+func validateSignatureVersion(signature nbv1.S3SignatureVersion, nsStoreName string) error {
+	if signature != "" && 
+		signature != nbv1.S3SignatureVersionV2 && 
+		signature != nbv1.S3SignatureVersionV4 {
+		return util.NewPersistentError("ConfigurationError",
```
dannyzaken, line 121: notice that the same `ConfigurationError` reason is set for all persistent errors here. this changes the reasons we had before (e.g. `InvalidSignatureVersion`, `InvalidEndpoint` etc.).

when encountering a persistent error in the reconcile loop it is eventually translated into an event and the reason is passed to the event, so we want to keep those unchaged. see here:

https://github.com/igorpick/noobaa-operator/blob/462a265ddf238fc334e60937c45355845cf73ce3/pkg/backingstore/reconciler.go#L190-L197

pkg/namespacestore/validation.go
igorpick: In my opinion, the same reason why you asked me to add the additional validation in the MakeExternalConnectionParams.
if someone changes the default URI to empty, raise the error.

dannyzaken: I think we didn't understand each other. of course you shouldn't have the same validations in `MakeExternalConnection`. 
I did think we should make sure that functions like `getAuthMethod` will validate that the signature version is valid and not just return `v4` in the else clause.

in this case, there is no code between the 2 checks, so the second one is redundant.

pkg/namespacestore/validation.go
```diff
@@ -0,0 +1,160 @@
...
+		signature != nbv1.S3SignatureVersionV4 {
+		return util.NewPersistentError("InvalidSignatureVersion",
+			fmt.Sprintf("Invalid s3 signature version %q for namespace store %q",
+			signature, nsStoreName))
+	}
+	return nil
+}
+
+//Endpoint validation and sets default
+func validateEndPoint(endPointPointer *string) error {
```
dannyzaken, line 129: I think that some of the functions here are generic ones and not only relevant for namespacestore. especially `validateEndPoint`, `validateSignatureVersion`, and `getAuthMethod`. maybe it's better to move them to `util.go`. we can later want to use them elsewhere

pkg/namespacestore/reconciler.go
```diff
@@ -546,100 +552,40 @@ func (r *Reconciler) MakeExternalConnectionParams() (*nb.AddExternalConnectionPa
...
-				u.Scheme = "https"
-				// if s3Options.SSLDisabled {
-				// 	u.Scheme = "http"
-				// }
-			}
-			conn.Endpoint = u.String()
+		conn.AuthMethod = method
+
+		//Validate and configure endPoint
+		err = validateEndPoint(&s3Compatible.Endpoint)
```
dannyzaken, line 564: this validation is indeed redundant

dannyzaken (commented):
some small comments. other than that it looks good. 
great tests!

pkg/namespacestore/validation.go
igorpick: I agree, but in my opinion, we should do that when we will use it elsewhere.
For now, we don't know where and when.

pkg/namespacestore/reconciler.go
```diff
@@ -775,3 +705,13 @@ func (r *Reconciler) ReconcileNamespaceStore() error {
...
 	return nil
 }
+
+//getAuthMethod get auth method based on s3 signature
+func getAuthMethod(signature nbv1.S3SignatureVersion, nsStoreName string) nb.CloudAuthMethod {
+
+	if signature == nbv1.S3SignatureVersionV4 {
+		return "AWS_V4"
+	} 
+	return "AWS_V2"
```
igorpick, line 715: Return empty string when signuture is empty

pkg/namespacestore/validation.go
```diff
@@ -0,0 +1,160 @@
...
+	endPoint := *endPointPointer
+
+	if endPoint == "" {
+		endPoint = defaultEndPointURI
+	}
+
+	if endPoint == "" {
+		return util.NewPersistentError("InvalidEndpoint", "endpoint url must be not empty")
+	}
```
dannyzaken, line 138: remove this check, it is redundant

dannyzaken (approved):

PR #687, opened 2021-07-14
if dbtype is not provided try to infer it from the db image name

dannyzaken:

since dbType is optional in our CRD it is sometimes omitted in the CR. if it is not supplied then try to infer it from the DB image so it will match

fix for https://bugzilla.redhat.com/show_bug.cgi?id=1969677


liranmauda (approved):
LGTM

agarwal-mudit:
Can we please backport this to 5.7?

Issue #684, opened 2021-07-11
The operator should update a backingstore external connection when the credentials are changed


dannyzaken:
### Environment info
- NooBaa Operator Version: **5.8**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
1. when a secret is updated with new credentials the operator tries to reconcile it by creating a new external connection in noobaa core
2. since a connection with the same name already exists, `ReconcileExternalConnection` fails and the reconcile is stuck in `Creating` phase

### Expected behavior
1. when the credentials are updated in the backingstore secret, the operator should call update_external_connection

### Steps to reproduce
1. create an S3\S3 compatible backing store
2. update the BS secret with other (valid) credentials
3. the BS is stuck in creating

PR #688, opened 2021-07-14
Added ability to update BackingStore external connection

kfir-payne:

related to https://github.com/noobaa/noobaa-operator/issues/684

### Explain the changes
1. Added a Watch on backing-store Secret events to detect a change in the secret and trigger  a reconcile.
2. Changed `EditExternalConnectionCredentialsAPI` (seems unused) to `UpdateExternalConnectionAPI` to call the relevant  RPC call when needed.
3. Added detection of a mismatch in the used connection by the operator and trying to update the connection.
4. the mismatch found on the reconciler will be based on secret key mismatch and will not detect a mismatch in the access key.

1. https://bugzilla.redhat.com/show_bug.cgi?id=1992090

### Testing Instructions:
1. Create two backingstores with the same cloud credentials.
2. Create two namespacestores with the same cloud credentials.

### DIFF
--- a/build/gen-odf-package.sh
+++ b/build/gen-odf-package.sh
@@ -27,7 +27,7 @@ head -n ${n} ${MANIFESTS}/${CSV_NAME} > ${temp_csv}
 cat >> ${temp_csv} << EOF
   relatedImages:
   - image: ${CORE_IMAGE}
-    name: noboaa-core
+    name: noobaa-core
   - image: ${DB_IMAGE}
     name: noobaa-db
   - image: ${OPERATOR_IMAGE}
--- a/pkg/backingstore/backingstore.go
+++ b/pkg/backingstore/backingstore.go
@@ -369,17 +369,46 @@ func createCommon(cmd *cobra.Command, args []string, storeType nbv1.StoreType, p
 		log.Fatalf(`❌ %s %s`, validationErr, cmd.UsageString())
 	}
 
+	suggestedSecret := util.CheckForIdenticalSecretsCreds(secret, util.MapStorTypeToMandatoryProperties[storeType])
+	if suggestedSecret != nil {
+		var decision string
+		log.Printf("Found a Secret in the system with the same credentials (%s)", suggestedSecret.Name)
+		log.Printf("Note that using more then one secret with the same credentials is not supported")
+		log.Printf("do you want to use it for this Backingstore? y/n")
+		fmt.Scanln(&decision)
+		if strings.ToLower(decision) == "y" {
+			log.Printf("Will use %s as the Backingstore %s Secret", suggestedSecret.Name, backStore.Name)
+			err := util.SetBackingStoreSecretRef(backStore, &corev1.SecretReference{
+				Name:      suggestedSecret.Name,
+				Namespace: suggestedSecret.Namespace,
+			})
+			if err != nil {
+				log.Fatalf(`❌ %s`, err)
+			}
+		} else if strings.ToLower(decision) == "n" {
+			log.Fatalf("Not creating Backingstore")
+		} else {
+			log.Fatalf(`❌ Invalid input, please select y/n`)
+		}
+	}
+
 	// Create backing store CR
 	util.Panic(controllerutil.SetControllerReference(sys, backStore, scheme.Scheme))
 	if !util.KubeCreateFailExisting(backStore) {
 		log.Fatalf(`❌ Could not create BackingStore %q in Namespace %q (conflict)`, backStore.Name, backStore.Namespace)
 	}
 
-	if GetBackingStoreSecret(backStore) != nil && secretName == "" {
+	secretRef, _ := util.GetBackingStoreSecret(backStore)
+	if secretRef != nil && secretName == "" && suggestedSecret == nil {
 		// Create secret
 		util.Panic(controllerutil.SetControllerReference(backStore, secret, scheme.Scheme))
 		if !util.KubeCreateFailExisting(secret) {
 			log.Fatalf(`❌ Could not create Secret %q in Namespace %q (conflict)`, secret.Name, secret.Namespace)
+		}
+	} else if secretRef != nil && secretName != "" {
+		_, err := util.GetSecretFromSecretReference(secretRef)
+		if err != nil {
+			util.Logger().Fatalf(`❌ Could not found Secret %q from SecretReference`, secret.Name)
 		}
 	}
 
@@ -450,6 +479,7 @@ func RunCreateAWSS3(cmd *cobra.Command, args []string) {
 		targetBucket := util.GetFlagStringOrPrompt(cmd, "target-bucket")
 		region, _ := cmd.Flags().GetString("region")
 		secretName, _ := cmd.Flags().GetString("secret-name")
+		mandatoryProperties := []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"}
 
 		if secretName == "" {
 			accessKey := util.GetFlagStringOrPromptPassword(cmd, "access-key")
@@ -457,10 +487,11 @@ func RunCreateAWSS3(cmd *cobra.Command, args []string) {
 			secret.StringData["AWS_ACCESS_KEY_ID"] = accessKey
 			secret.StringData["AWS_SECRET_ACCESS_KEY"] = secretKey
 		} else {
-			mandatoryProperties := []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"}
 			util.VerifyCredsInSecret(secretName, options.Namespace, mandatoryProperties)
 			secret.Name = secretName
+			secret.Namespace = options.Namespace
 		}
+
 		backStore.Spec.AWSS3 = &nbv1.AWSS3Spec{
 			TargetBucket: targetBucket,
 			Region:       region,
@@ -479,6 +510,7 @@ func RunCreateS3Compatible(cmd *cobra.Command, args []string) {
 		targetBucket := util.GetFlagStringOrPrompt(cmd, "target-bucket")
 		sigVer, _ := cmd.Flags().GetString("signature-version")
 		secretName, _ := cmd.Flags().GetString("secret-name")
+		mandatoryProperties := []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"}
 
 		if secretName == "" {
 			accessKey := util.GetFlagStringOrPromptPassword(cmd, "access-key")
@@ -486,9 +518,9 @@ func RunCreateS3Compatible(cmd *cobra.Command, args []string) {
 			secret.StringData["AWS_ACCESS_KEY_ID"] = accessKey
 			secret.StringData["AWS_SECRET_ACCESS_KEY"] = secretKey
 		} else {
-			mandatoryProperties := []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"}
 			util.VerifyCredsInSecret(secretName, options.Namespace, mandatoryProperties)
 			secret.Name = secretName
+			secret.Namespace = options.Namespace
 		}
 
 		backStore.Spec.S3Compatible = &nbv1.S3CompatibleSpec{
@@ -510,6 +542,7 @@ func RunCreateIBMCos(cmd *cobra.Command, args []string) {
 		targetBucket := util.GetFlagStringOrPrompt(cmd, "target-bucket")
 		// sigVer, _ := cmd.Flags().GetString("signature-version")
 		secretName, _ := cmd.Flags().GetString("secret-name")
+		mandatoryProperties := []string{"IBM_COS_ACCESS_KEY_ID", "IBM_COS_SECRET_ACCESS_KEY"}
 
 		if secretName == "" {
 			accessKey := util.GetFlagStringOrPromptPassword(cmd, "access-key")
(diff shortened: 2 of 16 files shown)


romayalon:
@kfir-payne I know it's not written in the original bug but we want to add this code to namespacestore too :)

pkg/controller/backingstore/backingstore_controller.go
```diff
@@ -71,6 +71,10 @@ func Add(mgr manager.Manager) error {
 	if err != nil {
 		return err
 	}
+	err = c.Watch(&source.Kind{Type: &corev1.Secret{}}, ownerHandler, &filterForOwnerPredicate, &logEventsPredicate)
```
romayalon, line 74: this will only watch on secrets their owner is the backingstore, what about other secrets?

kfir-payne: The goal is to watch the backing store secrets, and not trigger a reconcile when not needed.

kfir-payne:
@romayalon Sure, I'll add it to namespacestore too, the logic should be the same, anything you think that I should pay attention to?

pkg/backingstore/reconciler.go
```diff
@@ -826,14 +832,22 @@ func (r *Reconciler) fixAlternateKeysNames() {
...
-	// TODO we only support creation here, but not updates
 	if r.ExternalConnectionInfo != nil {
 		return nil
 	}
 	if r.AddExternalConnectionParams == nil {
 		return nil
 	}
 
+	if r.UpdateExternalConnectionParams != nil {
+		err := r.NBClient.UpdateExternalConnectionAPI(*r.UpdateExternalConnectionParams)
```
romayalon, line 843: what is the reason for not calling checkExternalConnection before updating? I know it's called from inside updateExternalConnection in noobaa-core but I think it's needed for keeping the same behaviour as for AddExternalConnection

pkg/controller/backingstore/backingstore_controller.go
romayalon: of course you don't want to trigger reconcile when it's not needed, but you can create a backingstore while specifying a secret - then the backingstore won't own the secret and then you won't get triggers on these secrets at all

pkg/backingstore/reconciler.go
kfir-payne: calling checkExternalConnection before updating will return a `CONNECTION_ALREADY_EXIST` error, as you said it's called from inside updateExternalConnection, and if there is something wrong with the connection it will be checked there.

romayalon: it will be checked but you won't throw a persistent error and the whole logic there is missing.

pkg/controller/backingstore/backingstore_controller.go
kfir-payne: Didn’t fully understood what you meant, can you explain again please?

dannyzaken: @kfir-payne look at how we reconcile bucket class on backingstore changes using `EnqueueRequestsFromMapFunc` in `bucketclass_controller`

pkg/backingstore/backingstore.go
```diff
@@ -343,21 +343,33 @@ func createCommon(cmd *cobra.Command, args []string, storeType nbv1.StoreType, p
...
 			util.VerifyCredsInSecret(secretName, options.Namespace, mandatoryProperties)
 			secret.Name = secretName
+			secret = util.GetSecretFromSecretReference(&corev1.SecretReference{
+				Name:      secret.Name,
+				Namespace: secret.Namespace,
+			})
+			util.SecretResetStringDataFromData(secret)
+		}
+
+		suggestedSecret := util.CheckForIdenticalSecretsCreds(secret, mandatoryProperties)
```
romayalon, line 368: why do you prevent that? and how similarly it's being handled in the reconcilers?
Also, I think that a better behaviour from NooBaa's side is not to fail the command, but just to show the already existing secret name and ask the user if noobaa should proceed creating the backingstore with the already existing secret you found / fail the command.

romayalon:
@kfir-payne Could you please explain the new behaviour ?

pkg/backingstore/backingstore.go
```diff
@@ -367,21 +373,51 @@ func createCommon(cmd *cobra.Command, args []string, storeType nbv1.StoreType, p
...
+
+		if suggestedSecret != nil {
+
+			var decision string
+			for {
+				log.Printf("Found a Secret in the system with the same credintials (%s), do you want to use it for this Backingstore? y/n", suggestedSecret.Name)
+				fmt.Scanln(&decision)
```
romayalon, line 408: do we still need it? didn't we say in our discussion that we don't need it at all and we will update the secrets?
if your answer is still yes I feel that putting this code in every type of backingstore/namespacestore CLI creation command is unneeded and we can put this logic in the createCommon function

pkg/backingstore/reconciler.go
```diff
@@ -599,6 +611,11 @@ func (r *Reconciler) ReadSystemInfo() error {
 			pool.CloudInfo.Endpoint != conn.Endpoint ||
 			pool.CloudInfo.Identity != conn.Identity {
 			r.Logger.Warnf("using existing pool but connection mismatch %+v pool %+v %+v", conn, pool, pool.CloudInfo)
+			r.UpdateExternalConnectionParams = &nb.UpdateExternalConnectionParams{
+				Name:     conn.Name,
+				Identity: conn.Identity,
+				Secret:   conn.Secret,
+			}
```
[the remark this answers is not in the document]
kfir-payne: this would be more complicated to implement, because the pool cloud info from `ReadSystemInfo()` not returning the secret key the connection is using.

pkg/controller/backingstore/backingstore_controller.go
```diff
@@ -73,6 +75,21 @@ func Add(mgr manager.Manager) error {
...
+		return err
+	}
+
+	secretsHandler := handler.EnqueueRequestsFromMapFunc(func(obj client.Object) []reconcile.Request {
+		return backingstore.MapSecretToBackingStores(types.NamespacedName{
+			Name: obj.GetName(),
+			Namespace: obj.GetNamespace(),
+		})
+	})
+	err = c.Watch(&source.Kind{Type: &corev1.Secret{}}, secretsHandler, logEventsPredicate)
```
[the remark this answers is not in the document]
kfir-payne: this watch is with a different handler

pkg/backingstore/backingstore.go
```diff
@@ -450,17 +466,25 @@ func RunCreateAWSS3(cmd *cobra.Command, args []string) {
...
 		if secretName == "" {
 			accessKey := util.GetFlagStringOrPromptPassword(cmd, "access-key")
 			secretKey := util.GetFlagStringOrPromptPassword(cmd, "secret-key")
 			secret.StringData["AWS_ACCESS_KEY_ID"] = accessKey
 			secret.StringData["AWS_SECRET_ACCESS_KEY"] = secretKey
 		} else {
-			mandatoryProperties := []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"}
 			util.VerifyCredsInSecret(secretName, options.Namespace, mandatoryProperties)
 			secret.Name = secretName
+			secret, err := util.GetSecretFromSecretReference(&corev1.SecretReference{
```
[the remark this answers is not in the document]
kfir-payne: it's not really a check, it's just to get the full secret object from the secret reference.

pkg/util/util.go
```diff
@@ -1614,3 +1623,183 @@ func GetAvailabeKubeCli() string {
...
+	bsList := &nbv1.BackingStoreList{
+		TypeMeta: metav1.TypeMeta{Kind: "BackingStoreList"},
+	}
+	KubeList(bsList, &client.ListOptions{Namespace: secret.Namespace})
+	KubeList(nsList, &client.ListOptions{Namespace: secret.Namespace})
+
+	for _, bs := range bsList.Items {
+		if bs.Spec.Type != nbv1.StoreTypePVPool {
+			usedSecret, err := GetSecretFromSecretReference(GetBackingStoreSecret(&bs))
+			if err != nil {
```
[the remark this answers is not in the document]
kfir-payne: there a little similarity in the functionality but there are some differences to handle with when is't BS or NS, therefor I don't think this should be moved out to a separate function.
for example, `if bs.Spec.Type != nbv1.StoreTypePVPool` or `if ns.Spec.Type != nbv1.NSStoreTypeNSFS` and the process of extracting the secret from the resource.

pkg/util/util.go
```diff
@@ -1614,3 +1623,183 @@ func GetAvailabeKubeCli() string {
...
+	if secret == nil {
+		return nil
+	}
+	nsList := &nbv1.NamespaceStoreList{
+		TypeMeta: metav1.TypeMeta{Kind: "NamespaceStoreList"},
+	}
+	bsList := &nbv1.BackingStoreList{
+		TypeMeta: metav1.TypeMeta{Kind: "BackingStoreList"},
+	}
+	KubeList(bsList, &client.ListOptions{Namespace: secret.Namespace})
```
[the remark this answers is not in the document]
dannyzaken: I think that the main problem with listing the secrets is how we can distinguish between noobaa and non-noobaa related secrets? unless we have a good way to do that I think it's ok to start from the BS\NSS lists

pkg/util/util.go
```diff
@@ -1614,3 +1627,287 @@ func GetAvailabeKubeCli() string {
...
+func SetNamespaceStoreSecretRef(ns *nbv1.NamespaceStore, ref *corev1.SecretReference) error {
+	switch ns.Spec.Type {
+	case nbv1.NSStoreTypeAWSS3:
+		ns.Spec.AWSS3.Secret = *ref
+		return nil
+	case nbv1.NSStoreTypeS3Compatible:
+		ns.Spec.S3Compatible.Secret = *ref
+		return nil
+	case nbv1.NSStoreTypeIBMCos:
+		ns.Spec.AzureBlob.Secret = *ref
```
romayalon, line 1725: ns.Spec.IBMCos.Secret = *ref

pkg/util/util.go
```diff
@@ -1614,3 +1627,287 @@ func GetAvailabeKubeCli() string {
...
+					}
+				}
+			}
+		}
+	}
+	return nil
+}
+
+// SetOwnerReference setting a owner reference of owner to dependent metadata with the field of blockOwnerDeletion: true
+// controllerutil.SetOwnerReference is doing the same thing but without blockOwnerDeletion: true
```
romayalon, line 1840: please add this link https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L96
to your comment so it'll be better explained where it is taken from.

Also, I think that the comment https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/controller/controllerutil/controllerutil.go#L59 is super important in order to understand the logic of the owner refernece and why we can not put conroller reference on the secret (explains the backingstore and namespacestore controller changes) , please add it there

pkg/backingstore/backingstore.go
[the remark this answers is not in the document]
kfir-payne: I moved this check from every single BS/NS creation to the `createCommon` function

pkg/backingstore/backingstore.go
```diff
@@ -369,13 +369,35 @@ func createCommon(cmd *cobra.Command, args []string, storeType nbv1.StoreType, p
...
+		if strings.ToLower(decision) == "y" {
+			log.Printf("Will use %s as the Backingstoe Secret", suggestedSecret.Name)
+			err := util.SetBackingStoreSecretRef(backStore, &corev1.SecretReference{
+				Name:      suggestedSecret.Name,
+				Namespace: suggestedSecret.Namespace,
+			})
+			if err != nil {
+				log.Fatalf(`❌ %s`, err)
+			}
+		} else if strings.ToLower(decision) == "n" {
```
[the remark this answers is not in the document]
kfir-payne: I can print a more "relaxed" message with no ❌, just explaining the creation can't be proceed
@romayalon WDYT?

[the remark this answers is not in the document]
romayalon: my intention was that I wouldn't even provide the n option, I would just fail

pkg/util/util.go
kfir-payne: Nice catch!

pkg/util/util.go
kfir-payne: Good idea

pkg/backingstore/backingstore.go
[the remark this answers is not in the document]
kfir-payne: in that case, I think your suggestion is even a worse user experience, since the user has no option at all.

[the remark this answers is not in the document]
romayalon: Providing n as an option and then failing is not a different option as well, you make him think that he can use a new secret instead of the suggested one but then you fail...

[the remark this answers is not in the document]
kfir-payne: that's why I'm suggesting printing a more indicative message explaining that he must use that secret, the `log.Fatalf` is there just to stop the process.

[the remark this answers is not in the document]
romayalon: I would explain that while providing the "y/n" options and not after he already chose n

romayalon (approved):
@MeridianExplorer @kfir-payne a discussion about tests for this PR is needed, specifically upgrade scenario.

build/gen-odf-package.sh
```diff
@@ -27,7 +27,7 @@ head -n ${n} ${MANIFESTS}/${CSV_NAME} > ${temp_csv}
 cat >> ${temp_csv} << EOF
   relatedImages:
   - image: ${CORE_IMAGE}
-    name: noboaa-core
```
dannyzaken, line 30: it took me some time to understand what is the difference here :) 
In general, I think that it's better to make these kind of changes in a different PR, so it will be easier to backport if needed.
In this specific case, it shouldn't be a porblem, but worth checking 
https://docs.openshift.com/container-platform/4.9/operators/operator_sdk/osdk-generating-csvs.html#olm-enabling-operator-for-restricted-network_osdk-generating-csvs

dannyzaken (approved):

PR #691, opened 2021-07-19
Extend Quota support

igorpick:
Added support for max objects quota.

- Provide an option to Configure Quota per Bucket on an OBC level and on the BucketClass level. If both exist, the OBC definition would override the BucketClass definition. 
- Set a quota of object bucket claim, which limits the max number of objects per bucket (MaxObjects).
- Set a quota of object bucket claim, which limits the max total size per bucket (MaxSize).
- Update BucketClass quota, will be effective in new buckets only using this BucketClass
- Add to NooBaa CLI new options for creating object bucket claims while providing one of the 2 types of quota


pkg/nb/types.go
```diff
@@ -649,7 +683,50 @@ func IntToHumanBytes(bi int64) string {
...
 	return BigInt{
-		Peta: int64(value / petaInBytes),
-		N:    int64(value % petaInBytes),
+		Peta: int64(value / PETABYTE),
+		N:    int64(value % PETABYTE),
+	}
+}
+
+// HumanBytesToValueAndUnit convert human bytes to value and unit. 10G -> value: 10, unit GIGABYTE
+func HumanBytesToValueAndUnit(str string) (int, string, error) {
```
dannyzaken, line 692: I think you should use kubernetes `resource.ParseQuantity` instead of parsing yourself. 
see rook implementation: 

https://github.com/rook/rook/blob/fa3ab50eff4795d357041805f0d5a873eb2f4fcc/pkg/operator/ceph/object/bucket/provisioner.go#L580-L591

igorpick: resource.ParseQuantity returns value not in bytes. for example 1M = 1000000 and not 1048576, however the noobaa-core parses the maxSize to bytes value.

dannyzaken: in kuberentes there is a difference between `1M` and `1Mi`, this is probably why you don't get what you expect
https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits

igorpick: Since the noobaa-core translates the value into bytes only, do you still want to use this method?
Or do you prefer to implement the same logic in the core as well?

PR #662, opened 2021-06-17
Changes due to updating dependencies to k8s 1.20.7 and controller-runtime 0.8.3

VolkerKozlowski:
Updated dependencies to k8s v1.20.7 and controller-runtime v0.8.3.

PR #692, opened 2021-07-21
Bumping k8s to 1.20.2 and  controller-runtime to 0.8.3

liranmauda:
### Explain the changes: 

- Bumping k8s to 1.20.2
- Bumping controller-runtime to 0.8.3
- Bumping `github.com/openshift/cloud-credential-operator` 

- event changes: 
  - `MetaOld` to `ObjectOld`
  -  `MetaNew` to `ObjectNew`
  -  `Meta` to `Object`

- Changed `runtime.Object` to `client.Object` (and in case of list to `client.ObjectList`)
- Updated the `EnqueueRequestsFromMapFunc` according to the new changes
- Add `context` where needed
- Added `util.SecretResetStringDataFromData` after Calling `controllerutil.CreateOrUpdate` in system reconcile 
- Removed the Panic from `func ObjectKey` as `client.ObjectKeyFromObject` does not return error anymore

### Update according to those Breaking changes:
1. https://github.com/kubernetes-sigs/controller-runtime/pull/1118
2. https://github.com/kubernetes-sigs/controller-runtime/commit/d61f56105d593b23652b04d9db194a82b18db9d8#diff-5a927c01385831c16ce8c52c9b3ac094d9e830c2719ca5431d3bc500daca47fe
3. https://github.com/kubernetes-sigs/controller-runtime/commit/f35d8116c61c051d7e2db1051e88d1e51bf33460

### Related: 
https://github.com/noobaa/noobaa-operator/pull/662

### Fixes:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1892335
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=1892337


liranmauda:
Please ignore the number of commits, LMLM in the code, and that it is not signed.
I will wrap it out when we approve it.

Also please pay attention to the fact that I had to add `util.SecretResetStringDataFromData` in several places, for some reason after the bump of controller-runtime, not all the places that we looked at StringData were populated.

dannyzaken:
@liranmauda I think we need to understand why `util.SecretResetStringDataFromData` is not working inside `KubeCheck`. having to call it outside of KubeCheck is very bug-prone

liranmauda:
Never said it is not working. 
The timing has changed, so sometimes we try to reach `StringData` and it is empty, or in different cases, the map was not even initialized yet.

liranmauda:
After @dannyzaken and I looked, we saw that [StringData](https://pkg.go.dev/k8s.io/api@v0.20.2/core/v1#Secret) is getting zeroed after Calling `controllerutil.CreateOrUpdate`

pkg/backingstore/reconciler.go
```diff
@@ -967,6 +967,9 @@ func (r *Reconciler) ReconcilePool() error {
 }
 
 func (r *Reconciler) reconcilePvPool() error {
+	if r.Secret.StringData == nil {
```
dannyzaken, line 970: is it needed now, after you added `SecretResetStringDataFromData` to `ReconcileObject`?

pkg/system/phase4_configuring.go
```diff
@@ -107,6 +107,7 @@ func (r *Reconciler) ReconcileSystemSecrets() error {
 	if err := r.ReconcileObject(r.SecretOp, r.SetDesiredSecretOp); err != nil {
 		return err
 	}
+
```
dannyzaken, line 110: undo changes in this file

dannyzaken (approved):

pkg/backingstore/reconciler.go
liranmauda: Tested it and yes, It is still needed.

PR #693, opened 2021-07-21
Skip namespacestores with invalid configuration only

igorpick:
In order to avoid the removing existing mount points when namespacestore gets the rejected state,
ignore the state and try to configure the mounts.

### DIFF
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -399,8 +399,6 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 
 func (r *Reconciler) setDesiredEndpointMounts(podSpec *corev1.PodSpec, container *corev1.Container) error {
 
-	log := r.Logger.WithField("source", "SystemReconciler:setDesiredEndpointMounts")
-
 	namespaceStoreList := &nbv1.NamespaceStoreList{}
 	if !util.KubeList(namespaceStoreList, client.InNamespace(options.Namespace)) {
 		return fmt.Errorf("Error: Cant list namespacestores")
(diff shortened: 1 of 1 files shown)


jeniawhite (approved):

pkg/system/phase4_configuring.go
```diff
@@ -455,6 +455,37 @@ func (r *Reconciler) setDesiredEndpointMounts(podSpec *corev1.PodSpec, container
 	return nil
 }
 
+// Duplicate code from validation.go namespacetore pkg.
+// Cannot import the namespacestore pkg, because the pkg imports the system pkg
+// TODO remove the code
+func (r *Reconciler) validateNsStoreNSFS(nsStore *nbv1.NamespaceStore) bool {
+	nsfs := nsStore.Spec.NSFS
+
+	if nsfs == nil {
```
jackyalbo, line 464: I think this check is repeating itself - but as we want to remove this code later - I think it's fine

pkg/system/phase4_configuring.go
```diff
@@ -409,8 +407,10 @@ func (r *Reconciler) setDesiredEndpointMounts(podSpec *corev1.PodSpec, container
 	container.VolumeMounts = r.DefaultDeploymentEndpoint.Containers[0].VolumeMounts
 
 	for _, nsStore := range namespaceStoreList.Items {
-		if nsStore.Status.Phase == nbv1.NamespaceStorePhaseRejected {
-			log.Warnf("namespacestore %s is in rejected phase", nsStore.Name)
```
jackyalbo, line 413: maybe leave this warning next to the continue - just change to namespace store is not valid. WDYT?

igorpick: You told me that you don't want to see this warning each 3 sec, no?

pkg/system/phase4_configuring.go
igorpick: Could you explain me pls, why do you think it's repeating?

jackyalbo: OK, We will have the status in the namespacestore describe. leave it

jackyalbo (approved):

pkg/system/phase4_configuring.go
jackyalbo: After our small discussion - it's fine we will get the reject reason from the namespacestore describe

Issue #689, opened 2021-07-14
Add "Are you sure? Y/N" when uninstalling NooBaa with --cleanup flag


romayalon:
### Environment info
- NooBaa Operator Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
1. noobaa uninstall --cleanup removes the CRDs and affecting all noobaa instances

### Expected behavior
1. add a warning and a yes/no confirmation for this dangerous command.

### Steps to reproduce
1. noobaa uninstall --cleanup

PR #694, opened 2021-07-25
Adding confirmation check for uninstall command

BarakSason:
Since executing unintall command with --cleanup flag removes all CRDs
and affecting all noobaa instances, a confirmation check is added in
order to inform the user and verify that the above is the intended
action.

Fixed: #689

Signed-off-by: Barak Sason Rofman <sason922@gmail.com>

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -107,6 +107,24 @@ func RunInstall(cmd *cobra.Command, args []string) {
 // RunUninstall runs a CLI command
 func RunUninstall(cmd *cobra.Command, args []string) {
 	log := util.Logger()
+	cleanup, _ := cmd.Flags().GetBool("cleanup")
+
+	if cleanup {
+		var decision string
+		
+		for {
+			log.Printf("--cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n ")
+			fmt.Scanln(&decision)
+
+			if (decision == "y") {
+				log.Printf("Will remove CRD (cluster scope)")
+				break
+			} else if (decision == "n") {
+				return
+			}
+		}
+	}
+
 	system.RunSystemVersionsStatus(cmd, args)
 	log.Printf("Namespace: %s", options.Namespace)
 	log.Printf("")
@@ -116,7 +134,6 @@ func RunUninstall(cmd *cobra.Command, args []string) {
 	log.Printf("Operator Delete:")
 	operator.RunUninstall(cmd, args)
 	log.Printf("")
-	cleanup, _ := cmd.Flags().GetBool("cleanup")
 	if cleanup {
 		log.Printf("CRD Delete:")
 		crd.RunDelete(cmd, args)
--- a/test/cli/test_cli_functions.sh
+++ b/test/cli/test_cli_functions.sh
@@ -534,7 +534,7 @@ function noobaa_uninstall {
     [ ${check_cleanup_data_flag} -eq 0 ] && cleanup_data="--cleanup_data"
 
     echo_time "💬  Running uninstall ${cleanup} ${cleanup_data}"
-    test_noobaa --timeout uninstall ${cleanup} ${cleanup_data}
+    yes | test_noobaa --timeout uninstall ${cleanup} ${cleanup_data}
     if [ ${check_cleanflag} -eq 0 ]
     then
         check_if_cleanup


liranmauda:
I would expect to bail out completely from the uninstall if the answer is no.
WDYT?

liranmauda:
Please fix the tests, when we use `--cleanup` flag inject yes.

test/cli/test_cli_functions.sh
```diff
@@ -534,7 +534,7 @@ function noobaa_uninstall {
     [ ${check_cleanup_data_flag} -eq 0 ] && cleanup_data="--cleanup_data"
 
     echo_time "💬  Running uninstall ${cleanup} ${cleanup_data}"
-    test_noobaa --timeout uninstall ${cleanup} ${cleanup_data}
+    echo "y" | test_noobaa --timeout uninstall ${cleanup} ${cleanup_data}
```
liranmauda, line 537:
```suggestion
    yes | test_noobaa --timeout uninstall ${cleanup} ${cleanup_data}
```

BarakSason: 1. Would you like me to change the user input to `yes/no` (at the moment the input is `y/n`)?
2. Are you sure that the above is valid? I've created a 1 line script which calls `noobaa uninstall` to test this functionality, and without the `echo` it doesn't work.

liranmauda (changes requested):

test/cli/test_cli_functions.sh
liranmauda: > 1. Would you like me to change the user input to `yes/no` (at the moment the input is `y/n`)?
I think `y/n` is good.
> 2. Are you sure that the above is valid? I've created a 1 line script which calls `noobaa uninstall` to test this functionality, and without the `echo` it doesn't work.
works for me: 
```
$ yes | build/_output/bin/noobaa-operator-local uninstall --cleanup
INFO[0000] --cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n
INFO[0000] Will remove CRD (cluster scope)
INFO[0000] CLI version: 5.9.0
INFO[0000] noobaa-image: noobaa/noobaa-core:master-20210717
INFO[0000] operator-image: liranmauda/noobaa-operator:crt8347
INFO[0000] noobaa-db-image: centos/postgresql-12-centos7
INFO[0000] Namespace: default
INFO[0000]
INFO[0000] System Delete:
```

BarakSason: Something is weird here.
You used `yes` which is not what the code expected (y/n), but for some reason instead of asking you again for input, it continued with the uninstall - Same happens to me (I don't see how the code is wrong though, any idea?).
Please try with `y` instead of `yes` - Without the echo it throws me in an endless loop:
```
INFO[0000] --cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n  
INFO[0000] --cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n  
INFO[0000] --cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n  
INFO[0000] --cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n  
INFO[0000] --cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n  
INFO[0000] --cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n  
INFO[0000] --cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n  
INFO[0000] --cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n  
```
Can you confirm please?

pkg/install/install.go
```diff
@@ -107,6 +107,24 @@ func RunInstall(cmd *cobra.Command, args []string) {
...
+
+	if cleanup {
+		var decision string
+		
+		for {
+			log.Printf("--cleanup removes the CRDs and affecting all noobaa instances, are you sure? y/n ")
+			fmt.Scanln(&decision)
+
+			if (decision == "y") {
+				cleanup = true
```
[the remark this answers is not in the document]
BarakSason: There is a check for the value of `cleanup` in line 137

test/cli/test_cli_functions.sh
liranmauda: `yes` is a Linux command that prints a string repetitively, and the default is `y`, for reference:  [yes man page](https://man7.org/linux/man-pages/man1/yes.1.html)
What you are experiencing is looks like it prints something else the `y` and `n` 
Can you check the yes command output?

pkg/install/install.go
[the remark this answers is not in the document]
liranmauda: But nothing is changing it along the way...
It will get to line 137 populated with the same value it got from line 110.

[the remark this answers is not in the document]
BarakSason: You're right, it'll be true anyway, will fix in next revision.

test/cli/test_cli_functions.sh
BarakSason: Got it, thanks

liranmauda (approved):

PR #701, opened 2021-08-03
5.7: Delete the Old Service Monitor 

liranmauda:
- DeleteOldServiceMonitor updates the service monitor by removing the old format. This function should only be valid after upgrade from 5.6 to 5.7

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1973179

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -48,6 +48,7 @@ rules:
   - update
   - list
   - watch
+  - delete
 - apiGroups:
   - ""
   resources:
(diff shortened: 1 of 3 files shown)


liranmauda:
Thinking on it we might need to do the same to 4.8 as customers that are on 4.7 and have this issue can upgrade to 4.8 …

nimrod-becker (dismissed):

nimrod-becker (approved):

PR #703, opened 2021-08-03
added make target to generate CSV for ODF

dannyzaken:

* added `gen-odf-package` make target to generate CSV and CRDs for ODF builds
* this target should be used with these required envs: 
```
csv-name=[csv yaml file name] core-image=[noobaa core image] operator-image=[noobaa operator image] db-image=[postgres db image]
```
* The files are generated under `build/_output/manifests`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,6 +21,7 @@ GOHOSTOS ?= $(shell go env GOHOSTOS)
 OUTPUT ?= build/_output
 BIN ?= $(OUTPUT)/bin
 OLM ?= $(OUTPUT)/olm
+MANIFESTS ?= $(OUTPUT)/manifests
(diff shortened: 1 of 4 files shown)


nimrod-becker:
@obnoxxx FYI

b-ranto:
I was able to produce a DS noobaa-operator-bundle build with this branch. I am not sure if the csv-name argument is actually required (afaik, the name should be fixed[1]) but that is ok, we can add that as an argument DS.

[1] I guess it might make sense if we wanted to call it `mcg-operator` instead of `noobaa-operator` DS?

b-ranto (approved):
lgtm

liranmauda (approved):
LGTM

PR #707, opened 2021-08-08
Modifying diagnose mechanism

BarakSason:
This patch changes the diagnose mechanism in the following ways:
1 - Executing diagnose will collect logs for pv-pool as well
2 - Instead of manually specifying for which pods logs should be
collected, logs are automatically collected for all pods (if in the
future additional pods will be added, logs will be collected for those
as well without requiring changes)
3 - The mechanism is enhanced internally, as prior to this changes
multiple listing of pods was being done, with this change only a single
listing is performed
4 - Some code cleaning is done as part of this change

Fixed: #260

Signed-off-by: Barak Sason Rofman <sason922@gmail.com>

### DIFF
--- a/pkg/diagnose/diagnose.go
+++ b/pkg/diagnose/diagnose.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"strings"
 	"time"
 
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
@@ -57,35 +56,14 @@ func RunCollect(cmd *cobra.Command, args []string) {
 		c.log.Fatalf(`❌ Could not create directory %s, reason: %s`, c.folderName, err)
 	}
 
-	c.CollectCR(&nbv1.BackingStoreList{
-		TypeMeta: metav1.TypeMeta{Kind: "BackingStoreList"},
-	})
+	// Define to select only noobaa pods within the namespace
+	podSelector, _ := labels.Parse("app=noobaa")
+	listOptions := client.ListOptions{Namespace: options.Namespace, LabelSelector: podSelector}
 
-	c.CollectCR(&nbv1.BucketClassList{
-		TypeMeta: metav1.TypeMeta{Kind: "BucketClassList"},
-	})
-
-	c.CollectCR(&nbv1.NooBaaList{
-		TypeMeta: metav1.TypeMeta{Kind: "NooBaaList"},
-	})
-
-	corePodSelector, _ := labels.Parse("noobaa-core=" + options.SystemName)
-	c.CollectPodLogs(corePodSelector)
-
-	operatorPodSelector, _ := labels.Parse("noobaa-operator=deployment")
-	c.CollectPodLogs(operatorPodSelector)
-
-	endpointPodSelector, _ := labels.Parse("noobaa-s3=" + options.SystemName)
-	c.CollectPodLogs(endpointPodSelector)
-
-	dbPodSelector, _ := labels.Parse("noobaa-db=" + options.SystemName)
-	if options.DBType == "postgres" {
-		dbPodSelector, _ = labels.Parse("noobaa-db=" + options.DBType)
-	}
-	c.CollectPodLogs(dbPodSelector)
-
-	// collectSystemMetrics()
-
+	c.CollectCRs()
+	c.CollectPodsLogs(listOptions)
+	c.CollectPVs(listOptions)
+	c.CollectPVCs(listOptions)
 	c.CollectSCC()
 
 	c.ExportDiagnostics(destDir)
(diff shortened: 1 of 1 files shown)


pkg/diagnose/diagnose.go
```diff
@@ -131,31 +117,27 @@ func (c *Collector) collectPodDescribe(pod *corev1.Pod) {
...
-	if !util.KubeList(corePodList, &client.ListOptions{Namespace: options.Namespace, LabelSelector: corePodSelector}) {
-		return
-	}
-	if len(corePodList.Items) == 0 {
-		c.log.Printf(`❌ No %s pods found\n`, currentPod)
+// CollectPodsLogs collects logs of all exsisting pods
+func (c *Collector) CollectPodsLogs() {
+	// List all pods
+	podList := &corev1.PodList{}
+	if !util.KubeList(podList, &client.ListOptions{Namespace: options.Namespace}) {
```
liranmauda, line 124: > 2 - Instead of manually specifying for which pods logs should be
> collected, logs are automatically collected for all pods (if in the
> future additional pods will be added, logs will be collected for those
> as well without requiring changes)

Is this mean that in an openshift cluster with OCS it will collect all pods logs regardless if they are noobaa related or not? 
In that case, I think it is not the desired outcome.

pkg/diagnose/diagnose.go
```diff
@@ -131,31 +117,27 @@ func (c *Collector) collectPodDescribe(pod *corev1.Pod) {
...
-// CollectPodLogs info
-func (c *Collector) CollectPodLogs(corePodSelector labels.Selector) {
-	corePodList := &corev1.PodList{}
-	currentPod := strings.Split(corePodSelector.String(), "=")[0]
-	if !util.KubeList(corePodList, &client.ListOptions{Namespace: options.Namespace, LabelSelector: corePodSelector}) {
-		return
-	}
-	if len(corePodList.Items) == 0 {
-		c.log.Printf(`❌ No %s pods found\n`, currentPod)
+// CollectPodsLogs collects logs of all exsisting pods
```
liranmauda, line 120:
```suggestion
// CollectPodsLogs collects logs of all existing pods
```

pkg/diagnose/diagnose.go
nimrod-becker: Agree, we cannot assume that our pods are the only ones in the NS

BarakSason: Added a pod selector to select only noobaa pods (assuming all noobaa pods have the "app=noobaa" label)

pkg/diagnose/diagnose.go
BarakSason: Done

pkg/diagnose/diagnose.go
```diff
@@ -131,31 +118,31 @@ func (c *Collector) collectPodDescribe(pod *corev1.Pod) {
...
-		c.log.Printf(`❌ No %s pods found\n`, currentPod)
+// CollectPodsLogs collects logs of all existing pods
+func (c *Collector) CollectPodsLogs() {
+	// Define to select only noobaa pods
+	podSelector, _ := labels.Parse("app=noobaa")
+
+	// List all pods and select only noobaa pods within the relevant namespace
+	podList := &corev1.PodList{}
+	if !util.KubeList(podList, &client.ListOptions{Namespace: options.Namespace,
+		LabelSelector: podSelector}) {
```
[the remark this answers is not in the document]
BarakSason: Done

liranmauda (approved):
LGTM

liranmauda:
> LGTM

Wait, We also need to collect the noobaa related `pv` and `pvc`

pkg/diagnose/diagnose.go
```diff
@@ -109,82 +87,101 @@ func (c *Collector) CollectCR(list client.ObjectList) {
...
+		TypeMeta: metav1.TypeMeta{Kind: "BucketClassList"},
+	})
+
+	c.CollectCR(&nbv1.NooBaaList{
+		TypeMeta: metav1.TypeMeta{Kind: "NooBaaList"},
+	})
+}
+
+// CollectDescribe collects output of the "describe pod" of a single pod
+func (c *Collector) CollectDescribe(Entity string, Name string) {
```
liranmauda, line 106:
```suggestion
func (c *Collector) CollectDescribe(Kind string, Name string) {
```

pkg/diagnose/diagnose.go
```diff
@@ -109,82 +87,101 @@ func (c *Collector) CollectCR(list client.ObjectList) {
...
+	cmd := exec.Command("kubectl", "describe", Entity, "-n", options.Namespace, Name)
```
liranmauda, line 107:
```suggestion
	cmd := exec.Command("kubectl", "describe", Kind, "-n", options.Namespace, Name)
```

pkg/diagnose/diagnose.go
```diff
@@ -109,82 +87,101 @@ func (c *Collector) CollectCR(list client.ObjectList) {
...
 	// handle custom path for kubeconfig file,
 	// see --kubeconfig cli options
 	if len(c.kubeconfig) > 0 {
 		cmd.Env = append(cmd.Env, "KUBECONFIG=" + c.kubeconfig)
 	}
 
 	// open the out file for writing
-	fileName := c.folderName + "/" + pod.Name + "-describe.txt"
+	fileName := c.folderName + "/" + Name + "-describe.txt"
```
liranmauda, line 115: We should probably add to the filename the kind: 
```suggestion
	fileName := c.folderName + "/" + Name + Kind + "-describe.txt"
```

liranmauda (changes requested):

liranmauda:
One General remark, can we print what we are currently collecting? 
it tasks some time to collect all logs, and without the printing, we look stuck.

pkg/diagnose/diagnose.go
BarakSason: Done

pkg/diagnose/diagnose.go
BarakSason: Done

BarakSason:
> One General remark, can we print what we are currently collecting?
> it tasks some time to collect all logs, and without the printing, we look stuck.

Done

liranmauda (approved):
LGTM

PR #710, opened 2021-08-10
Diagnose will collect scc related to noobaa

liranmauda:
- Diagnose will collect scc related to noobaa

### DIFF
--- a/pkg/diagnose/diagnose.go
+++ b/pkg/diagnose/diagnose.go
@@ -10,6 +10,7 @@ import (
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	"github.com/noobaa/noobaa-operator/v5/pkg/options"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+	secv1 "github.com/openshift/api/security/v1"
 	"github.com/sirupsen/logrus"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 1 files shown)


pkg/diagnose/diagnose.go
```diff
@@ -160,6 +163,47 @@ func (c *Collector) CollectPodLogs(corePodSelector labels.Selector) {
...
+// CollectSCC collects the SCC 
+func (c *Collector) CollectSCC() {
+	for _, name := range []string{"noobaa", "noobaa-endpoint"} {
+		scc := &secv1.SecurityContextConstraints {
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: options.Namespace,
+			},
+		}
+		_, _, err := util.KubeGet(scc)
```
jackyalbo, line 200: Why not KubeCheckOptional?

jackyalbo (commented):
LGTM

pkg/diagnose/diagnose.go
liranmauda: We only want to see that we can get is, chance it exists.
We don't want to check all the others (CRD, ect')

jackyalbo: We usually use KubeCheck, and I think it's much more readable just to do normal one if instead of handling the errors

jackyalbo (approved):

PR #715, opened 2021-08-13
initdb huge_pages

baum:
Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=1946792

- Extract Postgres initialization into second init container
- During initialization wrap the Postgres binary, force huge_pages=off for initdb

### DIFF
--- a/deploy/internal/configmap-postgres-initdb.yaml
+++ b/deploy/internal/configmap-postgres-initdb.yaml
@@ -1,0 +1,31 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: noobaa-postgres-initdb-sh
+  labels:
+    app: noobaa
+data:
+  initdb.sh: |
+          # If the config file is present, the DB is initiazed
+          # and we're out of here
+          export PGDATA=$HOME/data/userdata
+          if [ -f $PGDATA/postgresql.conf ]; then
+            echo postgresql.conf file is found
+            exit 0
+          fi
+
+          # Wrap the postgres binary, force huge_pages=off for initdb
+          # see https://bugzilla.redhat.com/show_bug.cgi?id=1946792
+          p=/opt/rh/rh-postgresql12/root/usr/bin/postgres
+          mv $p $p.orig
+          echo exec $p.orig \"\$@\" -c huge_pages=off > $p
+          chmod 755 $p
+
+          # The NooBaa DB runs with UID 10001 GID 0
+          sed -i -e 's/^\(postgres:[^:]\):[0-9]*:[0-9]*:/\1:10001:0:/' /etc/passwd
+
+          # Init the DB and exit once the DB is ready to run
+          sed -i -e 's/^exec.*$/exit 0/' \
+                 -e 's/^pg_ctl\sstart.*/pg_ctl start || true/'                                   \
+                    /usr/bin/run-postgresql
+          su postgres -c "bash -x /usr/bin/run-postgresql"
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -194,6 +194,31 @@ func (r *Reconciler) SetDesiredServiceDBForPostgres() error {
 	return nil
 }
 
+func (r *Reconciler) setPostgresUserPass(c *corev1.Container) {
+	for j := range c.Env {
+		switch c.Env[j].Name {
+		case "POSTGRESQL_USER":
+			c.Env[j].ValueFrom = &corev1.EnvVarSource{
+				SecretKeyRef: &corev1.SecretKeySelector{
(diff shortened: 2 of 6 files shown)


deploy/internal/configmap-postgres-initdb.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: noobaa-postgres-initdb-sh
+  labels:
+    app: noobaa
+data:
+  initdb.sh: |
```
liranmauda, line 8: Can we divide this into a script and not write the whole script in the yaml?

baum: @liranmauda , the idea is to inject this script into the existing Postgres image. WDYT?

dannyzaken:
@guymguym we had discussions in the past about how we should manage our RBAC resources. WDYT on these changes?

a few things that are done in this PR, as a result of the need to have more privileges for the DB init container:
1. a new `noobaa-db` account was added
2. the CSV was changed to only contain the `noobaa` account\role\binding.
3. noobaa install from the CLI does not install `noobaa-db` and `noobaa-endpoint` accounts\roles
4. `noobaa-endpoint` and `noobaa-db` RBAC resources along with SCC are created in the system reconcile.

we should also notice that noobaa-core is still using `noobaa` account, which gives it all of the k8s permissions noobaa operator has. we might want to change that in the future

guymguym:
@dannyzaken @baum 
1. I can't seem to understand why is this all needed for our postgres deployment and not any other postgres deployment.
2. Why do we deploy different RBAC for OLM vs CLI - that can't be good.
3. Why put a script in a configmap? Is this expected to be configured by a user? If not I would expect either a custom image or to write this script inline in the deployment yaml instead in its own configmap which is hinting towards a configuration.

guymguym:
First of all I want to paste the references to the issues that we are trying to workaround here:
1. https://github.com/kubernetes/kubernetes/issues/71233 this is the kubernetes issue that causes postgres to not be able to auto detect that hugepages are disabled, and instead the memory allocation passes, but then there's a runtime Bus error.
2. https://github.com/postgres/postgres/blob/1882d6cca161dcf9fa05ecab5abeb1a027a5cfd2/src/bin/initdb/initdb.c#L2775 - this is the postgres initdb code that runs this autodetection very early in the db creation flow, which means that we cannot affect it by changing the configuration files and we have to force the binary to use the `-c huge_pages=off` in some hacky way.

To be continued.

guymguym:
Hey @baum @dannyzaken 

My main concern with the proposed PR is its size and collateral damage for later.
We certainly want to workaround these issues, but in order to do so we are inserting quite a lot of changes, which I would want to understand if all are needed as a temporary fix.

To begin - I would separate the two commits to two separate PR's so that we can review and have a future reference to each one individually. 

Can the SCC/RBAC refactor commit be separate to another PR? Is it needed for this PR to work?

Is this approach of creating RBAC and SCC's by the operator reconcile recommended for openshift operators?

Can we suggest any PR to the postgres centos image repo instead to allow this workaround for anyone else who runs the same image - https://github.com/sclorg/postgresql-container ? I would much prefer to see the workaround as a PR to the image repo itself instead of making so many changes that might be needed by all other postgres users.

pkg/system/phase2_creating.go
```diff
@@ -194,6 +196,31 @@ func (r *Reconciler) SetDesiredServiceDBForPostgres() error {
...
 }
 
+func (r *Reconciler) setPostgresUserPass(c *corev1.Container) {
+	for j := range c.Env {
+		switch c.Env[j].Name {
+		case "POSTGRESQL_USER":
+			c.Env[j].ValueFrom = &corev1.EnvVarSource{
+				SecretKeyRef: &corev1.SecretKeySelector{
+					LocalObjectReference: corev1.LocalObjectReference{
+						Name: r.SecretDB.Name,
```
guymguym, line 206: Is there a reason we need to reconcile these values??
Why not just put those in the yaml itself like:
```yaml
env:
  - name: POSTGRESQL_USER
     valueFrom:
       secretKeyRef:
         name: noobaa-db
         key: user
  - name: POSTGRESQL_PASSWORD
     valueFrom:
       secretKeyRef:
         name: noobaa-db
         key: password
```

baum:
See [postgresql-container issue](https://github.com/sclorg/postgresql-container/issues/419)

pkg/system/phase2_creating.go
baum: @guymguym sounds good

dannyzaken: what about the upgrade, don't we need to add the new init container?

dannyzaken (approved):

PR #717, opened 2021-08-18
bump lib bucket provisioner version and add update callback in provisioner

romayalon:

### Explain the changes
1. Bumped lib bucket provisioner version to the latest which contains update event of OBCs (additional config)
2. Added update function to provisioner that handles bucket replication update.

1. 

### Testing Instructions:
1.


liranmauda (commented):
I think we should take out the replication handle from the update and take it into a function, then call it in the update. that way, when we need to handle more update scenarios (Quata?), it will be more readable.

pkg/obc/provisioner.go
```diff
@@ -203,6 +203,68 @@ func (p *Provisioner) Revoke(ob *nbv1.ObjectBucket) error {
...
+func (p *Provisioner) Update(ob *nbv1.ObjectBucket) error {
+	log := p.Logger
+	log.Infof("Update: got request to Update access to bucket %q", ob.Name)
+
+	r, err := NewBucketRequest(p, ob, nil)
+	if err != nil {
+		return err
+	}
+
+	newReplication := ob.Spec.Endpoint.AdditionalConfigData["replicationPolicy"]
```
liranmauda, line 216: Where do we initialize this map?

romayalon:
> I think we should take out the replication handle from the update and take it into a function, then call it in the update. that way, when we need to handle more update scenarios (Quata?), it will be more readable.

 thanks for the comment, it's one of the things I wanted to finalize.
I didn't finalize this PR yet, you can see I asked to now review it yet at the beginning of the description. will add WIP label.

liranmauda (approved):
LGTM

Issue #455, opened 2020-11-12
Feature Request: Allow custom pod templates


lallinger-arbeit:
As discussed yesterday in the community meeting we would like to set custom annotations to the noobaa-endpoint pods for our logging solution via the NooBaa CR. Also it would probably be best to allow customization of all the pod templates (core, db, endpoint) via the NooBaa CR.

snorwin:
+1

liranmauda:
Hi @lallinger-arbeit 
Can you please test If after PR #719 we can close this issue?

snorwin:
I did some test in our env and it works as expected with the latest operator version `master-20210912`.

lallinger-arbeit:
hi @liranmauda,
unfortunateley i current don't have any time to validate this and probably won't have until at least end of october..

liranmauda:
> hi @liranmauda,
> unfortunateley i currently don't have any time to validate this and probably won't have until at least end of october..

Sure, so as @snorwin tested it I will close this issue, and if we will need to add/change anything we can reopen (or open another issue).

PR #719, opened 2021-08-19
Custom annotations and labels for noobaa core and db

snorwin:
This PR tries to implement #455.

I tried to re-generate the CRD and deep copy, however I got the following error:
```
❯ make gen-api
go mod tidy
go mod vendor
✅ vendor
✅ gen
time -p build/_tools/operator-sdk-v0.17.2 generate k8s
INFO[0000] Running deepcopy code-generation for Custom Resource group versions: [noobaa:[v1alpha1], ] 
F0819 11:27:06.970359   30948 deepcopy.go:885] Hit an unsupported type invalid type for invalid type, from ./pkg/apis/noobaa/v1alpha1.AWSS3Spec
Command exited with non-zero status 255
real 0.94
user 0.03
sys 0.25
make: *** [Makefile:104: gen-api] Error 255
```


liranmauda:
Hi @snorwin 
Thanks for the PR 🙂 

In order to avoid the gen-api failure please do the following:
1. `. ./devenv.sh`
2. run `make gen`
3. run `make gen-api`
4. commit the files.

snorwin:
I was also thinking about refactoring the spec by creating a sub-spec for db related configurations instead of adding `db` as a prefix to the JSON tag , e.g.:
```
image: noobaa/noobaa-core:5.7.0
resources:
  limits:
    cpu: '1'
    memory: 4Gi
  requests:
    cpu: '1'
    memory: 4Gi

db:
  type: mongodb
  image: mongodb-36-centos7:1
  resources: 
    limits:
      cpu: '1'
      memory: 4Gi
    requests:
      cpu: '1'
      memory: 4Gi
```

However, with this approach you would loose backwards compatibility and should maybe considered for the next API version.

liranmauda:
Hi @snorwin 
I have missed your replay 🙂 

> However, with this approach you would loose backwards compatibility and should maybe considered for the next API version.

you are correct, We cannot lose backward compatibility, for now, you can open an issue if you think and we should look at it.

snorwin:
@liranmauda, no worries. However do you use a script to update `pkg/bundle/deploy.go`? It seems that running `make gen-api` isn't enough.

liranmauda:
> @liranmauda, no worries. However do you use a script to update `pkg/bundle/deploy.go`? It seems that running `make gen-api` isn't enough.

No, All we do is:
> 1. `. ./devenv.sh`
> 2. run `make gen`
> 3. run `make gen-api`
> 4. commit the files.

It should update, but you need to commit the changes.

snorwin:
@liranmauda I rebased the PR and re-run `make gen-api`. Can you please check again?

liranmauda:
Hi @snorwin 
The latest PR was introducing conflict with this one. 
It also adding Labels...

Can you please resolve the conflicts?

snorwin:
@liranmauda why are those labels just called `Labels` when only used for service monitor? Should I rename those to `smLabels` or `serviceMonitorLabels` ?

snorwin:
or should I follow your concept and do it like that:
```
.Spec.Labels["monitoring"] --> added to service monitor
.Spec.Labels["db"] --> added to DB pods
.Spec.Labels["core"] --> added to core
```

liranmauda:
> or should I follow your concept and do it like that:
> 
> ```
> .Spec.Labels["monitoring"] --> added to service monitor
> .Spec.Labels["db"] --> added to DB pods
> .Spec.Labels["core"] --> added to core pods
> ```

Sound Great.
On a side note, we are following the ocs-operator and rook concept 🙂

snorwin:
@liranmauda I rebased and refactored the PR.

liranmauda (approved):
LGTM, 
@dannyzaken @jeniawhite WDYT?

pkg/system/phase2_creating.go
```diff
@@ -408,6 +420,12 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 func (r *Reconciler) SetDesiredCoreApp() error {
 	r.CoreApp.Spec.Template.Labels["noobaa-core"] = r.Request.Name
 	r.CoreApp.Spec.Template.Labels["noobaa-mgmt"] = r.Request.Name
+	if coreLabels, ok := r.NooBaa.Spec.Labels["core"]; ok {
```
jeniawhite, line 423: I understand that the intention is to take the labels and annotations from the CR spec.
But do we actually want to override the previously assigned labels like system name (two lines above), isn't our intention to merge?
If that is not the case then I do not understand why would we assign labels and then override them completely in the next line.
This would break any pre-existing code that looked up for `noobaa-core` and `noobaa-mgmt` labels.

snorwin: good catch, sorry I messed up the order during all the rebase.

jeniawhite (approved):

PR #735, opened 2021-09-01
Add endpoint memory scaling metric

baum:
See [noobaa-core/issues/6709](https://github.com/noobaa/noobaa-core/issues/6709)

The endpoint deployment is being scaled by  [HPA](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/), see the [HPA declaration](https://github.com/noobaa/noobaa-operator/blob/master/deploy/internal/hpa-endpoint.yaml) currently only the CPU metrics is being used . To allow proper scaling for memory-bound workloads, avoiding endpoint pod's memory starvation add memory metrics.


kfir-payne (approved):

nimrod-becker (changes requested):
We need to discuss this, memory only won't help us. In fact, it will hurt.
We need to measure the pressure on our buffer semaphores (the waiting queue)

PR #742, opened 2021-10-04
Add tainted node detection to HA controller

baum:
Draft fix for [6695](https://github.com/noobaa/noobaa-core/issues/6695).

Extend HA controller to evict NooBaa pods from tainted nodes.


dannyzaken:
I discussed this PR with Alex. we agreed that a node being tainted should not be handled by the HA controller.
since issue #6695 is already closed, I am closing this PR

PR #743, opened 2021-10-06
Refactor SCC/RBAC

baum:
1. Noobaa operator no longer uses 'noobaa' SCC and relies on RBAC only.
2. Handle SCC RBAC creation for DB and Endpoint inside operator's
   reconcile loop.
3. OCS operator interface:
   a. OCS operator no longer needs to create 'noobaa' & 'noobaa-endpoint'
      SCCs. newNooBaaSCC() and newNooBaaEndpointSCC() functions should be
      removed from the OCS codebase.
   b. Use ''noobaa olm csv' to pass operator's RBAC resources to OCS.
4. Create a rbac/scc for the db.

### DIFF
--- a/pkg/olm/olm.go
+++ b/pkg/olm/olm.go
@@ -242,11 +242,6 @@ func GenerateCSV(opConf *operator.Conf, csvParams *generateCSVParams) *operv1.Cl
 		})
 	csv.Spec.InstallStrategy.StrategySpec.Permissions = append(csv.Spec.InstallStrategy.StrategySpec.Permissions,
 		operv1.StrategyDeploymentPermissions{
-			ServiceAccountName: opConf.SAEndpoint.Name,
-			Rules:              opConf.RoleEndpoint.Rules,
-		})
-	csv.Spec.InstallStrategy.StrategySpec.Permissions = append(csv.Spec.InstallStrategy.StrategySpec.Permissions,
-		operv1.StrategyDeploymentPermissions{
 			ServiceAccountName: opConf.SAUI.Name,
 			Rules:              opConf.RoleUI.Rules,
 		})
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -74,3 +74,9 @@ rules:
       - get
       - update
       - list
+  - apiGroups:
+      - security.openshift.io
+    resources:
+      - securitycontextconstraints
+    verbs:
+      - '*'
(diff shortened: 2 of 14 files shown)


jackyalbo (approved):

pkg/olm/olm.go
```diff
@@ -238,11 +238,6 @@ func GenerateCSV(opConf *operator.Conf, csvParams *generateCSVParams) *operv1.Cl
 			ServiceAccountName: opConf.SA.Name,
 			Rules:              opConf.Role.Rules,
 		})
-	csv.Spec.InstallStrategy.StrategySpec.Permissions = append(csv.Spec.InstallStrategy.StrategySpec.Permissions,
```
dannyzaken, line 241: If I understand correctly, this section of the CSV allows us to create the SA and roles using the OLM, is that true? if so I think it's a better mechanism than reconciling ourselves in the operator

dannyzaken (changes requested):
@baum. let's discuss it before we merge

pkg/olm/olm.go
baum: Before this change, the noobaa operator had no permissions to manipulate RBAC resources, so the external installer (CLI or uplevel operator) had to create RBAC resources for NooBaa. However, after this change, the noobaa operator is granted  RBAC manipulation privileges, rendering those as NooBaa internal implementation details, that the external installer should not be aware of. IMHO minimal interface with OLM is a positive. WDYT?

dannyzaken: the permissions to manipulate RBAC are only needed to reconcile the SA\role\binding, isn't it? if we move it to the CSV then we don't need these permissions in the operator. I think that in general, it is not so much the operator's responsibility to handle RBAC resources and if the SCCs could have been handled by the OLM we would have preferred it. 

these are the pros\cons that I can think of for keeping it in the CSV 
**Pros**
* noobaa-operator does not need to request permissions to manipulate RBAC, which is more secure
* We don't need the logic of reconciling RBAC resources on every noobaa CR reconciliation when this is only relevant once on installation\upgrade
* We get the update of the RBAC resources "for free" on upgrades. the OLM will just replace them for us

**Cons**
* We get less control on how these RBAC are applied to the cluster
* We need to create the RBAC resources when installing from CLI

dannyzaken (commented):
@baum, so `noobaa-core` pod is still using the `noobaa` service account, together with all the new permissions, right? 
Maybe worth separating it too?

PR #753, opened 2021-10-12
Reconcile db config maps

igorpick:
In upgrade use case, If db config maps were modified in new release, replace existing.
If user manually changed the config maps, replace the changes.

In both scenarios, if the config maps were replaced and postgres db was not restarted,  restart postgres db pods.
fix for https://bugzilla.redhat.com/show_bug.cgi?id=2012930


dannyzaken:
@igorpick you can merge once the tests pass. let's add more specific tests later

PR #774, opened 2021-10-26
Extend support bucket qouta

igorpick:
Added support for max objects quota.

Provide an option to Configure Quota per Bucket on an OBC level and on the BucketClass level. If both exist, the OBC definition would override the BucketClass definition.
Set a quota of object bucket claims, which limits the max number of objects per bucket (MaxObjects).
Set a quota of object bucket claims, which limits the max total size per bucket (MaxSize).
Update BucketClass quota, will be effective in new buckets only using this BucketClass
Add to NooBaa CLI new options for creating object bucket claims while providing one of the 2 types of quota

related old PR https://github.com/noobaa/noobaa-operator/pull/691


pkg/util/util.go
```diff
@@ -1789,3 +1795,32 @@ func LoadBucketReplicationJSON(replicationJSONFilePath string) (string, error) {
 
 	return string(bytes), nil
 }
+
+// ValidateQuotaConfig maxSize and maxObjects value of obc or bucketclass
+func ValidateQuotaConfig(name string, maxSize string, maxObjects string) error {
+
+	// Pisitive number
```
romayalon, line 1802:
```suggestion
	// Positive number
```

romayalon (approved):

dannyzaken (approved):

PR #790, opened 2021-11-08
Increase CLI test timeout/Restore core image version

baum:
1. Increase CLI test timeout
2. Restore core image version

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -30,7 +30,7 @@ const (
 	// ContainerImageRepo is the repo of the default image url
 	ContainerImageRepo = "noobaa-core"
 	// ContainerImageTag is the tag of the default image url
-	ContainerImageTag = "5.9.0-20210722"
+	ContainerImageTag = "5.9.0-20211118"
 	// ContainerImageSemverLowerBound is the lower bound for supported image versions
 	ContainerImageSemverLowerBound = "5.0.0"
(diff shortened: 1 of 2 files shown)


liranmauda (commented):
Why do we need 10 min timeout (instead of 3 which is big on its own)?

dannyzaken:
@liranmauda I agree that this is a long timeout. for some reason, with the newer images of noobaa-core 5.9 the tests fail on timeout. 
we still need to find the root cause but for now, we just want to use the newer image of 5.9 without failing the operator tests.
I will open an issue on that

dannyzaken (approved):

pkg/options/options.go
```diff
@@ -30,7 +30,7 @@ const (
 	// ContainerImageRepo is the repo of the default image url
 	ContainerImageRepo = "noobaa-core"
 	// ContainerImageTag is the tag of the default image url
-	ContainerImageTag = "5.9.0-20210722"
+	ContainerImageTag = "5.9.0-20211118"
```
dannyzaken, line 33: @liranmauda why is this image tag in the future? 🤔

liranmauda: Being a good developer means using future, unwritten, undeveloped builds 😂 

Anyway, it was supposed to be 20211018, you can change that to the latest...

dannyzaken:
I updated the core image in this PR https://github.com/noobaa/noobaa-operator/pull/824

PR #794, opened 2021-11-10
Adding `fsGroupChangePolicy` and `seLinuxOptions` to `securityContext`

liranmauda:
### Explain the changes
- Adding `fsGroupChangePolicy` and `seLinuxOptions` to `securityContext`

according to https://hackmd.io/7heLp_noQmqU_Ef7VaiCKg when we have a large number of files and we want to mount the volume we can get the `Error: context deadline exceeded`.

there are 2 root causes (which are detailed in the link above), and this PR tries to handle them both.

1. Fixes: https://github.com/noobaa/noobaa-core/issues/6708
2. Fixes: https://github.com/noobaa/noobaa-core/issues/6786
3. Fixes: https://github.com/noobaa/noobaa-core/issues/6707

### Testing Instructions:
1. Attach a directory with a large number of files to a bucket (or write a lot of files into an NSFS bucket).
2. Restart the endpoint
3. See that there are the endpoint is not in `CreateContainerError` and there are no `Error: context deadline exceeded`

### DIFF
--- a/deploy/scc_endpoint.yaml
+++ b/deploy/scc_endpoint.yaml
@@ -24,7 +24,7 @@ requiredDropCapabilities:
 runAsUser:
   type: RunAsAny
 seLinuxContext:
-  type: MustRunAs
+  type: RunAsAny
 supplementalGroups:
   type: RunAsAny
 users:
--- a/deploy/internal/deployment-endpoint.yaml
+++ b/deploy/internal/deployment-endpoint.yaml
@@ -46,6 +46,9 @@ spec:
               cpu: "1"
               memory: "2Gi"
           securityContext:
+            fsGroupChangePolicy: "OnRootMismatch"
+            seLinuxOptions:
+              type: "spc_t"
             capabilities:
               add: ["SETUID", "SETGID"]
           ports:
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -2890,7 +2890,7 @@ data:
           su postgres -c "bash -x /usr/bin/run-postgresql"
 `
 
-const Sha256_deploy_internal_deployment_endpoint_yaml = "d098321ca3038440e35b27277c1e38eedea852a5be25371e29a442d5f7987590"
+const Sha256_deploy_internal_deployment_endpoint_yaml = "26b0dc615d97e493cac78d7050e3156d01a1033e3b7b6397fd037a6c74971c0b"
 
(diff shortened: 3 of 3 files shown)


jackyalbo:
Did you check in Openshift? Maybe you need to update the SCC as well...

liranmauda:
@haircommander WDYT?

haircommander:
> Did you check in Openshift? Maybe you need to update the SCC as well...

yes, it's possible unless RunAsAny is already specified for SELinuxContext

the spec changes LGTM

liranmauda:
> > Did you check in Openshift? Maybe you need to update the SCC as well...
> 
> yes, it's possible unless RunAsAny is already specified for SELinuxContext
> 
> the spec changes LGTM

`MustRunAs` is specified for SELinuxContext, so I guess we are good there. 

https://github.com/noobaa/noobaa-operator/blob/de4dc9aff384cd7c6f35e2b56aa72eac75b977b4/deploy/scc_endpoint.yaml#L24-L29

liranmauda:
Doing this change will immediately lead to `context deadline exceeded`:
```
time="2021-11-11T11:34:59Z" level=info msg="SetPhase: \"Creating\"" sys=openshift-storage/noobaa
time="2021-11-11T11:35:29Z" level=fatal msg="Manager exited non-zero: [failed to wait for noobaa-controller caches to sync: no matches for kind \"NooBaaAccount\" in version \"noobaa.io/v1alpha1\", failed waiting for all runnables to end within grace period of 30s: context deadline exceeded]"
```
@haircommander any idea

haircommander:
> MustRunAs is specified for SELinuxContext, so I guess we are good there.

I think it needs to be RunAsAny, unless the value for MustRunAs is spc_t (which it doesn't look like). I am guessing MustRunAs with an empty SELinuxContext means the user is not allowed to specify any SELinuxContext, which likely is causing this error

liranmauda:
> RunAsAny

Yes, It worked. 
Tested with ODF 4.9 on OCP 4.9

deploy/scc_endpoint.yaml
```diff
@@ -24,7 +24,7 @@ requiredDropCapabilities:
 runAsUser:
   type: RunAsAny
 seLinuxContext:
-  type: MustRunAs
+  type: RunAsAny
```
dannyzaken, line 27: this should be changed in ocs-operator too

liranmauda: Sure, Once it is in.

dannyzaken (approved):

PR #803, opened 2021-11-22
Removing "allowed buckets" functionality

BarakSason:
The "allowed buckets" functionality interferes with
another access control mechanism which was implemented
and thus needs to be removed

Signed-off-by: Barak Sason Rofman sason922@gmail.com


BarakSason:
@guymguym This currently seems to be failing on some backingstore-related issue - https://github.com/noobaa/noobaa-operator/runs/4314618960?check_suite_focus=true#step:6:1101
The only backingstore-related related changes I see are the ones we did in order to solve the compilation issues which existed in the auto-generated file, could you please take a look and see if those might be the case?
@dannyzaken Maybe you could assist here as well?

guymguym:
@BarakSason Did you try to run the test locally and inspect the problem?

guymguym:
Try running the test with `make test-cli-flow`

igorpick (approved):

nimrod-becker:
Closing, will be superseded with a new PR

PR #804, opened 2021-11-24
Add Regenerate S3 Credentials to the CLI

liranmauda:
### Explain the changes
Add Regenerate S3 Credentials to the CLI

Handling 3 cases:
1. An account with no CRD and no secret
2. An account with no CRD but with secret
3. An account with CRD and with a secret

### Testing Instructions:
1.  regenerate the s3 keys see that they changed and try to use them

### DIFF
--- a/pkg/noobaaaccount/noobaaaccount.go
+++ b/pkg/noobaaaccount/noobaaaccount.go
@@ -6,7 +6,9 @@ import (
 
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
+	"github.com/noobaa/noobaa-operator/v5/pkg/nb"
 	"github.com/noobaa/noobaa-operator/v5/pkg/options"
+	"github.com/noobaa/noobaa-operator/v5/pkg/system"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
 
 	"github.com/spf13/cobra"
@@ -29,6 +31,7 @@ func Cmd() *cobra.Command {
 	}
 	cmd.AddCommand(
 		CmdCreate(),
+		CmdRegenerate(),
 		CmdDelete(),
 		CmdStatus(),
 		CmdList(),
@@ -51,6 +54,16 @@ func CmdCreate() *cobra.Command {
 	cmd.Flags().StringSlice("allowed_buckets", nil,
 		"Set the user allowed buckets list (use commas or multiple flags)")
 	cmd.Flags().String("default_resource", "", "Set the default resource, on which new buckets will be created")
+	return cmd
+}
+
+// CmdRegenerate returns a CLI command
+func CmdRegenerate() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "regenerate <noobaa-account-name>",
+		Short: "Regenerate S3 Credentials",
+		Run:   RunRegenerate,
+	}
 	return cmd
 }
 
@@ -175,6 +188,52 @@ func RunCreate(cmd *cobra.Command, args []string) {
 	}
 }
 
+// RunRegenerate runs a CLI command
+func RunRegenerate(cmd *cobra.Command, args []string) {
+	log := util.Logger()
+
+	if len(args) != 1 || args[0] == "" {
+		log.Fatalf(`❌ Missing expected arguments: <noobaa-account-name> %s`, cmd.UsageString())
+	}
+
+	var decision string
+	log.Printf("You are about to regenerate an account's security credentials.")
+	log.Printf("This will invalidate all connections between S3 clients and NooBaa which are connected using the current credentials.")
+	log.Printf("are you sure? y/n")
+	
+	for {
+		fmt.Scanln(&decision)
+		if decision == "y" {
+			break
+		} else if decision == "n" {
+			return 
+		}
+	}
+
+	name := args[0]
+
+	o := util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_noobaaaccount_cr_yaml)
+	noobaaAccount := o.(*nbv1.NooBaaAccount)
+	noobaaAccount.Name = name
+	noobaaAccount.Namespace = options.Namespace
+
+	if !util.KubeCheck(noobaaAccount) && (name != "admin@noobaa.io") {
+		err := GenerateNonCrdAccountKeys(name)
+		if err != nil {
+			log.Fatalf(`❌ Could not regenerate credentials for %q: %v`,name, err)
+		}
+	} else {
+		err := GenerateAccountKeys(name)
+		if err != nil {
+			log.Fatalf(`❌ Could not regenerate credentials for %q: %v`,name, err)
+		}
+	
+		RunStatus(cmd, args)
+	}
+
+	
+}
+
 // RunDelete runs a CLI command
 func RunDelete(cmd *cobra.Command, args []string) {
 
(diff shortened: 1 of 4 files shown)


pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -175,6 +188,39 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
+
+	if len(args) != 1 || args[0] == "" {
+		log.Fatalf(`❌ Missing expected arguments: <noobaa-account-name> %s`, cmd.UsageString())
+	}
+
+	var decision string
+	log.Printf("You are about to regenerate an account's security credentials.")
+	log.Printf("This will invalidate all connections between S3 clients and NooBaa which are connected using the current credentials.")
+	log.Printf("are you sure? y/n")
+	fmt.Scanln(&decision)
```
jackyalbo, line 203: Shouldn't the scan be inside the for?

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -175,6 +188,39 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
+		} else if decision == "n" {
+			return 
+		}
+	}
+
+	name := args[0]
+
+	err := GenerateAccountKeys(name)
+	if err != nil {
+		log.Fatalf(`❌ Could not regenerate credentials for %q`,name)
```
jackyalbo, line 217: maybe we should print the error. for easier debugging

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -360,3 +406,44 @@ func RunReconcile(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		return err
+	}
+ 
+	accessKeys = accountInfo.AccessKeys[0]
+
+	secret := util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret)
+	secret.Namespace = options.Namespace
+	secret.Name = fmt.Sprintf("noobaa-account-%s", name)
+	util.KubeCheckQuiet(secret)
```
jackyalbo, line 440: handle the case the secret was not found

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -360,3 +406,44 @@ func RunReconcile(cmd *cobra.Command, args []string) {
...
+
+	secret.StringData = map[string]string{}
+	secret.StringData["AWS_ACCESS_KEY_ID"] = accessKeys.AccessKey
+	secret.StringData["AWS_SECRET_ACCESS_KEY"] = accessKeys.SecretKey
+	util.KubeUpdate(secret)
```
jackyalbo, line 445: also better check update succeeded, and return error if not

jackyalbo (changes requested):

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -175,6 +188,39 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
+	for {
+		fmt.Scanln(&decision)
+		if decision == "y" {
+			break
+		} else if decision == "n" {
+			return 
+		}
+	}
+
+	name := args[0]
```
jackyalbo, line 213: Missed this on the previous review but I think we should check that this account exists, especially when first thing you do is to check if the secret exists and then the user will get the wrong error - no secret instead of no account with this name. also will probably do the check before of asking the "are you sure". something like this:
```suggestion
        o := util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_noobaaaccount_cr_yaml)
        noobaaAccount := o.(*nbv1.NooBaaAccount)
	noobaaAccount.Name = args[0]
	noobaaAccount.Namespace = options.Namespace

	if !util.KubeCheck(noobaaAccount) {
		log.Fatalf(`❌ Could not get NooBaaAccount %q in namespace %q`,
		noobaaAccount.Name, noobaaAccount.Namespace)
	}
```

liranmauda: I am not sure, as the account can exist but not an object, for example, admin@noobaa.io account has no `noobaaaccount` object.
With that said, if there is no such object then there will be no secret which means this will not work on defaults accounts. 
@nimrod-becker Do we want it enabled on default accounts?

nimrod-becker: Why? We want to be able to change for accounts which were created via UI/RPC and not only CRD, for those there would be no object

jackyalbo: ok, so first we need to handle that - cause no secret will be found (unless it's admin and then we need to do some special handling). and second we can read the account first using the API and then warn him in case the user doesn't exist - before doing anything else.

liranmauda: Ok, added the regenerate to none CRD based accounts.

We will check if there is a CRD then we will go to the proper flow.

a gap remains in other flows (list, delete) that we are handling only CRD based accounts.

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -203,35 +262,44 @@ func RunStatus(cmd *cobra.Command, args []string) {
...
 
-	noobaaAccount.Name = args[0]
-	secret.Name = fmt.Sprintf("noobaa-account-%s", args[0])
+	noobaaAccount.Name = name
+	secret.Name = fmt.Sprintf("noobaa-account-%s", name)
 	noobaaAccount.Namespace = options.Namespace
 	secret.Namespace = options.Namespace
 
-	if !util.KubeCheck(noobaaAccount) {
+	if !util.KubeCheck(noobaaAccount) && (name != "admin@noobaa.io") {
```
jackyalbo, line 276: I don't think we need to change the status - this is for CRD only accounts

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -175,6 +188,52 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
+
+	name := args[0]
+
+	o := util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_noobaaaccount_cr_yaml)
+	noobaaAccount := o.(*nbv1.NooBaaAccount)
+	noobaaAccount.Name = name
+	noobaaAccount.Namespace = options.Namespace
+
+	if !util.KubeCheck(noobaaAccount) && (name != "admin@noobaa.io") {
+		err := GenerateNonCrdAccountKeys(name)
```
jackyalbo, line 221: Not sure why to split to two functions? why not have all this logic in the function and to avoid the code duplication?

jackyalbo (changes requested):

pkg/noobaaaccount/noobaaaccount.go
liranmauda: The status prints the keys from the secret and we would like to print it also for the admin account.
In the admin account, the secret name is different from the email chance we need to change the name to and get the proper secret.
also in the admin account, there is no CRD so we don't want to check it.

pkg/noobaaaccount/noobaaaccount.go
liranmauda: Those are 2 different flows. 
and for readability.

jackyalbo (approved):

pkg/noobaaaccount/noobaaaccount.go
jackyalbo: I have to say that I don't this special handling - but after I understand that's it's a must - go for it.

PR #805, opened 2021-11-25
set EnableHTTPSTrafficOnly on create storage account

dannyzaken:
### Explain the changes
* when creating a storage account in azure utils - setting EnableHTTPSTrafficOnly param to true.
* This is required for creating a storage account for resource groups that enforce secure transfer for storage account

1. fixes https://bugzilla.redhat.com/show_bug.cgi?id=1970123

### Testing Instructions:
1.

### DIFF
--- a/pkg/system/azure_utils.go
+++ b/pkg/system/azure_utils.go
@@ -54,6 +54,7 @@ func (r *Reconciler) CreateStorageAccount(accountName, accountGroupName string) 
 			accountName, err, *result.Message)
 	}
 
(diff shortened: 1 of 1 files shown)


nimrod-becker (approved):

dannyzaken:
I tested it on azure with a policy set on the resource group to only allow storage account with secure transport. 
see [here](https://docs.microsoft.com/en-us/azure/storage/common/policy-reference) under "Secure transfer to storage accounts should be enabled"

PR #806, opened 2021-11-25
Add admission validation webhook server

kfir-payne:

### Explain the changes
1. add admission webhook server.
2. add validation of CREATE/UPDATE/DELETE operations for BS/NS resources.

1. 

### Testing Instructions:
1. `ginkgo -v pkg/admission/test/integ`
2. `ginkgo -v pkg/admission/test/unit`

### DIFF
--- a/pkg/admission/test/integ/admission_integ_test.go
+++ b/pkg/admission/test/integ/admission_integ_test.go
@@ -1,0 +1,293 @@
+package admissionintegtests
+
+import (
+	"context"
+
+	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	"github.com/noobaa/noobaa-operator/v5/pkg/backingstore"
+	"github.com/noobaa/noobaa-operator/v5/pkg/bucket"
+	"github.com/noobaa/noobaa-operator/v5/pkg/bucketclass"
+	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
+	"github.com/noobaa/noobaa-operator/v5/pkg/namespacestore"
+	"github.com/noobaa/noobaa-operator/v5/pkg/operator"
+	"github.com/noobaa/noobaa-operator/v5/pkg/options"
+	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+const (
+	namespace = "test"
+)
+
+var _ = Describe("Admission validate setup", func() {
+	Context("Validate admission resources", func() {
+		It("Should Exist", func() {
+			options.Namespace = namespace
+			c := operator.LoadOperatorConf(&cobra.Command{})
+			operator.LoadAdmissionConf(c)
+
+			Expect(util.KubeCheck(c.WebhookConfiguration)).To(BeTrue())
+			Expect(util.KubeCheck(c.WebhookSecret)).To(BeTrue())
+			Expect(util.KubeCheck(c.WebhookService)).To(BeTrue())
+		})
+	})
+})
+
+var _ = Describe("Admission server integration tests", func() {
+
+	var (
+		testBackingstore   *nbv1.BackingStore
+		testNamespacestore *nbv1.NamespaceStore
+		result             bool
+		err                error
+	)
+
+	Setup()
+
+	Describe("Create operations", func() {
+		BeforeEach(func() {
+			testBackingstore = util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_backingstore_cr_yaml).(*nbv1.BackingStore)
+			testBackingstore.Name = "bs-name"
+			testBackingstore.Namespace = namespace
+			testNamespacestore = util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_namespacestore_cr_yaml).(*nbv1.NamespaceStore)
+			testNamespacestore.Name = "ns-name"
+			testNamespacestore.Namespace = namespace
+		})
+		Context("Empty secret name", func() {
+			It("Should Deny", func() {
+				testBackingstore.Spec = nbv1.BackingStoreSpec{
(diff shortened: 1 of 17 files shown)


pkg/admission/validate_backingstore.go
```diff
@@ -0,0 +1,317 @@
...
+		return false, "Failed to identify Backingstore type"
+	}
+	return true, "allowed"
+}
+
+// ValidateBackingstoreDeletion validates the deleted backingstore not containing data buckets
+func (bsv *BackingStoreValidator) ValidateBackingstoreDeletion() (bool, string) {
+	systemInfo, err := bsv.Client.NBClient.ReadSystemAPI()
+	if err != nil {
+		log.Error("Failed to call ReadSystemInfo API")
```
dannyzaken, line 304: return on errors

pkg/admission/test/integ/admission_integ_test.go
```diff
@@ -0,0 +1,286 @@
...
+							Name:      "",
+							Namespace: namespace,
+						},
+					},
+				}
+
+				result, err = KubeCreate(testNamespacestore)
+				Expect(result).To(BeFalse())
+				Ω(err).Should(HaveOccurred())
+			})
```
romayalon, line 94: validate error message too

pkg/admission/validate_backingstore.go
```diff
@@ -0,0 +1,317 @@
...
+// ValidateBackingstore call appropriate validations based on the operation
+func (bsv *BackingStoreValidator) ValidateBackingstore() admissionv1.AdmissionReview {
+	bsv.arResponse = &admissionv1.AdmissionReview{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "AdmissionReview",
+			APIVersion: "admission.k8s.io/v1",
+		},
+		Response: &admissionv1.AdmissionResponse{
+			UID:     bsv.arRequest.Request.UID,
+			Allowed: false,
```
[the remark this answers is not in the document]
kfir-payne: in any case (CREATE/UPDATE/DELETE) these fields will be updated, so it doesn't really matter.
anyway I switched to `true`, to stay on the permissive guideline.

pkg/admission/validate_backingstore.go
```diff
@@ -0,0 +1,317 @@
...
+	bsv.BackingStore = &createdBs
+
+	allowed, message = bsv.ValidateBackingStoreType()
+	if !allowed {
+		bsv.arResponse.Response.Allowed = allowed
+		bsv.arResponse.Response.Result.Message = message
+		return
+	}
+
+	allowed, message = bsv.ValidateBSEmptySecretName()
```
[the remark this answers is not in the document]
kfir-payne: Agreed, will be much easier to implement when https://github.com/noobaa/noobaa-operator/pull/688 merged.

pkg/operator/manager.go
```diff
@@ -75,6 +77,14 @@ func RunOperator(cmd *cobra.Command, args []string) {
 	// 	log.Warnf("Failed ExposeMetricsPort: %s", err)
 	// }
 
+	_, ok := os.LookupEnv("ENABLE_NOOBAA_ADMISSION")
+	if ok {
```
dannyzaken, line 81: this will only check if the env exists. you also need to check its value

```suggestion
	enable_addmission, ok := os.LookupEnv("ENABLE_NOOBAA_ADMISSION")
	if ok && enable_admission == "true" {
```

pkg/admission/validate_backingstore.go
```diff
@@ -0,0 +1,321 @@
...
+	if err := json.Unmarshal(raw, &BS); err != nil {
+		bsv.Logger.Error("error deserializing backingstore")
+	}
+	return &BS
+}
+
+// DeserializeOldBS extract the old backingstore object from the request, exist only on UPDATE/DELETE requests
+func (bsv *BackingStoreValidator) DeserializeOldBS() *nbv1.BackingStore {
+	raw := bsv.arRequest.Request.OldObject.Raw
+	oldBS := nbv1.BackingStore{}
```
romayalon, line 88: you don't need here 2 functions DeserializeOldBS and DeserializeBS, just keep the DeserializeBS but send it the old/new backingstore as param

pkg/admission/validate_backingstore.go
```diff
@@ -0,0 +1,321 @@
...
+func (bsv *BackingStoreValidator) ValidateMaxVolumeCount() (bool, string) {
+	if bsv.BackingStore.Spec.PVPool.NumVolumes > MaximumVolumeCount {
+		return false, "Unsupported volume count, the maximum supported volume count is 20"
+	}
+	return true, "allowed"
+}
+
+// ValidateMinVolumeCount validates pvpool backingstore volume count is above 0
+func (bsv *BackingStoreValidator) ValidateMinVolumeCount() (bool, string) {
+	if bsv.BackingStore.Spec.PVPool.NumVolumes < 0 {
```
romayalon, line 255: you allow here the volume count to be 0

pkg/admission/validate_backingstore.go
```diff
@@ -0,0 +1,321 @@
...
+	if err := json.Unmarshal(raw, &oldBS); err != nil {
+		bsv.Logger.Error("error deserializing old backingstore")
+	}
+	return &oldBS
+}
+
+// ValidateCreate runs all the validations tests for CREATE operations
+func (bsv *BackingStoreValidator) ValidateCreate() {
+	bsv.BackingStore = bsv.DeserializeBS()
+	if !bsv.RunValidation(bsv.ValidateBackingStoreType()) {
```
romayalon, line 98: I feel like this kind of writing is less understood to the reader than for example:
if ok := ValidateBackingStoreType(bs); !ok {
RunValidation(bs) // this function name can be more acurate too...
return
}

WDYT?

kfir-payne: I think that if we change the name of `RunValidation` we could skip the `ok` var and do something like this:
```
if !bsv.GetValidationResult(bsv.ValidateBSEmptySecretName()) {
return
}
```

romayalon: hmm I find it confusing to call it getValidationResult when the other one actually gets the result
let's discuss it on a short call :)

pkg/util/util.go
```diff
@@ -276,6 +281,11 @@ func kubeCreateSkipOrFailExisting(obj client.Object, skipOrFail bool) bool {
 		log.Printf("❌ Invalid: %s %q: %s\n", gvk.Kind, objKey.Name, err)
 		return false
 	}
+	statusErr, ok := err.(*errors.StatusError)
```
romayalon, line 284: this is a general function and you print always a very specific Admission Deny prefix to the error, is that always the case?

romayalon (approved):

PR #816, opened 2021-12-16
Fix password logged in clear text

kfir-payne:

### Explain the changes
1. Add a custom `String()` for the `CreateSystemParams` struct to gain control on the printed values.

1. 

### Testing Instructions:
1.

### DIFF
--- a/pkg/nb/types.go
+++ b/pkg/nb/types.go
@@ -9,7 +9,15 @@ import (
 
 const (
 	petaInBytes = 1024 * 1024 * 1024 * 1024 * 1024
+	maskString  = "****"
 )
+
+// MaskedString is a string type for sensitive string, masked when formated
+type MaskedString string
+
+func (MaskedString) String() string {
+	return maskString
+}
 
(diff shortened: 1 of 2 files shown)


pkg/nb/types.go
```diff
@@ -718,3 +720,7 @@ func UInt64ToBigInt(value uint64) BigInt {
 		N:    int64(value % petaInBytes),
 	}
 }
+
+func (params CreateSystemParams) String() string {
```
dannyzaken, line 724: as discussed, to make it more reusabel let's implement a Stringer for a more basic string type

pkg/nb/types.go
```diff
@@ -9,8 +9,16 @@ import (
 
 const (
 	petaInBytes = 1024 * 1024 * 1024 * 1024 * 1024
+	maskString  = "************************"
```
dannyzaken, line 12: Maybe ~4 stars is sufficient :) will make logs more readable

dannyzaken (approved):

PR #818, opened 2021-12-17
Upgrade MongoDB to v3.6.23 in order to prevent segmentation fault

### LINKED ISSUE #817: noobaa-db-0 crash looping using MongoDB (v3.6.3) - Got signal: 11 (Segmentation fault).
### Environment info
- NooBaa Version:  noobaa/noobaa-core:master-20211205
- Platform: **Kubernetes v1.21.1+6438632 | OpenShift 4.8.17**

### Actual behavior
The `noobaa-db-0` (using the image `centos/mongodb-36-centos7:1`) ended up randomly in a crash looping state caused by seg faults.  As a result NooBaa itself wasn't working anymore. In order to fix the issue we upgraded the MongoDB to the latest 3.6.x version, which resolved the issue immediately.
```
2021-12-17T12:19:31.009+0000 I COMMAND  [conn21] command nbcore.system.js command: find { find: "system.js", $readPreference: { mode: "secondaryPreferred" }, $db: "nbcore" } planSummary: EOF keysExamined:0 docsExamined:0 cursorExhausted:1 numYields:0 nreturned:0 reslen:89 locks:{ Global: { acquireCount: { r: 6 } }, Database: { acquireCount: { r: 2, R: 1 } }, Collection: { acquireCount: { r: 2 } } } protocol:op_msg 0ms
2021-12-17T12:19:31.090+0000 F -        [conn21] Invalid access at address: 0
2021-12-17T12:19:31.090+0000 F -        [conn21] Got signal: 11 (Segmentation fault).
```

### Steps to reproduce
_(Unable to reproduce it.)_

snorwin:
Signed-off-by: Norwin Schnyder <norwin.schnyder+github@gmail.com>

### Explain the changes
Upgraded MongoDB to v3.6.23

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ $ noobaa options
 
 The following options can be passed to any command:
 
-      --db-image='centos/mongodb-36-centos7': The database container image
+      --db-image='mongo:3.6.23': The database container image
       --db-storage-class='': The database volume storage class name
       --db-volume-size-gb=0: The database volume size in GB
       --image-pull-secret='': Image pull secret (must be in same namespace)
--- a/deploy/internal/statefulset-db.yaml
+++ b/deploy/internal/statefulset-db.yaml
@@ -48,7 +48,7 @@ spec:
         command:
         - bash
         - -c
-        - /opt/rh/rh-mongodb36/root/usr/bin/mongod --port 27017 --bind_ip_all --dbpath /data/mongo/cluster/shard1
+        - mongod --port 27017 --bind_ip_all --dbpath /data/mongo/cluster/shard1
         resources:
           requests:
             cpu: "2"
--- a/doc/noobaa-crd.md
+++ b/doc/noobaa-crd.md
@@ -165,7 +165,7 @@ metadata:
   namespace: noobaa
 spec:
   image: noobaa/noobaa-core:v9999.9.9
-  dbImage: centos/mongodb-36-centos7
+  dbImage: mongo:3.6.23
   imagePullSecret:
     name: <SECRET-NAME>
 ```
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -3658,7 +3658,7 @@ spec:
                   resource: limits.memory
 `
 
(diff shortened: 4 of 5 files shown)


liranmauda:
Hi @snorwin 
First of all, Thanks for the PR 🙂 

We have a flag `--db-image` in order to change the db image. is it possible that it is sufficient? 

also, as a best practice, as we support less and less in mongo, we encourage noobaa users to move to postgres.
Is there a particular reason to use mongo and not migrate to postgres? 

Regarding to the change, We need to see that changing the command is working on the downstream.

snorwin:
Hi @liranmauda 
Unfortunately just changing the DB image in the NooBaa CR or using a flag isn't sufficient. The path hard coded in deploy/internal/statefulset-db.yaml limits the number of images which can be used, the latest RedHat/CentOS based image is 7 months old and contains a ~3 years old MongoDB version (v3.6.3). I think that changing at least the command of the DB container as suggest in this PR is absolutely necessary. Furthermore, it is backwards compatible with the 'centos/mongodb-36-centos7:1' image.

Is there a documented way to migrate a noobaa-db from MongoDB to Postgres without loosing the buckets and their state/content?

liranmauda:
Hi @snorwin 
The reason we did not change the mongoDB image is that it is the last version with the old license. 

Regarding Downstream, when I think of it it should not matter as we dropped mongoDB support completely (DS) and we only using postgres in versions 5.7 and up.

Regarding postgres migration, it happens automatically when switching to postgres.
you can see the discussion in https://github.com/noobaa/noobaa-core/issues/5666

Tell me if it helps.

snorwin:
Thanks for the clarification, I will therefore test the migration from MongoDB to Postgres as well.

snorwin:
@liranmauda it seems that the migration isn't possible atm, I created an issue for it: #822

snorwin:
@liranmauda any updates on this?

nimrod-becker:
As liran stated, due to the license change of MongoDB we are not bumping it.
The project has moved to work with PSQL.

PR #821, opened 2021-12-21
Add Bucketclass admission validations

kfir-payne:

### Explain the changes
1. add bucketclass validation to the admission webhook server.
2. add unit and integration  tests for the bucketclass validations.
3. bug fix in namespacestore quota validation.

1. 

### Testing Instructions:
1. `ginkgo -v pkg/admission/test/integ`
2. `ginkgo -v pkg/admission/test/unit`

### DIFF
--- a/pkg/admission/validate_bucketclass.go
+++ b/pkg/admission/validate_bucketclass.go
@@ -1,0 +1,92 @@
+package admission
+
+import (
+	"encoding/json"
+
+	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	"github.com/noobaa/noobaa-operator/v5/pkg/bucketclass"
+	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+	"github.com/sirupsen/logrus"
+	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// BucketClassValidator is the context of a bucketclass validation
+type BucketClassValidator struct {
+	Logger     *logrus.Entry
+	arRequest  *admissionv1.AdmissionReview
+	arResponse *admissionv1.AdmissionReview
+}
+
+// NewBucketClassValidator initializes a BucketClassValidator to be used for loading and validating a bucketclass
+func NewBucketClassValidator(arRequest admissionv1.AdmissionReview) *BucketClassValidator {
+	bcv := &BucketClassValidator{
+		Logger:    logrus.WithField("admission bucketclass validation", arRequest.Request.Namespace),
+		arRequest: &arRequest,
+		arResponse: &admissionv1.AdmissionReview{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "AdmissionReview",
+				APIVersion: "admission.k8s.io/v1",
+			},
+			Response: &admissionv1.AdmissionResponse{
+				UID:     arRequest.Request.UID,
+				Allowed: true,
+				Result: &metav1.Status{
+					Message: "allowed",
+				},
+			},
+		},
+	}
+	return bcv
+}
+
+// ValidateBucketClass call appropriate validations based on the operation
+func (bcv *BucketClassValidator) ValidateBucketClass() admissionv1.AdmissionReview {
+	switch bcv.arRequest.Request.Operation {
+	case admissionv1.Create:
+		bcv.ValidateCreate()
+	default:
+		bcv.Logger.Error("Failed to identify bucketclass operation type")
+	}
+	return *bcv.arResponse
+}
+
+// SetValidationResult responsible of assinging the return values of a validation into the response appropriate fields
+func (bcv *BucketClassValidator) SetValidationResult(isAllowed bool, message string) {
+	bcv.arResponse.Response.Allowed = isAllowed
+	bcv.arResponse.Response.Result.Message = message
+}
+
+// DeserializeBC extract the bucketclass object from the request
+func (bcv *BucketClassValidator) DeserializeBC(rawBS []byte) *nbv1.BucketClass {
+	BC := nbv1.BucketClass{}
+	if err := json.Unmarshal(rawBS, &BC); err != nil {
+		bcv.Logger.Error("error deserializing bucketclass")
+	}
+	return &BC
+}
+
+// ValidateCreate runs all the validations tests for CREATE operations
+func (bcv *BucketClassValidator) ValidateCreate() {
+	bc := bcv.DeserializeBC(bcv.arRequest.Request.Object.Raw)
+	if bc == nil {
+		return
+	}
+
+	if err := bucketclass.ValidateQuotaConfig(bc.Name, bc.Spec.Quota); err != nil && util.IsValidationError(err) {
+		bcv.SetValidationResult(false, err.Error())
+		return
+	}
+	if bc.Spec.NamespacePolicy != nil {
+		if err := bucketclass.ValidateNSFSSingleBC(bc); err != nil && util.IsValidationError(err) {
+			bcv.SetValidationResult(false, err.Error())
+			return
+		}
+	}
+	if bc.Spec.PlacementPolicy != nil {
+		if err := bucketclass.ValidateTiersNumber(bc.Spec.PlacementPolicy.Tiers); err != nil {
+			bcv.SetValidationResult(false, err.Error())
+			return
+		}
+	}
+}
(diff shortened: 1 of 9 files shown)


pkg/admission/validate_bucketclass.go
```diff
@@ -0,0 +1,207 @@
...
+func (bcv *BucketClassValidator) ValidateTiersNumber() (bool, string) {
+	numTiers := len(bcv.BucketClass.Spec.PlacementPolicy.Tiers)
+	if numTiers != 1 && numTiers != 2 {
+		return false, "Unsupported number of tiers, bucketClass supports only 1 or 2 tiers"
+	}
+	return true, "allowed"
+}
+
+// ValidateNSFSSingleBC validates that bucketclass configured to NS of type NSFS it will only be of type Single.
+func (bcv *BucketClassValidator) ValidateNSFSSingleBC() (bool, string) {
```
romayalon, line 115: don't we want to test for all namespace bucketclasses that their namespacestores exists and ready? not only nsfs and not just single
besides that, nsfs can be currently only of type single but it's another validation

pkg/admission/validate_bucketclass.go
```diff
@@ -0,0 +1,207 @@
...
+
+// ValidateCreate runs all the validations tests for CREATE operations
+func (bcv *BucketClassValidator) ValidateCreate() {
+	bcv.BucketClass = bcv.DeserializeBC(bcv.arRequest.Request.Object.Raw)
+
+	if ok, message := bcv.ValidateQuota(); !ok {
+		bcv.SetValidationResult(ok, message)
+		return
+	}
+	if bcv.BucketClass.Spec.NamespacePolicy != nil {
```
romayalon, line 83: 2 things here:
1. namespace bucketclass can be of type cache and then you need to validate its tiering
2. you should also test that backingstores exist and ready

pkg/admission/validate_bucketclass.go
```diff
@@ -0,0 +1,207 @@
...
+			bcv.Logger.Error("Failed to check namespacestore in NSFS single bucketclass type validation")
+			return true, "allowed"
+		}
+		if nsStore.Spec.Type == nbv1.NSStoreTypeNSFS {
+			return false, fmt.Sprintf("Invalid namespaceStore types, NSFS NamespaceStore %q is allowed on bucketclass of type Single", name)
+		}
+	}
+	return true, "allowed"
+}
+
```
romayalon, line 145: what about replication policy validation ? :)

pkg/admission/validate_bucketclass.go
```diff
@@ -0,0 +1,207 @@
...
+// ValidateQuota validates the quota configuration of the bucket class if provided
+func (bcv *BucketClassValidator) ValidateQuota() (bool, string) {
+	err := bucketclass.ValidateBucketClass(bcv.BucketClass)
```
romayalon, line 148: instead of calling validateBucketClass you should call validateQuotaConfig, in the future validateBucketClass will contain more validations..
which makes me think, why not add the validations to the bucketclass validation file and just call it ?

pkg/admission/validate_bucketclass.go
```diff
@@ -0,0 +1,207 @@
...
+			bcv.Logger.Errorf("failed to list namespacestores in the system")
+			return true, "allowed"
+		}
+
+		// check that none of the ns in namespaceStoresArr is in nsList.Items[]
+		for _, ns := range nsList.Items {
+			if util.Contains(ns.Name, namespaceStoresArr) {
+				return false, fmt.Sprintf("Failed to delete Bucketclass, please delete %s Namespacestore first", ns.Name)
+			}
+		}
```
romayalon, line 184: cache backingstores are not being validated in this case

pkg/util/util.go
```diff
@@ -75,6 +75,38 @@ type OAuth2Endpoints struct {
 	TokenEndpoint         string `json:"token_endpoint"`
 }
 
+// ValidationError is a custom error if the validation failed
+type ValidationError struct {
+	Msg string
+}
+
+// TechnicalError is a custom error if an operation in the validation process failed
+type TechnicalError struct {
```
dannyzaken, line 84: I don't think there is a need for a `TechnicalError` type. it will just force you to catch all non-validation errors and return them as a TecnicalError

pkg/admission/validate_bucketclass.go
```diff
@@ -0,0 +1,89 @@
...
+	BC := nbv1.BucketClass{}
+	if err := json.Unmarshal(rawBS, &BC); err != nil {
+		bcv.Logger.Error("error deserializing bucketclass")
+	}
+	return &BC
+}
+
+// ValidateCreate runs all the validations tests for CREATE operations
+func (bcv *BucketClassValidator) ValidateCreate() {
+	bc := bcv.DeserializeBC(bcv.arRequest.Request.Object.Raw)
```
romayalon, line 71: what will happen here if bc was unmarshaled and the bucketclass is empty?

kfir-payne: there is a check if the request body is empty on the `validator.go`, but we can add another check here.

pkg/bucketclass/validation.go
```diff
@@ -1,23 +1,125 @@
...
+	case nbv1.NSBucketClassTypeSingle:
+		namespaceStoresArr = append(namespaceStoresArr, bc.Spec.NamespacePolicy.Single.Resource)
+	}
+
+	return namespaceStoresArr
+}
+
+// ValidateNSFSSingleBC validates that bucketclass configured to NS of type NSFS it will only be of type Single.
+func ValidateNSFSSingleBC(bc *nbv1.BucketClass) error {
+	namespaceStoresArr := GetBucketclassNamespaceStoreArray(bc)
```
romayalon, line 58: where is the check that the bucketclass is of type single?

pkg/bucketclass/validation.go
```diff
@@ -1,23 +1,125 @@
...
+			return util.ValidationError{
+				Msg: fmt.Sprintf("invalid namespaceStore types, nsfs namespacestore %q is allowed on bucketclass of type single", name),
+			}
+		}
+	}
+	return nil
+}
+
+// ValidateDependentResourcesOnDeletion validates that BS/NS associated with the bucketclass will be deleted first
+func ValidateDependentResourcesOnDeletion(bc *nbv1.BucketClass) error {
```
romayalon, line 81: this function should be deleted, right?

pkg/admission/validate_bucketclass.go
romayalon: I think it can fail for invalid structure etc too

pkg/bucketclass/validation.go
```diff
@@ -1,23 +1,77 @@
...
+			TypeMeta: metav1.TypeMeta{Kind: "NamespaceStore"},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: bc.Namespace,
+			},
+		}
+		if !util.KubeCheck(nsStore) {
+			return fmt.Errorf("failed to check namespacestore in NSFS single bucketclass type validation")
+		}
+		if nsStore.Spec.Type == nbv1.NSStoreTypeNSFS && bc.Spec.NamespacePolicy.Type != nbv1.NSBucketClassTypeSingle {
```
romayalon, line 70: move `bc.Spec.NamespacePolicy.Type != nbv1.NSBucketClassTypeSingle`  before the first line of this function so you won't even collect the namespacestores and do this test if this bucketclass is of type single

romayalon (approved):

pkg/admission/validate_bucketclass.go
```diff
@@ -0,0 +1,92 @@
...
+	return bcv
+}
+
+// ValidateBucketClass call appropriate validations based on the operation
+func (bcv *BucketClassValidator) ValidateBucketClass() admissionv1.AdmissionReview {
+	switch bcv.arRequest.Request.Operation {
+	case admissionv1.Create:
+		bcv.ValidateCreate()
+	default:
+		bcv.Logger.Error("Failed to identify bucketclass operation type")
```
dannyzaken, line 49: this will be printed for any operation other than Create. can make the logs confusing

kfir-payne: In the ValidatingWebhookConfiguration file, only CREATE operations for bucketclass resource should be webhooked, so any other operation shouldn't even be processed by the admission server, the default case here is just best practice design to be safe.

dannyzaken (approved):

pkg/admission/validate_bucketclass.go
kfir-payne: same goes for the resource identification switch case in `validator.go`

PR #825, opened 2021-12-28
Refactor Backingstore and Namespacestore validations

kfir-payne:

### Explain the changes
1. move backingstore and namespacestore validations to a single file so that the cli and the operator could use them, also to avoid duplicate code.

1. 

### Testing Instructions:
1.

### DIFF
--- a/pkg/admission/test/unit/admission_unit_test.go
+++ b/pkg/admission/test/unit/admission_unit_test.go
@@ -1,10 +1,13 @@
 package admissionunittests
 
 import (
-	"github.com/noobaa/noobaa-operator/v5/pkg/admission"
+	"fmt"
+
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	"github.com/noobaa/noobaa-operator/v5/pkg/backingstore"
 	"github.com/noobaa/noobaa-operator/v5/pkg/bucketclass"
 	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
+	"github.com/noobaa/noobaa-operator/v5/pkg/namespacestore"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -15,28 +18,22 @@ import (
 var _ = Describe("BackingStore admission unit tests", func() {
 
 	var (
-		bsv          *admission.BackingStoreValidator
-		backingstore *nbv1.BackingStore
-		result       bool
-		message      string
+		bs  *nbv1.BackingStore
+		err error
 	)
 
 	BeforeEach(func() {
-		backingstore = util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_backingstore_cr_yaml).(*nbv1.BackingStore)
-		backingstore.Name = "bs-name"
-		backingstore.Namespace = "test"
-		bsv = &admission.BackingStoreValidator{
-			BackingStore: backingstore,
-		}
-		result = false
-		message = ""
+		bs = util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_backingstore_cr_yaml).(*nbv1.BackingStore)
+		bs.Name = "bs-name"
+		bs.Namespace = "test"
+
 	})
 
 	Describe("Validate create operations", func() {
 		Describe("General backingstore validations", func() {
 			Context("Empty secret name", func() {
 				It("Should Deny", func() {
-					bsv.BackingStore.Spec = nbv1.BackingStoreSpec{
+					bs.Spec = nbv1.BackingStoreSpec{
 						Type: nbv1.StoreTypeAWSS3,
 						AWSS3: &nbv1.AWSS3Spec{
 							TargetBucket: "some-target-bucket",
@@ -46,12 +43,12 @@ var _ = Describe("BackingStore admission unit tests", func() {
 							},
 						},
 					}
-					result, message = bsv.ValidateBSEmptySecretName()
-					Expect(result).To(BeFalse())
-					Expect(message).To(Equal("Failed creating the Backingstore, please provide secret name"))
+					err = backingstore.ValidateBSEmptySecretName(*bs)
+					Ω(err).Should(HaveOccurred())
+					Expect(err.Error()).To(Equal("Failed creating the Backingstore, please provide secret name"))
 				})
 				It("Should Allow", func() {
-					bsv.BackingStore.Spec = nbv1.BackingStoreSpec{
+					bs.Spec = nbv1.BackingStoreSpec{
 						Type: nbv1.StoreTypeAWSS3,
 						AWSS3: &nbv1.AWSS3Spec{
 							TargetBucket: "some-target-bucket",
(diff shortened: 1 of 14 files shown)


pkg/admission/validate_namespacestore.go
```diff
@@ -2,272 +2,107 @@ package admission
...
-func (nsv *NamespaceStoreValidator) SetValidationResult(isAllowed bool, message string) {
-	nsv.arResponse.Response.Allowed = isAllowed
-	nsv.arResponse.Response.Result.Message = message
-}
-
 // DeserializeNS extract the namespacestore object from the request
-func (nsv *NamespaceStoreValidator) DeserializeNS(rawNS []byte) *nbv1.NamespaceStore {
+func (nsv *ResourceValidator) DeserializeNS(rawNS []byte) *nbv1.NamespaceStore {
 	NS := nbv1.NamespaceStore{}
 	if err := json.Unmarshal(rawNS, &NS); err != nil {
```
romayalon, line 55: I know it's not related to your changes but notice that the parameter rawNS can be extracted from nsv, so it's redundant

pkg/admission/test/unit/admission_unit_test.go
```diff
@@ -405,124 +353,98 @@ var _ = Describe("NamespaceStore admission unit tests", func() {
 							},
 						},
 					}
-					result, message = nsv.ValidateNSEmptySecretName()
-					Expect(result).To(BeTrue())
-					Expect(message).To(Equal("allowed"))
-				})
-			})
-			Context("Invalid store type", func() {
```
romayalon, line 413: why all the invalid store type tests were deleted?

pkg/namespacestore/validation.go
```diff
@@ -7,40 +7,47 @@ import (
...
 
 	case nbv1.NSStoreTypeIBMCos:
-		return validateNsStoreIBMCos(nsStore)
+		return ValidateNsStoreIBMCos(nsStore)
 
 	case nbv1.NSStoreTypeAzureBlob:
 		return nil
 
 	default:
-		return util.NewPersistentError("ConfigurationError",
```
romayalon, line 37: by removing the persistent error, the places that already call this function will keep reconciling, re-check that change if this is the desired behavior

pkg/admission/validate_namespacestore.go
kfir-payne: but sometime I use this function to deserialize `arRequest.Request.Object.Raw` and sometimes I use it to deserialize `bsv.arRequest.Request.OldObject.Raw`.
I keep it `ResourceValidator` struct function to print the appropriate error if the `json.Unmarshal()` failed.

pkg/namespacestore/validation.go
kfir-payne: correct, I fixed it by triggering a `NewPersistentError` in the places that already call this function (BTW its called only twice, once in `namespacestore.go` and if the validations fails it will `log.Fatalf()`, and the second place is in the reconciler which then I trigger a persistent error).

pkg/admission/test/unit/admission_unit_test.go
kfir-payne: re-added them :)

pkg/namespacestore/validation.go
romayalon: since validation.go under backingstore is a new file we currently don't have this issue in backingstore, why won't we change the equivalent places in backingstore.go and backingstore reconciler to call to correct validations in validation file?

kfir-payne: you are right, but I think this change is out of scope for this PR, the end goal is to implement what you described.

dannyzaken: let's add the backingstore reconciler validations as part of this refactor

pkg/bucketclass/reconciler.go
```diff
@@ -209,11 +209,6 @@ func (r *Reconciler) ReconcilePhaseVerifying() error {
 			fmt.Sprintf("NooBaa system %q not found or deleted", r.NooBaa.Name))
 	}
 	if r.BucketClass.Spec.PlacementPolicy != nil {
-		numTiers := len(r.BucketClass.Spec.PlacementPolicy.Tiers)
```
romayalon, line 212: where do you add the call for the validation tests that were removed in this file?

kfir-payne: 10 lines above, with:
```
err := ValidateBucketClass(r.BucketClass)
if err != nil {
    return util.NewPersistentError("ValidationError", err.Error())
}
```

kfir-payne: the validation call was already there, I just needed to remove the validations from the reconciler and add them to `bucketclass.ValidateBucketClass()`

pkg/backingstore/reconciler.go
```diff
@@ -528,36 +532,18 @@ func (r *Reconciler) ReadSystemInfo() error {
...
-		const minimalVolumeSize = int64(16 * 1024 * 1024 * 1024) // 16Gi=16*1024^3
 		var volumeSize int64
+
 		pvPool := r.BackingStore.Spec.PVPool
 		if pvPool.VolumeResources != nil {
 			qty := pvPool.VolumeResources.Requests[corev1.ResourceName(corev1.ResourceStorage)]
 			volumeSize = qty.Value()
-		}
-		if volumeSize < minimalVolumeSize {
 			if volumeSize == 0 {
```
romayalon, line 542: volume size will never be 0 because we reject it in the validation, we should decide if we want to keep the same behavior or not

kfir-payne: we can change the validation to allow 0 and then the reconciler will assign the default value.

kfir-payne: @romayalon @dannyzaken WDYT?

romayalon: I think it'll work

pkg/backingstore/reconciler.go
```diff
@@ -576,10 +562,6 @@ func (r *Reconciler) ReadSystemInfo() error {
 				return err
 			}
 
-			if len(hostsInfo.Hosts) > pvPool.NumVolumes { // scaling down - not supported
```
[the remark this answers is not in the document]
kfir-payne: your right, I'll fix it

pkg/backingstore/reconciler.go
dannyzaken: any idea why we allowed 0 here before?

kfir-payne: yeah its weird, in the cli we forbid 0 value, only above the 16Gi minimum.

kfir-payne: > 

if the reconciler identifies 0 value it will assign the default size, which is 20G

pkg/backingstore/validation.go
```diff
@@ -0,0 +1,222 @@
...
+		return util.ValidationError{
+			Msg: "Unsupported BackingStore name length, please provide a name shorter then 43 characters",
+		}
+	}
+	return nil
+}
+
+// ValidatePvpoolMinVolSize validates pvpool volume size is above 16GB
+func ValidatePvpoolMinVolSize(bs nbv1.BackingStore) error {
+	min := *resource.NewScaledQuantity(int64(MinimumVolumeSize), resource.Giga)
```
dannyzaken, line 106: notice that `VolumeResources` is optional and can be nil. I think that this is why the volume size can be 0 in reconciler

kfir-payne: but before assigning the default value we specifically check that `VolumeResources != nil` and that the `Request == 0`:
```
const defaultVolumeSize = int64(20 * 1024 * 1024 * 1024) // 20Gi=20*1024^3
const minimalVolumeSize = int64(16 * 1024 * 1024 * 1024) // 16Gi=16*1024^3
var volumeSize int64
pvPool := r.BackingStore.Spec.PVPool
if pvPool.VolumeResources != nil {
	qty := pvPool.VolumeResources.Requests[corev1.ResourceName(corev1.ResourceStorage)]
	volumeSize = qty.Value()
}
if volumeSize < minimalVolumeSize {
	if volumeSize == 0 {
		volumeSize = int64(defaultVolumeSize)
	} else {
		return util.NewPersistentError("SmallVolumeSize",
			fmt.Sprintf("NooBaa BackingStore %q is in rejected phase due to insufficient size, min is %d=%gGB", r.BackingStore.Name, minimalVolumeSize, (float64(minimalVolumeSize)/(math.Pow(1024, 3)))))
	}
}
```

dannyzaken (approved):
LGTM

PR #826, opened 2021-12-29
Support AWS STS in Noobaa

ezio-auditore:
### Explain the changes
1. Added spec field for backing stores crd to include Role ARN for AWS STS
2. Added spec field for noobaa crd to include Role ARN for AWS STS
3. Added Projected Service account token mount to noobaa-operator and nooba-core pods
4. Added function to create AWS STS backing store
5. Added noobaa install cli option `--aws-sts-role-arn`
6. Added check for AWS STS cluster
7. Added Admission Webhook validation for BackingStore
8. Added Validation check for NamespaceStore to not include AWS STS ARN
- [ ] Add unit tests
Signed-off-by: Kaustav Majumder <kmajumde@redhat.com>

### Testing Instructions:
1. Create an STS based Kubernetes or Openshift cluster [see more..](https://docs.openshift.com/container-platform/4.7/authentication/managing_cloud_provider_credentials/cco-mode-sts.html#sts-mode-installing-manual-config_cco-mode-sts)
2. Create Role for S3 full access and assign to Identity Provider which was created during cluster install
3. Use noobaa cli to pass in the ARN for the role created previously 
`nb install --aws-sts-role-arn=arn:aws:iam::261532230807:role/kmajumder_noobaa_sts`
4. Check if the default backing store reaches ready state.

### DIFF
--- a/deploy/internal/statefulset-core.yaml
+++ b/deploy/internal/statefulset-core.yaml
@@ -33,6 +33,13 @@ spec:
           secret:
             secretName: noobaa-s3-serving-cert
             optional: true
+        - name: oidc-token
+          projected:
+            sources:
+            - serviceAccountToken:
+                path: oidc-token
+                expirationSeconds: 3600
+                audience: api
       containers:
         #----------------#
         # CORE CONTAINER #
@@ -47,6 +54,9 @@ spec:
               readOnly: true
             - name: s3-secret
               mountPath: /etc/s3-secret
+              readOnly: true
+            - mountPath: /var/run/secrets/openshift/serviceaccount
+              name: oidc-token
               readOnly: true
           resources:
             requests:
--- a/pkg/apis/noobaa/v1alpha1/backingstore_types.go
+++ b/pkg/apis/noobaa/v1alpha1/backingstore_types.go
@@ -148,6 +148,7 @@ type AWSS3Spec struct {
 
 	// Secret refers to a secret that provides the credentials
 	// The secret should define AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY
+	// +optional
 	Secret corev1.SecretReference `json:"secret"`
 
 	// Region is the AWS region
@@ -157,6 +158,9 @@ type AWSS3Spec struct {
 	// SSLDisabled allows to disable SSL and use plain http
 	// +optional
 	SSLDisabled bool `json:"sslDisabled,omitempty"`
+
+	// +optional
+	AWSSTSRoleARN *string `json:"awsSTSRoleARN,omitempty"`
 }
 
 // S3CompatibleSpec specifies a backing store of type s3-compatible
--- a/pkg/apis/noobaa/v1alpha1/noobaa_types.go
+++ b/pkg/apis/noobaa/v1alpha1/noobaa_types.go
@@ -117,7 +117,7 @@ type NooBaaSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum=all;nsfs;warn;default_level
 	DebugLevel int `json:"debugLevel,omitempty"`
-	
+
 	// PVPoolDefaultStorageClass (optional) overrides the default cluster StorageClass for the pv-pool volumes.
 	// This affects where the system stores data chunks (encrypted).
 	// Updates to this field will only affect new pv-pools,
@@ -175,6 +175,9 @@ type NooBaaSpec struct {
 	// +nullable
 	// +optional
 	DisableLoadBalancerService bool `json:"disableLoadBalancerService,omitempty"`
+
+	// +optional
+	DefaultBackingStoreSpec *BackingStoreSpec `json:"defaultBackingStoreSpec,omitempty"`
 }
 
 // SecuritySpec is security spec to include various security items such as kms
@@ -295,19 +298,19 @@ const (
 // These are NooBaa condition statuses
 const (
 	// External KMS initialized
-	ConditionKMSInit        corev1.ConditionStatus = "Init"
+	ConditionKMSInit corev1.ConditionStatus = "Init"
 
 	// The root key was synchronized from external KMS
-	ConditionKMSSync        corev1.ConditionStatus = "Sync"
+	ConditionKMSSync corev1.ConditionStatus = "Sync"
 
 	// Invalid external KMS definition
-	ConditionKMSInvalid     corev1.ConditionStatus = "Invalid"
+	ConditionKMSInvalid corev1.ConditionStatus = "Invalid"
 
 	// Error reading secret from external KMS
-	ConditionKMSErrorRead   corev1.ConditionStatus = "ErrorRead"
+	ConditionKMSErrorRead corev1.ConditionStatus = "ErrorRead"
 
 	// Error writing initial root key to eternal KMS
-	ConditionKMSErrorWrite  corev1.ConditionStatus = "ErrorWrite"
+	ConditionKMSErrorWrite corev1.ConditionStatus = "ErrorWrite"
 )
 
 // AccountsStatus is the status info of admin account
--- a/pkg/backingstore/validation.go
+++ b/pkg/backingstore/validation.go
@@ -49,11 +49,30 @@ func ValidateBackingStore(bs nbv1.BackingStore) error {
 	return nil
 }
 
+// ValidateBSEmptyAWSSTSARN validates if AWS STS ARN is provided for AWSS3 backingstores
+func ValidateBSEmptyAWSSTSARN(bs nbv1.BackingStore) error {
+	if bs.Spec.Type == nbv1.StoreTypeAWSS3 {
+		if bs.Spec.AWSS3.AWSSTSRoleARN != nil {
+			if len(*bs.Spec.AWSS3.AWSSTSRoleARN) == 0 {
+				return util.ValidationError{
+					Msg: "Failed creating the Backingstore, please provide  AWS STS ARN",
+				}
+			}
+		}
+	}
+	return nil
+}
+
 // ValidateBSEmptySecretName validates a secret name is provided for cloud backingstores
 func ValidateBSEmptySecretName(bs nbv1.BackingStore) error {
 	switch bs.Spec.Type {
 	case nbv1.StoreTypeAWSS3:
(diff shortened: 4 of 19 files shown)


nimrod-becker:
@jackyalbo @liranmauda @jeniawhite

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -175,6 +175,9 @@ type NooBaaSpec struct {
 	// +nullable
 	// +optional
 	DisableLoadBalancerService bool `json:"disableLoadBalancerService,omitempty"`
+
+	// +optional
+	AWSSTSRoleARN *string `json:"awsSTSRoleARN,omitempty"`
```
dannyzaken, line 180: didn't we discuss adding the default backing store spec here instead of ARN?
@nb-ohad @nimrod-becker

pkg/apis/noobaa/v1alpha1/backingstore_types.go
```diff
@@ -157,6 +158,9 @@ type AWSS3Spec struct {
 	// SSLDisabled allows to disable SSL and use plain http
 	// +optional
 	SSLDisabled bool `json:"sslDisabled,omitempty"`
+
+	// +optional
+	AWSSTSRoleARN *string `json:"awsSTSRoleARN,omitempty"`
```
dannyzaken, line 163: I'm not so sure it's good to have both `Secret` and `AWSSTSRoleARN` as optional. currently, we don't validate it in the admission controller. we rely on the static validation to reject a backingstore without a secret

dannyzaken: can we maybe use the same secret to hold either the ARN or the access keys?

nb-ohad: @dannyzaken The ARN is not confidential information so why complicate stuff (the need to create a secret in the first place)? Is there a benefit in doing so?

pkg/apis/noobaa/v1alpha1/noobaa_types.go
nb-ohad: Yes, we did, and I also remember that we all agreed that that will be the way to go. 
@ezio-auditore is this some leftover code? or are we still waiting for some uncommitted work?

pkg/apis/noobaa/v1alpha1/backingstore_types.go
dannyzaken: The current implementation breaks the validations of the `AWSS3Spec`. do you have a different suggestion?

nb-ohad: If we are moving this into the secret I believe we will need a new field in the spec to describe what kind of auth lives in the secret so we could validate that the current auth is provided. 
Something like: AWSAuthKind (Credentials, STS) 
WDYT?

dannyzaken: After some thought, I agree that using a secret for ARN makes it a bit more complicated to use. especially if, in the long run, we expect STS to be the more common method.

dannyzaken: Let's just add validation in the admission that checks if both the secret and ARN are missing

dannyzaken (commented):
this PR is missing required changes in namespace store reconcile. notice that AWSS3Spec is also used for namespace stores

https://github.com/noobaa/noobaa-operator/blob/2a7b7bd38101715e29d387549a490d6c49d11199/pkg/apis/noobaa/v1alpha1/namespacestore_types.go#L63-L65

deploy/internal/statefulset-core.yaml
```diff
@@ -33,6 +33,13 @@ spec:
           secret:
             secretName: noobaa-s3-serving-cert
             optional: true
+        - name: oidc-token
```
dannyzaken, line 36: do we always want to mount this?

nb-ohad: I think we tested that for the case where it does not exist and nothing harmful happened.
@ezio-auditore Can you please confirm?

nb-ohad:
@dannyzaken 
> this PR is missing required changes in namespace store reconcile. notice that AWSS3Spec is also used for namespace stores

If I remember correctly namespace store support was scoped out. 
I think we agreed to have a validation check in the reconciler that generates a failure if STS is used for a namespace store.

pkg/apis/noobaa/v1alpha1/noobaa_types.go
ezio-auditore: Yes it is leftover code. I have update the latest commit with the changes

pkg/apis/noobaa/v1alpha1/backingstore_types.go
ezio-auditore: Added validation to check if secret and ARN are both missing

ezio-auditore:
> @dannyzaken
> 
> > this PR is missing required changes in namespace store reconcile. notice that AWSS3Spec is also used for namespace stores
> 
> If I remember correctly namespace store support was scoped out. I think we agreed to have a validation check in the reconciler that generates a failure if STS is used for a namespace store.

Added a validation check to throw error if STS is used in namespace store

pkg/namespacestore/validation.go
```diff
@@ -28,6 +28,11 @@ func ValidateNamespaceStore(nsStore *nbv1.NamespaceStore) error {
 		return ValidateNsStoreNSFS(nsStore)
 
 	case nbv1.NSStoreTypeAWSS3:
+		if nsStore.Spec.AWSS3.AWSSTSRoleARN != nil {
```
dannyzaken, line 31: you also need to handle a `nil` secret in `ValidateNSEmptySecretName` which is called above

pkg/backingstore/validation.go
```diff
@@ -54,8 +54,10 @@ func ValidateBSEmptySecretName(bs nbv1.BackingStore) error {
 	switch bs.Spec.Type {
 	case nbv1.StoreTypeAWSS3:
 		if len(bs.Spec.AWSS3.Secret.Name) == 0 {
-			return util.ValidationError{
-				Msg: "Failed creating the Backingstore, please provide secret name",
+			if len(*bs.Spec.AWSS3.AWSSTSRoleARN) == 0 {
```
kfir-payne, line 57: I think it will be best to separate these validation to different functions, to be more accurate with error messages and also to keep function responsibility clear and single as a good design.

pkg/namespacestore/validation.go
kfir-payne, line 31: > I think it will be best to separate these validation to different functions, to be more accurate with error messages and also to keep function responsibility clear and single as a good design.

also here.

pkg/system/phase4_configuring.go
```diff
@@ -718,6 +726,56 @@ func (r *Reconciler) prepareAWSBackingStore() error {
 	return nil
 }
 
+func (r *Reconciler) prepareAWSSTSBackingStore() error {
+	region, err := util.GetAWSRegion()
```
dannyzaken, line 730: all this logic seems irrelevant here. you should already have the BS spec in noobaa CR, so just create it

dannyzaken (commented):
I think that the flow of reconciling the default backing store should be much simpler when the spec is provided in noobaa CR. Also, your implementation here does not take into account all other BS types. 

My assumption here is that when a deafultBackingstoreSpec is provided in the CR, it should already contain all of the relevant information (ARN, target bucket, etc.). This should be the same for any backingstore type.

I think that the changes for noobaa reconciler should be as follows:

**phase 1**  - validate the default backingstore spec. you can use the validations in backingstore package (we don't have an admission controller validations for noobaa CRD)

**phase 2** - there is nothing to do in this phase if a BS spec is provided, just skip if . the main reason we create the credentials request in phase 2 is because it can take time for the credentials to be valid. 

**phase 3** - nothing to change

**phase 4** - if a BS spec is provided in the noobaa CR **and** a default BS does not yet exist (we will handle changes in next releases) just create the default BS. nothing else to do. else use the old flow

pkg/namespacestore/validation.go
```diff
@@ -28,6 +28,11 @@ func ValidateNamespaceStore(nsStore *nbv1.NamespaceStore) error {
...
+			return util.ValidationError{
+				Msg: "AWS STS feature cannot be used for NamespaceStore",
```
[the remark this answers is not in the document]
ezio-auditore: Done

pkg/backingstore/validation.go
ezio-auditore: Seperated into different function

pkg/namespacestore/validation.go
ezio-auditore: Done

ezio-auditore: Done

pkg/namespacestore/validation.go
```diff
@@ -181,6 +186,9 @@ func ValidateNSEmptySecretName(ns nbv1.NamespaceStore) error {
 	switch ns.Spec.Type {
 	case nbv1.NSStoreTypeAWSS3:
 		if len(ns.Spec.AWSS3.Secret.Name) == 0 {
```
dannyzaken, line 188: `Secret` is now optional, so you need to validate it's not nil before dereferencing it

pkg/system/phase2_creating.go
```diff
@@ -59,6 +59,10 @@ func (r *Reconciler) ReconcilePhaseCreating() error {
 	if err := r.ReconcileObjectOptional(r.RouteS3, nil); err != nil {
 		return err
 	}
+	// If DefaultBackingStore Spec is set do nothing
+	if r.NooBaa.Spec.DefaultBackingStoreSpec != nil {
```
dannyzaken, line 63: I think it should be inside `ReconcileBackingStoreCredentials`

pkg/namespacestore/validation.go
dannyzaken: sorry. thought this is a pointer. ignore

dannyzaken (approved):

PR #827, opened 2021-12-30
Add noobaa-odf-ui service account's role rules

baum:
### Test:

- Run "noobaa olm csv"

Expected generated YAML should contain the following permission:

```yaml
  ...
  install:
    spec:
      permissions:
      ...
      - rules:
        - apiGroups:
          - noobaa.io
          resources:
          - noobaas
          - bucketclasses
          verbs:
          - get
          - watch
          - list
        serviceAccountName: noobaa-odf-ui
```
See https://bugzilla.redhat.com/show_bug.cgi?id=2031705

### DIFF
--- a/deploy/role_ui.yaml
+++ b/deploy/role_ui.yaml
@@ -1,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: noobaa-odf-ui
+rules:
+  - verbs:
+      - get
+      - watch
+      - list
+    apiGroups:
+      - noobaa.io
+    resources:
+      - noobaas
+      - bucketclasses
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -237,9 +237,11 @@ func RunYaml(cmd *cobra.Command, args []string) {
 type Conf struct {
 	NS                         *corev1.Namespace
 	SA                         *corev1.ServiceAccount
+	SAUI                       *corev1.ServiceAccount
 	SAEndpoint                 *corev1.ServiceAccount
 	Role                       *rbacv1.Role
 	RoleEndpoint               *rbacv1.Role
+	RoleUI                     *rbacv1.Role
 	RoleBinding                *rbacv1.RoleBinding
 	RoleBindingEndpoint        *rbacv1.RoleBinding
 	ClusterRole                *rbacv1.ClusterRole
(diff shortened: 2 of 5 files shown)


deploy/role_ui.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: noobaa-ui
```
dannyzaken, line 4: maybe change to `noobaa-odf-ui`, just to be more clear? I'm not really sure what is a good name here

deploy/role_binding_ui.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
```
dannyzaken, line 2: I don't think we need a role binding. we just want to create a role that can be bound to other service-accounts by the user

pkg/operator/operator.go
```diff
@@ -107,10 +107,13 @@ func RunInstall(cmd *cobra.Command, args []string) {
 	util.KubeCreateSkipExisting(c.NS)
 	util.KubeCreateSkipExisting(c.SA)
 	util.KubeCreateSkipExisting(c.SAEndpoint)
+	util.KubeCreateSkipExisting(c.SAUI)
```
dannyzaken, line 110: I think we can skip this change in the CLI and only have it in the OLM

deploy/role_ui.yaml
baum: ✅

pkg/operator/operator.go
baum: ✅

pkg/operator/operator.go
```diff
@@ -259,8 +261,10 @@ func LoadOperatorConf(cmd *cobra.Command) *Conf {
 	c.NS = util.KubeObject(bundle.File_deploy_namespace_yaml).(*corev1.Namespace)
 	c.SA = util.KubeObject(bundle.File_deploy_service_account_yaml).(*corev1.ServiceAccount)
 	c.SAEndpoint = util.KubeObject(bundle.File_deploy_service_account_endpoint_yaml).(*corev1.ServiceAccount)
+	c.SAUI = util.KubeObject(bundle.File_deploy_service_account_ui_yaml).(*corev1.ServiceAccount)
```
dannyzaken, line 264: do we need the service account?

dannyzaken (approved):

pkg/operator/operator.go
baum: required by olm, referenced [here](https://github.com/baum/noobaa-operator/blob/727593403702dcf80d920cea3c921f9dc236ff3d/pkg/olm/olm.go#L248-L252)

PR #828, opened 2021-12-30
STS service and route addition

romayalon:

### Explain the changes
1. Added STS route and service.

1. 

### Testing Instructions:
1.

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -59,6 +59,12 @@ func (r *Reconciler) ReconcilePhaseCreating() error {
 	if err := r.ReconcileObjectOptional(r.RouteS3, nil); err != nil {
 		return err
 	}
+	if err := r.ReconcileObject(r.ServiceSts, r.SetDesiredServiceSts); err != nil {
+		return err
+	}
+	if err := r.ReconcileObjectOptional(r.RouteSts, nil); err != nil {
+		return err
+	}
 	// the credentials that are created by cloud-credentials-operator sometimes take time
 	// to be valid (requests sometimes returns InvalidAccessKeyId for 1-2 minutes)
 	// creating the credential request as early as possible to try and avoid it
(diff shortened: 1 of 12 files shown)


pkg/system/phase2_creating.go
```diff
@@ -192,6 +198,19 @@ func (r *Reconciler) SetDesiredServiceS3() error {
...
+// SetDesiredServiceSts updates the ServiceSts as desired for reconciling
+func (r *Reconciler) SetDesiredServiceSts() error {
+	if r.NooBaa.Spec.DisableLoadBalancerService {
+		r.ServiceSts.Spec.Type = corev1.ServiceTypeClusterIP
+	} else {
+		// It is here in case disableLoadBalancerService is removed from the crd or changed to false
+		r.ServiceSts.Spec.Type = corev1.ServiceTypeLoadBalancer
+	}
+	r.ServiceSts.Spec.Selector["noobaa-s3"] = r.Request.Name
+	r.ServiceSts.Labels["noobaa-sts-svc"] = "true"
```
kfir-payne, line 210: the "noobaa-sts-svc" label is already configured in `service-sts.yaml` to true, it will be redundant to set it here too, no?

kfir-payne:
From what I see the intention is to deploy the service anyway, a not on the user request, correct?

pkg/system/phase2_creating.go
romayalon: As far as I understand this label is used by the service monitor, probably it's redundant, WDYT?

romayalon:
@kfir-payne 

> From what I see the intention is to deploy the service anyway, a not on the user request, correct?

correct

pkg/system/phase2_creating.go
kfir-payne: I don't see how this label is used by the service monitor, but either way, setting the label here and also in the yaml file is not required, if the service monitor do use this label then we should set it only once.

romayalon: currently, it's not used by the service monitor, but we might would like to add that in the future
you can take a look at https://github.com/noobaa/noobaa-operator/pull/619 and see that this is due to upgrade
issues we fixed so it's needed

romayalon: I think i'll just remove it in the yaml and in here too

liranmauda (approved):
LGTM

PR #833, opened 2022-01-10
Update diagnostics to support both kubectl and oc

jackyalbo:

### Explain the changes
1.  We have issue with using kubectl in mustgather - will be using kubectl if we have it in diagnostics, otherwise oc.
2.  If both don't exist we will fail the diagnostic.

1.  Fixed https://bugzilla.redhat.com/show_bug.cgi?id=2026342

### Testing Instructions:
1.

### DIFF
--- a/pkg/dbdump/dbdump.go
+++ b/pkg/dbdump/dbdump.go
@@ -32,6 +32,7 @@ type Collector struct {
 	remoteFolderPath	string // Remote folder (at pod) to store the raw db dump
 	remoteTarPath		string // Remote path (at pod) of tar'd db dump
 	kubeconfig 			string
+	kubeCommand			string
 	log					*logrus.Entry
 }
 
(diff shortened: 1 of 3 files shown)


liranmauda:
It should be done also in `dbdump.go`
For example:
https://github.com/noobaa/noobaa-operator/blob/07044a2f48cf8b3eb5ad6e64cc0d9ad7c6b33009/pkg/dbdump/dbdump.go#L74

jackyalbo:
> It should be done also in `dbdump.go` For example:
> 
> https://github.com/noobaa/noobaa-operator/blob/07044a2f48cf8b3eb5ad6e64cc0d9ad7c6b33009/pkg/dbdump/dbdump.go#L74

done. please re-review.

liranmauda (approved):
LGTM

PR #837, opened 2022-01-23
Fix r.OBC nil pointer dereference

baum:
### Changes:
- Fix r.OBC nil pointer dereference

## Issues:
-  https://bugzilla.redhat.com/show_bug.cgi?id=2038884

### DIFF
--- a/pkg/obc/provisioner.go
+++ b/pkg/obc/provisioner.go
@@ -341,9 +341,7 @@ func NewBucketRequest(
 			},
 		}
 		if !util.KubeCheck(r.BucketClass) {
-			msg := fmt.Sprintf("BucketClass %q not found in provisioner namespace %q", bucketClassName, p.Namespace)
-			p.recorder.Event(r.OBC, "Warning", "MissingBucketClass", msg)
-			return nil, fmt.Errorf(msg)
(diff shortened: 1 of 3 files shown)


dannyzaken (commented):
looks like this PR contains irrelevant changes (looks like changes from #827)

pkg/obc/provisioner.go
```diff
@@ -301,7 +301,9 @@ func NewBucketRequest(
 		}
 		if r.BucketClass.Status.Phase != nbv1.BucketClassPhaseReady {
 			msg := fmt.Sprintf("BucketClass %q is not ready", bucketClassName)
-			p.recorder.Event(r.OBC, "Warning", "BucketClassNotReady", msg)
+			if r.OBC != nil {
+				p.recorder.Event(r.OBC, "Warning", "BucketClassNotReady", msg)
```
dannyzaken, line 305: maybe I'm wrong, but I think that the panic was on the event for `MissingBucketClass` (line 299)

baum: good point; added nil pointer condition also on line 299.

dannyzaken (approved):

PR #857, opened 2022-02-21
Add Regenerate S3 Credentials for OBC to the CLI

liranmauda:
### Explain the changes
- Add Regenerate S3 Credentials for OBC to the CLI
- Fixed the RPC call for list accounts

### Testing Instructions:
1.  Regenerate OBC credentials
2. See that s3 ops are working with the new credentials

### DIFF
--- a/pkg/obc/obc.go
+++ b/pkg/obc/obc.go
@@ -27,6 +27,7 @@ func Cmd() *cobra.Command {
 	}
 	cmd.AddCommand(
 		CmdCreate(),
+		CmdRegenerate(),
 		CmdDelete(),
 		CmdStatus(),
 		CmdList(),
@@ -55,6 +56,16 @@ func CmdCreate() *cobra.Command {
 		"Set quota max objects quantity config to requested bucket")
 	cmd.Flags().String("max-size", "",
 		"Set quota max size config to requested bucket")
+	return cmd
+}
+
+// CmdRegenerate returns a CLI command
+func CmdRegenerate() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "regenerate <bucket-claim-name>",
+		Short: "Regenerate OBC S3 Credentials",
+		Run:   RunRegenerate,
+	}
 	return cmd
 }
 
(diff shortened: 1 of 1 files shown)


pkg/obc/obc.go
```diff
@@ -435,3 +495,80 @@ func CheckPhase(obc *nbv1.ObjectBucketClaim) {
...
+		log.Fatal(err)
+	}
+
+	if len(list.Accounts) == 0 {
+		log.Fatalf(`❌  No accounts found.\n`)
+	}
+
+	AccountName := ""
+	for k := range list.Accounts {
+		if strings.Contains(list.Accounts[k].Email, bucketName) {
```
romayalon, line 530: That's a dangerous approach for checking if this is the account we should regenerate..
for example, if the current account's email in the loop is ```obc-account.obc1obc1.%x@noobaa.io``` and the bucketName is obc1, you would change another account that you are not supposed to...

pkg/obc/obc.go
```diff
@@ -435,3 +495,80 @@ func CheckPhase(obc *nbv1.ObjectBucketClaim) {
...
+	// Checking that we can find the secret before we are calling the RPC to change the credentials.
+	secret := util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret)
+	secret.Namespace = options.Namespace
+	secret.Name = name
+	if !util.KubeCheckQuiet(secret) {
+		log.Fatalf(`❌  Could not find secret: %s, will not regenerate keys.`, secret.Name)
+	}
+
+	nbClient := sysClient.NBClient
+	list, err := nbClient.ListAccountsAPI(nb.ListAccountsParams{	})
```
romayalon, line 519: listing all accounts is redundant, the account name is being kept in the ob

romayalon (changes requested):

pkg/nb/api.go
```diff
@@ -18,7 +18,7 @@ type Client interface {
 	ReadNamespaceResourceOperatorInfoAPI(ReadNamespaceResourceParams) (NamespaceResourceOperatorInfo, error)
 	SetNamespaceStoreInfo(NamespaceStoreInfo) error
 
-	ListAccountsAPI() (ListAccountsReply, error)
```
romayalon, line 21: all the list account changes are not needed anymore, can be removed

liranmauda: It still needs to be fixed. 
We can divide it into a different PR, but I see no point in that. 
I prefer living it here, not even divide to commits.
WDYT?

romayalon: What is the issue that needs to be fixed?
The ListAccountsParams is just an empty struct, why do we need to pass it as params? look at listBucketAPI, there are no params there too. Also, I couldn't find a place in the operator code that uses this function.

liranmauda: the operator code is not using this RPC call, that's true, but this is still an issue in this call. 
If the code is there he might work.
I can think of 2 things:
1. Fix the not working calls for RPC even if they are not being used.
2. Remove those calls completely.

I don't think leaving a call that is not working is the correct way.

Regarding this specific call, It expects to get an empty object as a param, and this is what I have added 😃

pkg/obc/obc.go
```diff
@@ -435,3 +506,63 @@ func CheckPhase(obc *nbv1.ObjectBucketClaim) {
...
+	}
+ 
+	accessKeys = accountInfo.AccessKeys[0]
+
+	secret.StringData = map[string]string{}
+	secret.StringData["AWS_ACCESS_KEY_ID"] = accessKeys.AccessKey
+	secret.StringData["AWS_SECRET_ACCESS_KEY"] = accessKeys.SecretKey
+
+	//If we will not be able to update the secret we will print the credentials as they allready been changed by the RPC
+	if !util.KubeUpdate(secret) {
```
romayalon, line 559: @liranmauda  @dannyzaken  can we handle this situation better? maybe somehow hang the failure on the OBC/OB or somehow get the bucket into a rejected state and ask the user to regenerate again?

liranmauda: At this point, the account credentials are already changed and we are failing to update the secret.
The apps that are using this secret will also disconnect, and as we cant revert the best way to work around it is to regenerate again, and unless there is a fundamental issue with the secret that prevents us from updating it, it will fix it.
as a middle solution, we just print the new credentials (which can be manually edited into the secret as b64, by us/support/user).

@dannyzaken @romayalon

pkg/nb/api.go
dannyzaken: @liranmauda, what is not working with the current method signature? I also don't see a reason to add `ListAccountsParams`. can you elaborate on the need?

dannyzaken: Ok, so I see that it's not matching the account_api. in any case, I suggest doing it in a different PR and providing all other possible params (filter, etc.)

liranmauda: Opened #872 and will remove this code

dannyzaken (approved):

romayalon (approved):

PR #858, opened 2022-02-23
Add nsfs_account_config to noobaaaccount CRD

liranmauda:
### Explain the changes
- Adding nsfs_account_config to noobaaaccount CRD
- Adding the fields to CreateAccountParams and calling it on demand

### Gap 
1.  No webhooks yet
2. No CLI is provided yet.

both will be provided in the 2 next PRs. 

### Limitation
Removing `nsfs_account_config:` is not supported in the core, which means once we are adding it to the account CRD, we cannot remove it. validating it will be done in the webhook validation PR.

### Testing Instructions:
A. Update existing account CRD
1. Create an account with CRD using the CLI.
2. Edit the account CRD to contain `nsfs_account_config:` in its spec:
```
spec:
  allow_bucket_creation: true
  allowed_buckets:
    full_permission: true
    permission_list: []
  default_resource: noobaa-default-backing-store
  nsfs_account_config:
    gid: 777
    new_buckets_path: /
    nsfs_only: true
    uid: 776
```
3. save and see that it is updating the account.
4. Change a field in the `nsfs_account_config:` and see that it is getting updates.

B. Create account CRD that contains `nsfs_account_config:`
1. use a noobaaaccount yaml which has `nsfs_account_config:` in its spec and apply it.
2. See that the account that is being created have those field.
3. Change a field in the `nsfs_account_config:` and see that it is getting updates.


pkg/nb/types.go
```diff
@@ -391,6 +395,14 @@ type AccountAllowedBuckets struct {
 	PermissionList []string `json:"permission_list"`
 }
 
+// AccountNsfsConfig is the configuration of NSFS of CreateAccountParams
```
baum, line 398: Hello @liranmauda. The change looks good 😊 It might be a good idea to dedup `type AccountNsfsConfig struct {}` declaration here and also at [pkg/apis/noobaa/v1alpha1/noobaaaccount_types.go](https://github.com/liranmauda/noobaa-operator/blob/ab2e3dba85f35c8c2c54138fd7628b4db9601169/pkg/apis/noobaa/v1alpha1/noobaaaccount_types.go#L80-L86). 

[This](https://github.com/baum/noobaa-operator/commit/5cf5d24272d117b936a9fa8f769d5c361bd0e7e8) might be a way to achieve that.

liranmauda: Thanks.
Done.

pkg/nb/types.go
```diff
@@ -393,14 +399,15 @@ type AccountAllowedBuckets struct {
...
 type CreateAccountParams struct {
-	Name              string                `json:"name"`
-	Email             string                `json:"email"`
-	HasLogin          bool                  `json:"has_login"`
-	S3Access          bool                  `json:"s3_access"`
-	AllowBucketCreate bool                  `json:"allow_bucket_creation"`
-	AllowedBuckets    AccountAllowedBuckets `json:"allowed_buckets"`
-	DefaultResource   string                `json:"default_resource,omitempty"`
-	BucketClaimOwner  string                `json:"bucket_claim_owner,omitempty"`
+	Name              string					`json:"name"`
```
kfir-payne, line 402: notice the indentation here, you have a mix of spaces and tabs, can be solved by using spaces or tabs exclusively.
also happens in `UpdateAccountS3AccessParams, AccountNsfsConfig` and in these two if statements
`if r.NooBaaAccount.Spec.NsfsAccountConfig != nil` in `pkg/noobaaaccount/reconciler.go`

baum (approved):
LGTM

PR #859, opened 2022-02-24
Adding CLI flags is for creating nsfs account

liranmauda:
### Explain the changes

1.  Adding CLI flags is for creating nsfs account
2. Adding `uid`,`gid`,`new_buckets_path` and, `nsfs_only` flags to change the defaults.
3. Add GetFlaIntOrPrompt function to util that checks if the flag is -1 and if it is getting the input from stdin

1. Fixes https://bugzilla.redhat.com/show_bug.cgi?id=2047250

### Testing Instructions:
1. create an account via the CLI using those flags and see that it is an nsfs account

### DIFF
--- a/pkg/noobaaaccount/noobaaaccount.go
+++ b/pkg/noobaaaccount/noobaaaccount.go
@@ -55,6 +55,11 @@ func CmdCreate() *cobra.Command {
 	cmd.Flags().StringSlice("allowed_buckets", nil,
 		"Set the user allowed buckets list (use commas or multiple flags)")
 	cmd.Flags().String("default_resource", "", "Set the default resource, on which new buckets will be created")
+	cmd.Flags().Bool("nsfs_account_config", false, "This flag is for creating nsfs account")
+	cmd.Flags().Int("uid", -1, "Set the nsfs uid")
+	cmd.Flags().Int("gid", -1, "Set the nsfs gid")
+	cmd.Flags().String("new_buckets_path", "/", "Change the path where new buckets will be created")
+	cmd.Flags().Bool("nsfs_only", true, "Set if this account is used only for nsfs")
 	return cmd
 }
 
@@ -144,6 +149,7 @@ func RunCreate(cmd *cobra.Command, args []string) {
 	allowedBuckets := []string{}
 	fullPermission, _ := cmd.Flags().GetBool("full_permission")
 	bucketList, _ := cmd.Flags().GetStringSlice("allowed_buckets")
+
 	allowedBuckets = append(allowedBuckets, bucketList...)
 	if !fullPermission && len(allowedBuckets) == 0 {
 		log.Fatalf(`❌ Must provide at least one allowed buckets, or full_permission`)
@@ -151,8 +157,14 @@ func RunCreate(cmd *cobra.Command, args []string) {
 	if len(allowedBuckets) > 0 &&  fullPermission {
 		log.Fatalf(`❌ Can't provide both full_permission and an allowed buckets list`)
 	}
+
 	allowBucketCreate, _ := cmd.Flags().GetBool("allow_bucket_create")
 	defaultResource, _ := cmd.Flags().GetString("default_resource")
+
+	nsfsAccountConfig, _ := cmd.Flags().GetBool("nsfs_account_config")
+
+	newBucketsPath, _ := cmd.Flags().GetString("new_buckets_path")
+	nsfsOnly, _ := cmd.Flags().GetBool("nsfs_only")
 
 	// Check and get system
 	o := util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_noobaa_cr_yaml)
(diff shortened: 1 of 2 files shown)


liranmauda:
This PR is a draft as it is built upon #858
Once 858 is merged we will rebase from master and mark this one as ready to review.

liranmauda:
### Open questions
- Are those default values make sense? 
- Do we want not to use flags and just give a yaml/json as an input? (personally, I prefer using flags)

kfir-payne:
about the open questions:
1. Haven’t experimented with `nsfs` too much, but these default values do seem a little too permissive.
2. It really is a long list of configuration values and to supply all of them in the cli command would make it too complicated in my opinion, I think they should be delivered in a config file (yaml/json), after all it is `nsfs_account_config`

guymguym:
@kfir-payne the cli is meant to provide simplified flags interface with sensible defaults, if you want yaml you can use kubectl directly to create the crd, right?

liranmauda:
777 is kind of misleading...
UID and GID are the user and group and not the file's... 
our container uses 10001.
I think that I can do a -1 default and force to change it.

pkg/util/util.go
```diff
@@ -1091,7 +1091,25 @@ func IsValidS3BucketName(name string) bool {
 	return validBucketNameRegex.MatchString(name)
 }
 
-// GetFlagStringOrPrompt returns flag value but if empty will promtp to read from stdin
+// GetFlaIntOrPrompt returns flag value but if empty will prompt to read from stdin
+func GetFlaIntOrPrompt(cmd *cobra.Command, flag string) int {
```
romayalon, line 1095: GetFlaIntOrPrompt missing a g

pkg/util/util.go
```diff
@@ -1091,7 +1091,25 @@ func IsValidS3BucketName(name string) bool {
...
+	val, _ := cmd.Flags().GetInt(flag)
+	if val != -1 {
+		return val
+	}
+	fmt.Printf("Enter %s: ", flag)
+	_, err := fmt.Scan(&val)
+	if strings.Contains(err.Error(), "expected integer") {
```
romayalon, line 1102: do we want to test here for negative numbers?

romayalon:
@liranmauda we should add a test also :)

pkg/util/util.go
liranmauda: It is general util, We cannot assume that a negative is forbidden.
The main reason for this error handling is to avoid unclean exit with panic when we expect an integer and we get a string.

romayalon: so you can test it in the cli?

kfir-payne:
@guymguym I meant to something similar to what we do in the replication config, 
provide a path to `nsfs_account_config` file in a cli flag.
but maybe it's an overkill, again you can use the crd as you suggested.
just thinking out loud :)

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -55,6 +55,11 @@ func CmdCreate() *cobra.Command {
 	cmd.Flags().StringSlice("allowed_buckets", nil,
 		"Set the user allowed buckets list (use commas or multiple flags)")
 	cmd.Flags().String("default_resource", "", "Set the default resource, on which new buckets will be created")
+	cmd.Flags().Bool("nsfs_account_config", false, "This flag is for creating nsfs account")
```
jackyalbo, line 58: not sure why we need this flag? can't we understand from getting uid/gid that we want to use nsfs?

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -55,6 +55,11 @@ func CmdCreate() *cobra.Command {
...
+	cmd.Flags().Int("uid", -1, "Set the nsfs uid")
+	cmd.Flags().Int("gid", -1, "Set the nsfs gid")
+	cmd.Flags().String("new_buckets_path", "/", "Change the default nsfs mount path")
```
jackyalbo, line 61: change the default nsfs mount path sounds misleading -  maybe keep with Set the new bucket (default) path

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -55,6 +55,11 @@ func CmdCreate() *cobra.Command {
...
+	cmd.Flags().Bool("nsfs_only", true, "Set if this account for nsfs_only")
```
jackyalbo, line 62: sound weird... maybe Set if this account is used only for nsfs. maybe @romayalon can help with all of the above descriptions.

jackyalbo (changes requested):

pkg/noobaaaccount/noobaaaccount.go
jackyalbo: or maybe explain that it will set the path where new buckets of this account will be created.

pkg/noobaaaccount/noobaaaccount.go
liranmauda: 1. We want it to be explicit, as nsfs account config is optional.
2. uid/gid have no defaults, and in an even that we want nsfs config we want to pop a prompt that ask the values.

pkg/noobaaaccount/noobaaaccount.go
liranmauda: I am not sure I agree, this is the mount path for the nsfs.

jackyalbo: this is per account right? not really the mount point

pkg/noobaaaccount/noobaaaccount.go
jackyalbo: uid/gid have -1 default no? if we get one of these we can pop a prompt - not that I really sure why this prompt is needed - you can just fail requests with illegal/missing uid/gid.

jackyalbo (approved):
As it's urgent I will approve. We should probably add tests in the future and we should probably design this kind of cli changes first.

PR #697, opened 2021-07-29
Support building noobaa-operator on ARM64 and other architectures

AlessandroPomponio:
As of now NooBaa Operator cannot compile on architectures different from AMD64 at the moment due to two reasons:

- `x86_64` being hardcoded in [install-operator-sdk.sh](https://github.com/noobaa/noobaa-operator/blob/master/build/install-operator-sdk.sh).
- `operator-framework/operator-registry` only providing AMD64 images.

Both of these issues cause an `exec format error`. 

This PR allows `install-operator-sdk.sh` to download the binary for the correct architecture when launching the build process and updates the `Makefile` to include building locally the `operator-registry` image. With these changes, the operator can successfully compile on different architectures, such as `aarch64`.

Note that the installation with `nb install` will still hang due to `nooba-core` not supporting other architectures yet.

PR #861, opened 2022-02-28
Adding webhook to noobaaaccounts - nsfs config only

liranmauda:
### Explain the changes
- Adding webhook to noobaaaccounts - nsfs config only

### DIFF
--- a/pkg/admission/validate_noobaaaccount.go
+++ b/pkg/admission/validate_noobaaaccount.go
@@ -1,0 +1,93 @@
+package admission
+
+import (
+	"encoding/json"
+	"reflect"
+
+	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+	"github.com/noobaa/noobaa-operator/v5/pkg/validations"
+	"github.com/sirupsen/logrus"
+	admissionv1 "k8s.io/api/admission/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// NewNoobaaAccountValidator initializes a NoobaaAccountValidator to be used for loading and validating a noobaaaccount
+func NewNoobaaAccountValidator(arRequest admissionv1.AdmissionReview) *ResourceValidator {
+	nav := &ResourceValidator{
+
+		Logger:    logrus.WithField("admission noobaaaccount validation", arRequest.Request.Namespace),
+		arRequest: &arRequest,
+		arResponse: &admissionv1.AdmissionReview{
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "AdmissionReview",
+				APIVersion: "admission.k8s.io/v1",
+			},
+			Response: &admissionv1.AdmissionResponse{
+				UID:     arRequest.Request.UID,
+				Allowed: true,
+				Result: &metav1.Status{
+					Message: "allowed",
+				},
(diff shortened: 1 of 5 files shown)


kfir-payne:
In general, I suggest to add integration and unit tests.

pkg/admission/validate_noobaaaccount.go
```diff
@@ -0,0 +1,90 @@
...
+	if na == nil || oldNA == nil {
+		return
+	}
+
+	if err := validations.ValidateRemoveNSFSConfig(*na, *oldNA); err != nil && util.IsValidationError(err) {
+		nav.SetValidationResult(false, err.Error())
+		return
+	}
+
+	if err := validations.ValidateNSFSConfig(*na); err != nil && util.IsValidationError(err) {
```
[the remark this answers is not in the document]
liranmauda: true, also in create, as those fields are optional.

build/install-operator-sdk.sh
```diff
@@ -20,8 +21,8 @@ then
     fi
 fi
 
-PLATFORM="x86_64"
```
kfir-payne, line 23: can you explain these changes?

pkg/admission/validate_noobaaaccount.go
```diff
@@ -0,0 +1,101 @@
...
+			return
+		}
+	}	
+}
+
+// shouldRunValidateNSFSConfig check if we are adding the NsfsAccountConfig or if we are updating it.
+func shouldRunValidateNSFSConfig(na nbv1.NooBaaAccount, oldNa nbv1.NooBaaAccount) bool {
+	return oldNa.Spec.NsfsAccountConfig == nil &&  na.Spec.NsfsAccountConfig != nil ||
+		na.Spec.NsfsAccountConfig != nil && !reflect.DeepEqual(oldNa.Spec.NsfsAccountConfig, na.Spec.NsfsAccountConfig)
+}
```
kfir-payne, line 101: missing line in end of file

pkg/admission/validate_noobaaaccount.go
```diff
@@ -0,0 +1,101 @@
...
+		if err := validations.ValidateNSFSConfig(*na); err != nil && util.IsValidationError(err) {
+			nav.SetValidationResult(false, err.Error())
+			return
+		}
+	}	
+}
+
+// shouldRunValidateNSFSConfig check if we are adding the NsfsAccountConfig or if we are updating it.
+func shouldRunValidateNSFSConfig(na nbv1.NooBaaAccount, oldNa nbv1.NooBaaAccount) bool {
+	return oldNa.Spec.NsfsAccountConfig == nil &&  na.Spec.NsfsAccountConfig != nil ||
```
kfir-payne, line 99: why not just check `reflect.DeepEqual(oldNa.Spec.NsfsAccountConfig, na.Spec.NsfsAccountConfig)`?

build/install-operator-sdk.sh
liranmauda: This is a fix for #697 
Unrelated to this PR, but small so adding it here in another commit.

pkg/admission/validate_noobaaaccount.go
liranmauda: We don't necessarily have `(oldNa.Spec.NsfsAccountConfig` it is optional and can be added (but not removed).
So we want to validate in 2 cases:
1. if we are adding it.
2. if we updating it's fields.

kfir-payne (approved):

PR #876, opened 2022-03-09
bump to Go 1.17, K8s 1.22.2, Rook 1.8.7

umangachapagain:
### Explain the changes
1. Bumping dependencies
Go 1.17
Kubernetes v1.22.2 (v1.23.4 breaks due to other dependencies)
Rook v1.8.7
Controller runtime  v0.10.2 (v0.11.1 breaks due to other dependencies)

2. Removing some replace directives 
3. Excluding breaking client-go releases

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


liranmauda (changes requested):
Hi @umangachapagain Thanks for this PR!
We need to fix the tests, so in all the Github Actions we need to bump the go to 1.17 and change the minikube version to start with the correct kuberentes version

umangachapagain:
@liranmauda Will work on that.
While trying to update generated changes, I found that [pkg/apis/noobaa/v1alpha1/zz_generated.openapi.go](https://github.com/noobaa/noobaa-operator/blob/44c3e7ef6aae161d8150a4cc80f6fa39225ee9d6/pkg/apis/noobaa/v1alpha1/zz_generated.openapi.go) file isn't updated and breaks builds. Can we remove that?

.travis/install-minikube.sh
```diff
@@ -16,8 +16,8 @@ sudo apt-get update && sudo apt-get install socat && sudo apt-get install conntr
 
 MINIKUBE_DEBUG=""
 #MINIKUBE_DEBUG="--alsologtostderr --v=5"
-export MINIKUBE_VERSION=v1.18.1
-export KUBERNETES_VERSION=v1.20.2
+export MINIKUBE_VERSION= v1.23.0
```
liranmauda, line 19:
```suggestion
export MINIKUBE_VERSION=v1.23.0
```

liranmauda:
> @liranmauda Will work on that. While trying to update generated changes, I found that [pkg/apis/noobaa/v1alpha1/zz_generated.openapi.go](https://github.com/noobaa/noobaa-operator/blob/44c3e7ef6aae161d8150a4cc80f6fa39225ee9d6/pkg/apis/noobaa/v1alpha1/zz_generated.openapi.go) file isn't updated and breaks builds. Can we remove that?

I am not familiar with this file purpose. 
@guymguym it seems that openapi generation was dropped by operator sdk a while ago. 
Can we drop it?

go.mod
```diff
@@ -1,87 +1,192 @@
...
+	k8s.io/kube-openapi v0.0.0-20211115234752-e816edb12b65
+	k8s.io/kubectl v0.23.4
 	nhooyr.io/websocket v1.7.4
-	sigs.k8s.io/controller-runtime v0.9.5
-	sigs.k8s.io/yaml v1.2.0
+	sigs.k8s.io/controller-runtime v0.11.1
+	sigs.k8s.io/yaml v1.3.0
+)
+
+require (
```
dannyzaken, line 53: @umangachapagain why does the `require` repeat here?

umangachapagain: 1st require block lists direct dependencies. 2nd require block lists indirect dependencies.
This is done by Go tools automatically.

go.mod
```diff
@@ -1,87 +1,191 @@
...
-	k8s.io/kube-proxy => k8s.io/kube-proxy v0.21.3
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.21.3
-	k8s.io/kubectl => k8s.io/kubectl v0.21.3
-	k8s.io/kubelet => k8s.io/kubelet v0.21.3
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.21.3
-	k8s.io/metrics => k8s.io/metrics v0.21.3
-	k8s.io/mount-utils => k8s.io/mount-utils v0.21.3
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.21.3
-
+	k8s.io/kube-openapi => k8s.io/kube-openapi v0.0.0-20210305001622-591a79e4bda7
```
umangachapagain, line 161: This is a hack to fix openapi generated code. Ideally, we should fix generation code.
Proper fix would probably require using new sdk and other generation tools.

umangachapagain:
Can't reproduce this test error locally. Don't understand where it is coming from.

guymguym (commented):
Does this change our minimum kubernetes version supported by the operator? We declare minKubeVersion 1.16.0 in the olm csv -

https://github.com/noobaa/noobaa-operator/blob/master/deploy/olm/noobaa-operator.clusterserviceversion.yaml

umangachapagain:
> Does this change our minimum kubernetes version supported by the operator? We declare minKubeVersion 1.16.0 in the olm csv -
> 
> https://github.com/noobaa/noobaa-operator/blob/master/deploy/olm/noobaa-operator.clusterserviceversion.yaml

We should definitely switch it to 1.22 at least.
But, the issue in test is related Go version. It says max supported version is Go 1.15 but I can't find where that is defined. And why was it working before with Go 1.16.

liranmauda:
> > Does this change our minimum kubernetes version supported by the operator? We declare minKubeVersion 1.16.0 in the olm csv -
> > https://github.com/noobaa/noobaa-operator/blob/master/deploy/olm/noobaa-operator.clusterserviceversion.yaml
> 
> We should definitely switch it to 1.22 at least. But, the issue in test is related Go version. It says max supported version is Go 1.15 but I can't find where that is defined. And why was it working before with Go 1.16.

It was not with Go 1.16
The directive was set to Go 1.15
The directive states the minimal version of go that is supported, and when the CI tried to use Go 1.15 it allowed it.
The CI used the default Go version, and actually did not test using go 1.16

liranmauda:
@umangachapagain 

This was a tricky one to hunt 😄 
It is because the `actions/setup-go@v2` are setting up the `PATH` for the user, and as we are using here `sudo` it fail

You can replace:
https://github.com/noobaa/noobaa-operator/blob/de553bed293f4b7376e067c16ccaade1c36666f9/.github/workflows/run_ci.yml#L68
with:
`sudo env "PATH=$PATH" make test-olm || exit 1`

liranmauda:
looks like it is now stuck on:
```
++ kubectl get csv noobaa-operator.v5.10.0 -o 'jsonpath={.status.phase}'
+ '[' Pending '!=' Succeeded ']'
+ echo -n .
+ sleep 1
```
@dannyzaken @guymguym 
do you have any idea how to resolve?

umangachapagain:
> looks like it is now stuck on:
> 
> ```
> ++ kubectl get csv noobaa-operator.v5.10.0 -o 'jsonpath={.status.phase}'
> + '[' Pending '!=' Succeeded ']'
> + echo -n .
> + sleep 1
> ```
> 
> @dannyzaken @guymguym do you have any idea how to resolve?

This happened even locally. CSV was never created.

guymguym:
Just a long shot - but does the version of operator-sdk we use supports this new golang version?

umangachapagain:
> Just a long shot - but does the version of operator-sdk we use supports this new golang version?

No. That version of sdk is probably not supported anymore.

dannyzaken:
I'm not sure if operator-sdk is the problem here. 
in any case, operator-sdk is only used in a few make targets and there isn't a real dependency on it. we can try to use the newer versions of operator-sdk for these make targets.

dannyzaken:
> > @liranmauda Will work on that. While trying to update generated changes, I found that [pkg/apis/noobaa/v1alpha1/zz_generated.openapi.go](https://github.com/noobaa/noobaa-operator/blob/44c3e7ef6aae161d8150a4cc80f6fa39225ee9d6/pkg/apis/noobaa/v1alpha1/zz_generated.openapi.go) file isn't updated and breaks builds. Can we remove that?
> 
> I am not familiar with this file purpose. @guymguym it seems that openapi generation was dropped by operator sdk a while ago. Can we drop it?

@umangachapagain I think that `zz_generated.openapi.go` can be removed. this should eliminate the need to `replace` open-api in go.mod

other than that, the tests seem to fail due to OLM issues on k8s ver v1.22.2.
In PR #899 I tried to use a newer OLM version, but it is still not working. we need to investigate that.

Once we solve the OLM tests issues, I think we can merge this PR

umangachapagain:
> @umangachapagain I think that `zz_generated.openapi.go` can be removed. this should eliminate the need to `replace` open-api in go.mod
> 
Created a separate PR for removing it #900 . It removes direct dependency on kube-openapi as expected.
>
> other than that, the tests seem to fail due to OLM issues on k8s ver v1.22.2. In PR #899 I tried to use a newer OLM version, but it is still not working. we need to investigate that.
> 
> Once we solve the OLM tests issues, I think we can merge this PR
>
Agree. Let's merge the other fixes, then we I can update this one.

dannyzaken:
@umangachapagain, let's restore the minikube and k8s versions in `install-minikube.sh` to the way it was before. we can do it in a different PR, and it's not necessarily related to this PR. hope it will resolve the tests failure and we can merge this PR after

https://github.com/noobaa/noobaa-operator/blob/44c3e7ef6aae161d8150a4cc80f6fa39225ee9d6/.travis%2Finstall-minikube.sh#L19-L20

dannyzaken (approved):

PR #877, opened 2022-03-13
Fix noobaaaccount to be able to get also namespacestore as a resource

liranmauda:
### Explain the changes
- Fix noobaaaccount CLI to be able to get also namespacestore as a resource

1. Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2047250

### Testing Instructions:
1. create nsfs namespacestore
```
$ noobaa namespacestore create nsfs fs1 --pvc-name=‘nsfs-vol’ --fs-backend=‘GPFS’
...
INFO[0003] :white_check_mark: NamespaceStore “fs1” Phase is Ready
# NamespaceStore spec:
nsfs:
  fsBackend: GPFS
  pvcName: nsfs-vol
  subPath: “”
type: nsfs

```
2. create nsfs account and see that it is possible to use that  nsfs namespacestore
```
$ noobaa account create --default_resource fs1 --nsfs_account_config --nsfs_only --full_permission --uid 342 --gid 987 testaccount2
...
INFO[0007] :white_check_mark: NooBaaAccount “testaccount2” Phase is Ready
# NooBaaAccount spec:
allow_bucket_creation: true
allowed_buckets:
  full_permission: true
  permission_list: []
default_resource: fs1
nsfs_account_config:
  gid: 987
  new_buckets_path: /
  nsfs_only: true
  uid: 342
INFO[0007] :white_check_mark: Exists: Secret “noobaa-account-testaccount2"
Connection info:
  AWS_ACCESS_KEY_ID      :  **************
  AWS_SECRET_ACCESS_KEY  : **************
```

- [x] Tests added

### DIFF
--- a/pkg/noobaaaccount/noobaaaccount.go
+++ b/pkg/noobaaaccount/noobaaaccount.go
@@ -206,18 +206,18 @@ func RunCreate(cmd *cobra.Command, args []string) {
 	if defaultResource == "" { // if user doesn't provide default resource we will use the default backingstore
 		defaultResource = sys.Name + "-default-backing-store"
 	} 
-	// check that default backing store exists
-	defaultRes := &nbv1.BackingStore{
-		TypeMeta: metav1.TypeMeta{Kind: "BackingStore"},
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      defaultResource,
-			Namespace: options.Namespace,
-		},
+
+	isResourceBackingStore := checkResourceBackingStore(defaultResource) 
+	isResourceNamespaceStore := checkResourceNamespaceStore(defaultResource)
+
+	if isResourceBackingStore && isResourceNamespaceStore {
+		log.Fatalf(`❌  got BackingStore and NamespaceStore %q in namespace %q`,
+			defaultResource, options.Namespace) 
+	} else if !isResourceBackingStore && !isResourceNamespaceStore {
+		log.Fatalf(`❌ Could not get BackingStore or NamespaceStore %q in namespace %q`,
+			defaultResource, options.Namespace)
 	}
-	if !util.KubeCheck(defaultRes) {
-		log.Fatalf(`❌ Could not get BackingStore %q in namespace %q`,
-			defaultResource, noobaaAccount.Namespace)
-	}
+
 	noobaaAccount.Spec.DefaultResource = defaultResource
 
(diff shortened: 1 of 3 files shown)


pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -206,18 +206,17 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      defaultResource,
-			Namespace: options.Namespace,
-		},
-	}
-	if !util.KubeCheck(defaultRes) {
-		log.Fatalf(`❌ Could not get BackingStore %q in namespace %q`,
-			defaultResource, noobaaAccount.Namespace)
+
+	isResourceBackingStore, isResourceNamespaceStore := checkResource(defaultResource)
```
jackyalbo, line 210: I think I would split to 2 functions, and first check if it's backingstore and else check if it's namespacestore. And if he has both with the same name (can it really happen?) we would prefer the backingstore. 
Another option (more work) can be to divide the CRD to defaultBackingStore (like is was originally) and to defaultNSStore and then all of this dual checkup won't be needed - and the user can be specific in what he prefers... WDYT?

test/cli/test_cli_functions.sh
```diff
@@ -650,6 +652,20 @@ function get_admin_password {
     echo ${password}
 }
 
+function account_nsfs_cycle {
```
jackyalbo, line 655: nice! thanks for adding the tests :)

pkg/noobaaaccount/noobaaaccount.go
liranmauda: > I think I would split to 2 functions, and first check if it's backingstore and else check if it's namespacestore. And if he has both with the same name (can it really happen?) we would prefer the backingstore. 

I don't think we should prefer the backingstore as this will change the type of poll the user intended. 
I think we should do PersistentError and let the user handle it. 
for that reason I am not sure we should divide into 2 functions as I will always want to check them both. 
I think it can really happen as they are 2 different kubernetes objects. 

>Another option (more work) can be to divide the CRD to defaultBackingStore (like is was originally) and to defaultNSStore and then all of this dual checkup won't be needed - and the user can be specific in what he prefers... WDYT?

I don't think we should add another resource to the CRD, we don't want to have 2 of them at the same time.
Of course, this can be resolved by using admission control.

jackyalbo (approved):

PR #901, opened 2022-04-10
Log based replication params change

kfir-payne:

### Explain the changes
1. add `LogBasedReplicationInfo` field to `BucketReplicationParams` for support in log based replication required information
2. dependent on https://github.com/noobaa/noobaa-core/pull/6937

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/nb/types.go
+++ b/pkg/nb/types.go
@@ -578,8 +578,14 @@ type UpdateBucketClassParams struct {
 
 // BucketReplicationParams is the params of bucket_api.put_bucket_replication()
 type BucketReplicationParams struct {
-	Name              string        `json:"name"`
-	ReplicationPolicy []interface{} `json:"replication_policy"`
+	Name              string            `json:"name"`
+	ReplicationPolicy ReplicationPolicy `json:"replication_policy"`
+}
+
+// ReplicationPolicy is the struct representing replication configuration
+type ReplicationPolicy struct {
+	Rules              []interface{} `json:"rules,omitempty"`
+	LogReplicationInfo interface{}   `json:"log_replication_info,omitempty"`
 }
 
 // DeleteBucketReplicationParams is the params of bucket_api.delete_bucket_replication()
--- a/pkg/obc/provisioner.go
+++ b/pkg/obc/provisioner.go
@@ -734,14 +734,14 @@ func (r *BucketRequest) prepareReplicationParams(replicationPolicy string, updat
 		return nil, deleteReplicationParams, nil
 	}
 
-	var replicationRules []interface{}
+	var replicationRules nb.ReplicationPolicy
 	err := json.Unmarshal([]byte(replicationPolicy), &replicationRules)
 	if err != nil {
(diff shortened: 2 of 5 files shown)


pkg/nb/types.go
```diff
@@ -578,8 +578,19 @@ type UpdateBucketClassParams struct {
...
+}
+
+// LogsInfo is a struct to define required information for log based replication
+type LogsInfo struct {
+
+	//bucket name of the bucket that holds the logs
+	LogsBucket string `json:"LogsBucket,omitempty"`
+
+	//subpath under the bucket to find the logs
+	SubPath string `json:"maxObjects,omitempty"`
```
romayalon, line 593: fix maxObjects to subpath also for both LogsBucket and SubPath it should be in lower case

pkg/nb/types.go
```diff
@@ -578,8 +578,19 @@ type UpdateBucketClassParams struct {
 
 // BucketReplicationParams is the params of bucket_api.put_bucket_replication()
 type BucketReplicationParams struct {
```
romayalon, line 580: While creating an obc and passing a replication policy through the CLI, we are unmarshaling the json, seems to me that it would break with these changes, right? 
Can you share the input you pass to NooBaa?

romayalon:
I think that in noobaa core you should throw a different rpc error code for invalid logs based replication (in validate_replication) so the error will be better explained and in here add the appropriate handling like we do after calling ValidateReplicationAPI

pkg/nb/types.go
```diff
@@ -578,8 +578,19 @@ type UpdateBucketClassParams struct {
...
-	Name              string        `json:"name"`
-	ReplicationPolicy []interface{} `json:"replication_policy"`
+	Name                    string        `json:"name"`
+	ReplicationPolicy       []interface{} `json:"replication_policy"`
+	LogBasedReplicationInfo LogsInfo      `json:"logs_based_replication_info"`
+}
+
+// LogsInfo is a struct to define required information for log based replication
+type LogsInfo struct {
```
romayalon, line 587: maybe change it to logs location like you have in core?

pkg/nb/types.go
kfir-payne: if I'll insert `log_replication_info` inside `replication_policy` (as @dannyzaken suggested in https://github.com/noobaa/noobaa-core/pull/6937#discussion_r846781560) then it will resolve the unmarshaling issue, WDYT?

romayalon: yes, I think it'll work :)

test/cli/resources/empty-secret.yaml
```diff
@@ -4,3 +4,6 @@ metadata:
   name: empty-secret
 type: Opaque
 stringData:
+
+  AWS_ACCESS_KEY_ID: 9ojWah8Zdp9iZM62a8A3
```
dannyzaken, line 8: ?

kfir-payne: fixed

dannyzaken (approved):

PR #902, opened 2022-04-11
Fixed cli tests not running

liranmauda:
### Explain the changes
1. Fixed cli tests not running


liranmauda:
> I think we will have to change? 🤔
> 
> ```diff
> - if ${CM}
> + if [ $CM == "true" ]
> ```

I don't think so.
test (`[`) is resolving into true or false so if we directly use true/false it works.

guymguym:
Liran I don't think its a good idea to assume that a string must be exactly true/false because `if $CM` will fail to execute if the string in CM is not a known shell command. Since this is a logical true or false value, I would compare to a string as suggested.

PR #903, opened 2022-04-11
Add noobaa admission controller to prevent CR deletion

utkarsh-pro:
Signed-off-by: Utkarsh Srivastava <srivastavautkarsh8097@gmail.com>

### Explain the changes
1. Adds admission controller for Noobaa CR - Prevents resource deletion
2. Preserve the uninstall functionality by altering the WebhookConfiguration before uninstalling.

### Testing Instructions:
1. Install noobaa `noobaa install -n <ns> --admission`
2. Attempt to delete Noobaa CR `kubectl delete noobaas.noobaa.io -n<ns> noobaa` => Request should fail with error.
4. Uninstall noobaa `noobaa uninstall -n <ns>` => Should uninstall noobaa

- [ ] Doc added/updated
- [x] Tests added

PR #905, opened 2022-04-14
Add Noobaa CR cleanup policy

utkarsh-pro:
Signed-off-by: Utkarsh Srivastava <srivastavautkarsh8097@gmail.com>

### Explain the changes
1. Adds `spec.cleanupPolicy.allowNoobaaDeletion` in NooBaa CR which disables the Noobaa CR deletion prevention feature added in https://github.com/noobaa/noobaa-operator/pull/903.
    1. If no policy is specified, it defaults to `false`.
2. Changes the CLI uninstall flow. Now, `noobaa uninstall` just sets `spec.cleanupPolicy.allowNoobaaDeletion` to `true` which triggers noobaa operator to disable the Noobaa CR deletion feature hence allowing the `uninstall` to proceed as usual.

### Testing Instructions:
1. Install noobaa operator: `nb install --admission -n <ns>`
2. Run: `kubectl delete noobaas.noobaa.io noobaa -n <ns>` - should fail

### DIFF
--- a/pkg/apis/noobaa/v1alpha1/noobaa_types.go
+++ b/pkg/apis/noobaa/v1alpha1/noobaa_types.go
@@ -403,6 +403,9 @@ const (
 // CleanupPolicySpec specifies the cleanup policy
 type CleanupPolicySpec struct {
 	Confirmation CleanupConfirmationProperty `json:"confirmation,omitempty"`
+
+	// +optional
+	AllowNoobaaDeletion bool `json:"allowNoobaaDeletion,omitempty"`
 }
 
 // CleanupConfirmationProperty is a string that specifies cleanup confirmation
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -8,6 +8,7 @@ import (
 	"net"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -23,6 +24,7 @@ import (
 	"github.com/noobaa/noobaa-operator/v5/pkg/nb"
 	"github.com/noobaa/noobaa-operator/v5/pkg/options"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+	admissionv1 "k8s.io/api/admissionregistration/v1"
 
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/sirupsen/logrus"
@@ -61,6 +63,9 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 		"noobaa operator started phase 4/4 - \"Configuring\"",
 	)
 
+	if err := r.ReconcileNoobaaDeletionPolicy(); err != nil {
+		return err
+	}
 	if err := r.ReconcileSystemSecrets(); err != nil {
 		return err
 	}
@@ -94,6 +99,7 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 	if err := r.ReconcileDeploymentEndpointStatus(); err != nil {
 		return err
 	}
+
 	return nil
 }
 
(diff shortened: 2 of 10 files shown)


utkarsh-pro:
@romayalon ^

pkg/system/phase4_configuring.go
```diff
@@ -61,6 +63,9 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 		"noobaa operator started phase 4/4 - \"Configuring\"",
 	)
 
+	if err := r.ReconcileNoobaaDeletionPolicy(); err != nil {
```
utkarsh-pro, line 66: @romayalon here is the deletion process. It is part of reconcilation loop.

pkg/system/phase4_configuring.go
```diff
@@ -1523,3 +1529,136 @@ func (r *Reconciler) ReconcileNamespaceStores(namespaceResources []nb.NamespaceR
 	}
 	return nil
 }
+
+// ReconcileNoobaaDeletionPolicy reconciles the deletion policy of a Noobaa resource
+// and configures the noobaa admission controller appropriately.
+func (r *Reconciler) ReconcileNoobaaDeletionPolicy() error {
+	noobaaAdmissionEnabled := os.Getenv("ENABLE_NOOBAA_ADMISSION")
+	if noobaaAdmissionEnabled == "" || strings.ToLower(noobaaAdmissionEnabled) == "false" {
```
romayalon, line 1537: why not use strings.ToLower(noobaaAdmissionEnabled) != "true"?

pkg/system/system.go
```diff
@@ -273,6 +274,20 @@ func RunDelete(cmd *cobra.Command, args []string) {
 
 	log.Infof("Notice: Deletion of External secrets should be preformed manually")
 
+	scopeNamespaced := admissionv1.ScopeType("Namespaced")
+	rule := &admissionv1.RuleWithOperations{
```
romayalon, line 278: can we reuse this code chunk? it appears 3 times

pkg/system/phase4_configuring.go
```diff
@@ -1523,3 +1529,136 @@ func (r *Reconciler) ReconcileNamespaceStores(namespaceResources []nb.NamespaceR
...
+	}
+
+	if r.NooBaa.Spec.CleanupPolicy.AllowNoobaaDeletion {
+		return RemoveRuleFromNoobaaAdmissionWebhook(rule)
+	}
+
+	return AddRuleToNoobaaAdmissionWebhook(rule)
+}
+
+// RemoveRuleFromNoobaaAdmissionWebhook removes the rule from the noobaa internal admission webhook
```
romayalon, line 1559:
```suggestion
// RemoveRuleFromNoobaaAdmissionWebhook removes the rule from the noobaa internal admission webhook if it exists
```

pkg/system/phase4_configuring.go
```diff
@@ -1523,3 +1529,136 @@ func (r *Reconciler) ReconcileNamespaceStores(namespaceResources []nb.NamespaceR
...
+	}
+
+	if !util.KubeUpdate(wc) {
+		return fmt.Errorf("failed to update webhook")
+	}
+
+	return nil
+}
+
+// AddRuleToAdmissionWebhook adds the rule to the noobaa admission webhook
```
romayalon, line 1581:
```suggestion
// AddRuleToNoobaaAdmissionWebhook adds the rule to the noobaa admission webhook if it doesn't exist
```

pkg/system/phase4_configuring.go
utkarsh-pro: Oh, actually that was just because I keep the branches slim 😅 . Have made the change though ✔️ .

pkg/system/system.go
utkarsh-pro: Done ✔️

pkg/system/phase4_configuring.go
romayalon: but ```if strings.ToLower(noobaaAdmissionEnabled) != "true"```  evaluates to true you should return

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -402,7 +402,8 @@ const (
 
 // CleanupPolicySpec specifies the cleanup policy
 type CleanupPolicySpec struct {
-	Confirmation CleanupConfirmationProperty `json:"confirmation,omitempty"`
+	Confirmation        CleanupConfirmationProperty `json:"confirmation,omitempty"`
+	AllowNoobaaDeletion bool                        `json:"allowNoobaaDeletion,omitempty"`
```
romayalon, line 406: please set the new property to +optional?

pkg/system/phase4_configuring.go
```diff
@@ -1523,3 +1529,142 @@ func (r *Reconciler) ReconcileNamespaceStores(namespaceResources []nb.NamespaceR
...
+}
+
+// AddRuleToNoobaaAdmissionWebhook adds the rule to the noobaa admission webhook if it
+// doesn't exists
+func AddRuleToNoobaaAdmissionWebhook(rule *admissionv1.RuleWithOperations) error {
+	wc := &admissionv1.ValidatingWebhookConfiguration{}
+	wc.SetName("admission-validation-webhook")
+
+	// Find the pre-existing admission webhook
+	if _, _, err := util.KubeGet(wc); err != nil {
```
romayalon, line 1580: @kfir-payne is there a situation where wc exists but it's not used by NooBaa? 
I mean, maybe it's used by other component?

kfir-payne: I don't thinks so, the `admission-validation-webhook` supposed to be used only by Noobaa

pkg/system/phase4_configuring.go
```diff
@@ -1523,3 +1529,142 @@ func (r *Reconciler) ReconcileNamespaceStores(namespaceResources []nb.NamespaceR
...
+// if rule was added or else returns false
+//
+// NOTE: It will not add the rule if it exists
+func addAdmissionRule(wc *admissionv1.ValidatingWebhookConfiguration, rule *admissionv1.RuleWithOperations) bool {
+	_, found := findAdmissionRule(wc, rule)
+	if found {
+		return false
+	}
+
+	wc.Webhooks[0].Rules = append(wc.Webhooks[0].Rules, *rule)
```
kfir-payne, line 1622: @utkarsh-pro don't you need to `KubeUpdate(wc)` after appending a new rule?
or after any change done to it?

pkg/system/phase4_configuring.go
kfir-payne: I don't think so, `admission-validation-webhook` supposed to be used only by Noobaa

kfir-payne (approved):

dannyzaken:
This PR blocks the deletion of noobaa CRs by default. I think it is not a good default and might confuse non-ODF users who are not using the CLI.
I suggest changing `allowNoobaaDeletion` flag to something like `guardNoobaaDeletion`, so anyone that wants to get this behavior (e.g. ODF) should set the flag and remove before deletion
@utkarsh-pro @kfir-payne @romayalon WDYT?

right now this change is blocking uninstall flow - https://bugzilla.redhat.com/show_bug.cgi?id=2088587

dannyzaken:
This PR blocks the deletion of noobaa CRs by default. I think it is not a good default and might confuse non-ODF users who are not using the CLI.
I suggest changing `allowNoobaaDeletion` flag to something like `guardNoobaaDeletion`, so anyone that wants to get this behavior (e.g. ODF) should set the flag and remove before deletion
@utkarsh-pro @kfir-payne @romayalon WDYT?

right now this change is blocking uninstall flow - https://bugzilla.redhat.com/show_bug.cgi?id=2088587

liranmauda:
> This PR blocks the deletion of noobaa CRs by default. I think it is not a good default and might confuse non-ODF users who are not using the CLI. I suggest changing `allowNoobaaDeletion` flag to something like `guardNoobaaDeletion`, so anyone that wants to get this behavior (e.g. ODF) should set the flag and remove before deletion @utkarsh-pro @kfir-payne @romayalon WDYT?
> 
> right now this change is blocking uninstall flow - https://bugzilla.redhat.com/show_bug.cgi?id=2088587

Isn't it kind of counter productive? 
if we want to block any of the CRs deletions it should be the default IMHO 
I think that If someone want to delete the CR he should manually and knowingly change the flag in the CR to be true and then he can delete it.

With that said, OCS/ODF operator should have an interface to do it code wise. 

WDYT?

dannyzaken:
IMHO having to set something in a CR to be able to delete it, is a strange default behavior for a kubernetes resource.
in the specific use case of ODF the user shouldn't be aware of it. ocs operator can set it and remove it when needed

romayalon:
@dannyzaken 
Actually this PR https://github.com/noobaa/noobaa-operator/pull/903 blocks the deletion of the CR by default, that's the exact request that we were asked to do in https://bugzilla.redhat.com/show_bug.cgi?id=1981732

romayalon:
I asked Utkarsh to add AllowNooBaaDeletion flag in order to release OCS/ other non-cli installations.

romayalon:
@dannyzaken  The real discussion should be if we want to go with the guard approach regarding the webhook on noobaa CR and when to put it

utkarsh-pro:
> I asked Utkarsh to add AllowNooBaaDeletion flag in order to release OCS/ other non-cli installations. @utkarsh-pro we are supposed to pass this flag in ocs uninstallations, any news about that process?

@romayalon, I have untested changes locally. Testing anything related to OCS has been a bit painful, however by the end of the week we should have a PR up on the OCS operator repository.

utkarsh-pro:
@romayalon https://github.com/red-hat-storage/ocs-operator/pull/1698 is the PR.

PR #907, opened 2022-04-27
create a pv-pool default backing store in azure gov platforms

dannyzaken:

### Explain the changes
* for Azure gov\DOD, creating an Azure default backing store requires a more complicated solution than the one used for public Azure.
* identifying azure gov\dod regions based on the "topology.kubernetes.io/region" label on the nodes.
* when running on azure gov, IsAzurePlatformNonGovernment returns false. as a result, a PV pool backing store is created in phase 4

1. Fix BZ https://bugzilla.redhat.com/show_bug.cgi?id=1944572

### Testing Instructions:
1. deploy noobaa on an Azure gov cluster
2. the deployment should complete successfully and a PV pool default-backingstore should be created 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -535,7 +535,7 @@ func (r *Reconciler) ReconcileBackingStoreCredentials() error {
 	if util.IsAWSPlatform() {
 		return r.ReconcileAWSCredentials()
 	}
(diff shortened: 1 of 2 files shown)


liranmauda (approved):
LGTM

nimrod-becker (approved):

PR #925, opened 2022-05-24
Fixing LC_COLLATE - operator side

jackyalbo:

### Explain the changes
1.  We need to set also collation for the initdb container in order to fix 2068110

1.  Fixed BZ2068110

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/internal/statefulset-postgres-db.yaml
+++ b/deploy/internal/statefulset-postgres-db.yaml
@@ -43,6 +43,8 @@ spec:
         env:
           - name: POSTGRESQL_DATABASE
             value: nbcore
+          - name: LC_COLLATE
+            value: C
           - name: POSTGRESQL_USER
             valueFrom:
               secretKeyRef:
@@ -81,6 +83,8 @@ spec:
         env:
           - name: POSTGRESQL_DATABASE
             value: nbcore
+          - name: LC_COLLATE
+            value: C
           - name: POSTGRESQL_USER
             valueFrom:
               secretKeyRef:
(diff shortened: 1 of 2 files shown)


liranmauda (approved):
LGTM

dannyzaken (changes requested):
@jackyalbo it looks good for new installations, but what will happen with existing systems?
AFAIR you cannot just change the LC_COLLATE env for existing DB, so you should at least ignore this env if it doesn't exist. 
For a complete solution, we need to figure out how we to handle existing DBs

jackyalbo:
> @jackyalbo it looks good for new installations, but what will happen with existing systems? AFAIR you cannot just change the LC_COLLATE env for existing DB. For a partial solution that only handles new systems, you should at least ignore this env in existing systems. For a complete solution, we need to figure out how to handle existing DBs

@dannyzaken that won't effect already created DBs. It is relevant only for new databases. BTW, that's why I added this env to both containers as only the one really creates the DB (the init container) changes the collation, having another container with different collation just won't have any effect (but left for both as a good practice)

jackyalbo:
> @jackyalbo it looks good for new installations, but what will happen with existing systems? AFAIR you cannot just change the LC_COLLATE env for existing DB. For a partial solution that only handles new systems, you should at least ignore this env in existing systems. For a complete solution, we need to figure out how to handle existing DBs

@dannyzaken that won't effect already created DBs. It is relevant only for new databases. BTW, that's why I added this env to both containers as only the one really creates the DB (the init container) changes the collation, having another container with different collation just won't have any effect (but left for both as a good practice)

dannyzaken (approved):

PR #926, opened 2022-05-24
Create service account token if it is missing

baum:
Add a procedure to create noobaa service account token if it is missing


liranmauda (commented):
LGTM

liranmauda (approved):

PR #930, opened 2022-05-30
Adding secretef Webhook For Backingstore and Namespacestore

liranmauda:
### Explain the changes
- Adding secretef Webhook For Backingstore and Namespacestore to validate that it has namespace in it.

Closing the gap from #914 

### Issues:
1. Fixed https://bugzilla.redhat.com/show_bug.cgi?id=2084201

### Testing Instructions:
1. Try to apply a backingstore without a namespace in its secretef 
2. Try to update backingstore to remove the namespace from its secretef 
3. Do the above steps on a namespacestore.

- [ ] Tests added


liranmauda:
@dannyzaken @kfir-payne please take a look.
The PR is in draft so we can look at its logic, once we think it is ok, I will add tests and remove from draft.

kfir-payne (commented):
I think it would be easier to address those comments after a rebase because the structure of calling the validation had changed.
Also tests are missing.

dannyzaken:
@liranmauda I think that for now we can skip the webhook validation for that. let's close

liranmauda:
> @liranmauda I think that for now we can skip the webhook validation for that. let's close

Closing.

PR #937, opened 2022-06-07
Fix set secret owner reference

kfir-payne:

### Explain the changes
1. Setting secret owner reference only if the secret already owned by another noobaa resource, as an indication that secret was created by a noobaa resource and can be deleted when all dependent resources are deleted.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


pkg/backingstore/reconciler.go
```diff
@@ -306,14 +306,24 @@ func (r *Reconciler) LoadBackingStoreSecret() error {
...
-							if !util.KubeUpdate(secret) {
-								return fmt.Errorf("failed to update secret: %q owner reference", r.BackingStore.Name)
+					shouldUpdateOwnerRef := false
+					if secret.ObjectMeta.OwnerReferences != nil {
+						for _, ownerRef := range secret.ObjectMeta.OwnerReferences {
+							if strings.Contains(ownerRef.APIVersion, "noobaa") {
+								shouldUpdateOwnerRef = true
+							}
+						}
+					}
```
dannyzaken, line 316: maybe extract to a common util function?

pkg/util/util.go
```diff
@@ -1961,3 +1961,13 @@ func referSameObject(a, b metav1.OwnerReference) bool {
 
 	return aGV.Group == bGV.Group && a.Kind == b.Kind && a.Name == b.Name
 }
+
+// CheckNoobaaOwnerRefExistence receives an array of owner references and returns true if one of them is of a noobaa resource
+func CheckNoobaaOwnerRefExistence(ownerReferences []metav1.OwnerReference) bool {
```
dannyzaken, line 1966: I suggest to use a cleared name, maybe something like this:
```suggestion
func isOwnedByNoobaa(ownerReferences []metav1.OwnerReference) bool {
```

dannyzaken (commented):
one petty comment. other than that LGTM

dannyzaken (approved):

PR #942, opened 2022-06-19
Support STS credentials for Namespace store

liranmauda:
### Explain the changes
1. Adding support for Namespace STS
2. Adding CLI for creating NS STS
3. Changing the validation to allow STS on NS

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -257,73 +257,75 @@ func (r *Reconciler) ReconcilePhases() error {
 
 // LoadBackingStoreSecret loads the secret to the reconciler struct
 func (r *Reconciler) LoadBackingStoreSecret() error {
-	if !util.IsSTSClusterBS(r.BackingStore) {
-		secretRef, err := util.GetBackingStoreSecret(r.BackingStore)
-		if err != nil {
-			return err
-		}
+	if util.IsSTSClusterBS(r.BackingStore) {
+		return nil
+	}
+	
+	secretRef, err := util.GetBackingStoreSecret(r.BackingStore)
+	if err != nil {
+		return err
+	}
 
-		if secretRef != nil {
-			r.Secret.Name = secretRef.Name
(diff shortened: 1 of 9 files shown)


liranmauda:
This PR is still in draft as it was not tested.
@ezio-auditore can you also review?

pkg/backingstore/reconciler.go
```diff
@@ -257,7 +257,7 @@ func (r *Reconciler) ReconcilePhases() error {
 
 // LoadBackingStoreSecret loads the secret to the reconciler struct
 func (r *Reconciler) LoadBackingStoreSecret() error {
-	if !util.IsSTSClusterBS(r.BackingStore) {
+	if !util.IsSTSClusterBS(r.BackingStore) {  //LMLM TODO: why is this negative condition, why not return nill if it is true?
```
ezio-auditore, line 260: If it is not an sts cluster and the secret is also not found it should return an error, And if it is an STS cluster it should not check for the secret.

ezio-auditore (commented):
Besides some minor comments, LGTM

pkg/backingstore/reconciler.go
liranmauda: The thing with negative conditions is that it is harder for humans to understand... 

The translation to a positive statement is in the second half of your sentence: 
> if it is an STS cluster it should not check for the secret.
Which can in turn be written like:
```
	if util.IsSTSClusterBS(r.BackingStore) {
		return nil
	}
```
@ezio-auditore  WDYT?

ezio-auditore: ok, makes sense.

dannyzaken:
pending on core changes and @liranmauda is trying to set up an env

liranmauda:
> pending on core changes and @liranmauda is trying to set up an env

https://github.com/noobaa/noobaa-core/pull/7074
@dannyzaken @romayalon Please review.

jackyalbo (approved):

Issue #683, opened 2021-07-08
Default backingstore has no secret


igorpick:
### Environment info
- NooBaa Operator Version: **VERSION**
- Platform: **Kubernetes 1.14.1 | minikube 1.1.1 | OpenShift 4.1 | other: specify**

### Actual behavior
1. The default backingstore noobaa-default-backing-store has no a secret configuration.
2. It causes to panic message in noobaa cli when getting the backingstore status.

### Expected behavior
1.  The default backingstore is configured with secret backing-store-pv-pool-noobaa-default-backing-store name

### Steps to reproduce
1. noobaa install
2. noobaa backingstore status noobaa-default-backing-store

### More information - Screenshots / Logs / Other output
```
$kubectl get backingstores.noobaa.io/noobaa-default-backing-store -o yaml

kind: BackingStore
metadata:
  creationTimestamp: "2021-07-05T07:55:06Z"
  finalizers:
  - noobaa.io/finalizer
  generation: 1
  labels:
    app: noobaa
  name: noobaa-default-backing-store
  namespace: noobaa
  ownerReferences:
  - apiVersion: noobaa.io/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: NooBaa
    name: noobaa
    uid: 92b832fc-0001-4570-bba8-ee8b0c2937f5
  resourceVersion: "215684"
  uid: c488a9db-8321-4a8b-a9f4-df2774ce9a5d
spec:
  pvPool:
    numVolumes: 1
    resources:
      requests:
        storage: 50Gi
    secret: {}
  type: pv-pool
status:
  conditions:
  - lastHeartbeatTime: "2021-07-05T07:55:06Z"
    lastTransitionTime: "2021-07-08T11:55:38Z"
    message: BackingStorePhaseReady
    reason: 'Backing store mode: LOW_CAPACITY'
    status: "True"
    type: Available
  - lastHeartbeatTime: "2021-07-05T07:55:06Z"
    lastTransitionTime: "2021-07-08T11:55:38Z"
    message: BackingStorePhaseReady
    reason: 'Backing store mode: LOW_CAPACITY'
    status: "False"
    type: Progressing
  - lastHeartbeatTime: "2021-07-05T07:55:06Z"
    lastTransitionTime: "2021-07-08T11:55:37Z"
    message: BackingStorePhaseReady
    reason: 'Backing store mode: LOW_CAPACITY'
    status: "False"
    type: Degraded
[...]

PR #949, opened 2022-06-24
Fixed default secret for backingstore

nik-redhat:
### Explain the changes
1. Set the default secret as a combination of backingstore spec type and
name for default backing store.

Fixes: #683 

### Testing Instructions:
1. Run the `noobaa backingstore status noobaa-default-backing-store` command, it won't panic.

Signed-off-by: nik-redhat <nladha@redhat.com>

### DIFF
--- a/pkg/backingstore/backingstore.go
+++ b/pkg/backingstore/backingstore.go
@@ -892,7 +892,7 @@ func RunStatus(cmd *cobra.Command, args []string) {
 			if secret.Namespace == "" {
 				secret.Namespace = backStore.Namespace
 			}
-			if !util.KubeCheck(secret) {
+			if backStore.Spec.Type != nbv1.StoreTypePVPool && !util.KubeCheck(secret) {
 				log.Errorf(`❌ Could not get Secret %q in namespace %q`,
 					secret.Name, secret.Namespace)
 			}
@@ -907,7 +907,7 @@ func RunStatus(cmd *cobra.Command, args []string) {
 	util.Panic(err)
 	fmt.Print(string(output))
 	fmt.Println()
-	if secretRef != nil {
+	if secretRef != nil && secret.Name != "" {
 		_, err = sigyaml.Marshal(secret.StringData)
 		util.Panic(err)
 		fmt.Println()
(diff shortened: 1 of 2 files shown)


nik-redhat:
Requesting for review.

nik-redhat:
@dannyzaken can you please review this PR?

pkg/backingstore/backingstore.go
```diff
@@ -892,6 +892,9 @@ func RunStatus(cmd *cobra.Command, args []string) {
 			if secret.Namespace == "" {
 				secret.Namespace = backStore.Namespace
 			}
+			if secret.Name == "" {
```
jackyalbo, line 895: Just think that maybe on pvPool we should skip printing the secret instead of this fix - as it not really relevant for the user.
if backStore.Spec.Type == nbv1.StoreTypePVPool -> skip secret print

jackyalbo (changes requested):

pkg/backingstore/backingstore.go
nik-redhat: Ok.
Hopefully, the update should do the trick.

liranmauda:
Hi @nik-redhat 
Please squash the commits, and rebase.

jackyalbo:
We are also using this secretref later in the code (and it seems that it's not null). Can you also fix this part as well?

nik-redhat:
> We are also using this secretref later in the code (and it seems that it's not null). Can you also fix this part as well? line 910

Yes, it is not null. But, I don't foresee any issue in that code. It is only converting to JSON, if the value is empty it won't mind while converting. Or, am I missing something here?

nik-redhat:
> Hi @nik-redhat Please rebase, and squash the commits.

Done

jackyalbo:
> > We are also using this secretref later in the code (and it seems that it's not null). Can you also fix this part as well? line 910
> 
> Yes, it is not null. But, I don't foresee any issue in that code. It is only converting to JSON, if the value is empty it won't mind while converting. Or, am I missing something here?

 I just think we should avoid that code. No need to call print or JSON marshal on something we are not interested in printing - like this empty object. Maybe just add a check about the secret.Name...

nik-redhat:
> > > We are also using this secretref later in the code (and it seems that it's not null). Can you also fix this part as well? line 910
> > 
> > 
> > Yes, it is not null. But, I don't foresee any issue in that code. It is only converting to JSON, if the value is empty it won't mind while converting. Or, am I missing something here?
> 
> I just think we should avoid that code. No need to call print or JSON marshal on something we are not interested in printing - like this empty object. Maybe just add a check about the secret.Name...

Ack, made the update.

jackyalbo (approved):
LGTM, just rebase your code so we can push and close the issue. Thanks

PR #953, opened 2022-06-28
Fix check external connection persistent error

kfir-payne:

### Explain the changes
1. Throw regular error when INVALID_SCHEMA_PARAMS error returned instead of a persistent error.
2. resolve a situation when the api params don't match due to an upgrade.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


liranmauda (commented):
Why do we want regular error? 
I think we better have a Persistent error so we will find those in QE tests and fix the `INVALID_SCHEMA_PARAMS` 
***Edit***
I do understand that in  a situation where the operator is in newer version and the core did not upgrade yet we want not to stuck the operator, But I wanted to verify that this is the case... 
@kfir-payne Am I understand correctly?

liranmauda (approved):

PR #955, opened 2022-06-30
Remove RPC error type check

kfir-payne:

### Explain the changes
1. remove RPC error type check returned from check_external_connection call, any RPC error should trigger a temporary error and there is no need to check if it's a `INVALID_SCHEMA_PARAMS` error.

1. https://bugzilla.redhat.com/show_bug.cgi?id=2101380

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


pkg/backingstore/reconciler.go
```diff
@@ -925,9 +925,7 @@ func (r *Reconciler) CheckExternalConnection(connInfo *nb.CheckExternalConnectio
 	res, err := r.NBClient.CheckExternalConnectionAPI(*connInfo)
 	if err != nil {
 		if rpcErr, isRPCErr := err.(*nb.RPCError); isRPCErr {
```
dannyzaken, line 927: I think you can just remove this as well

PR #970, opened 2022-08-10
cli: Print on newline, if more than one value

nik-redhat:
Signed-off-by: nik-redhat <nladha@redhat.com>

### Explain the changes
1. Added a new util function `PrettyPrint` to print each item of a string array in newline if the length is more than 1.

### Issues:
1. Fixed #267

PR #964, opened 2022-07-20
KMIP KMS module

baum:
# KMIP KMS module

## Implementation
- This module uses [github.com/gemalto/kmip-go](https://github.com/ThalesGroup/kmip-go) for KMIP client code and requires the [PR](https://github.com/ThalesGroup/kmip-go/pull/19).
- Initial integration test using CipherTrust as a KMIP server is [here](https://github.com/baum/kmip-go/blob/master/ciphertrust_test.go)
- To store the secret data use the KMIP protocol  [Register Operation](https://docs.oasis-open.org/kmip/kmip-spec/v2.0/os/kmip-spec-v2.0-os.html#_Toc6497565) - _“This operation requests the server to register a Managed Object that was created by the client or obtained by the client through some other means, allowing the server to manage the object.”_
- To fetch the stored secret data use the KMIP protocol  [Get Operation](https://docs.oasis-open.org/kmip/kmip-spec/v2.0/os/kmip-spec-v2.0-os.html#_Toc6497545) - _"This operation requests that the server returns the Managed Object specified by its Unique Identifier."_
- To delete the stored secret data use the KMIP protocol [Destroy Operation](https://docs.oasis-open.org/kmip/kmip-spec/v2.0/os/kmip-spec-v2.0-os.html#_Toc6497541) - _"This operation is used to indicate to the server that the key material for the specified Managed Object SHALL be destroyed or rendered inaccessible. "_
- The KMIP Unique Identifier of stored secret data is persisted locally in Certificates secret.

### KMS YAML
```yaml
  security:
    kms:
      connectionDetails:
        KMIP_ENDPOINT: kmip:5696
        KMS_PROVIDER: kmip
        TLS_SERVER_NAME: kmip.ciphertrustmanager.local
      tokenSecretName: kms-kmip-certs
```

### Connection details keys:
| Key      | Type |Description |
| --------------------- | --------------------- |------------------- |
| **KMS_PROVIDER**          | _Mandatory_        | Use `kmip` value   |
| **KMIP_ENDPOINT**         | _Mandatory_       | KMIP endpoint address   |
| **TLS_SERVER_NAME**   | _Optional_           | The endpoint server name. Useful when the KMIP endpoint does not have a DNS entry   |
| **READ_TIMEOUT**          | _Optional_           | Network read timeout, in seconds. The default value is 10   |
| **WRITE_TIMEOUT**        | _Optional_           | Network write timeout, in seconds. The default value is 10  |

### Certificates secret
Should include the following keys:
[...]

PR #811, opened 2021-12-08
RPC WS refactor

baum:
RPC WS refactor

    - Use channel/goroutine for sending messages
    - Rework connect/reconnect logic
    - Remove WS specific logic from common RPC layer

# Description

## Abstract
[RPCConnWS](https://github.com/baum/noobaa-operator/blob/906f04ba735675b26fe7e2090f91b2646aebd41d/pkg/nb/rpc_ws.go#L34-L47) implements [RPCConn](https://github.com/baum/noobaa-operator/blob/906f04ba735675b26fe7e2090f91b2646aebd41d/pkg/nb/rpc.go#L47-L54) interface used by the RPC layer in terms of [nhooyr.io/websocket](https://github.com/nhooyr/websocket) transport. RPC messages are dispatched according to address URL protocol in [GetConnection()](https://github.com/baum/noobaa-operator/blob/906f04ba735675b26fe7e2090f91b2646aebd41d/pkg/nb/rpc.go#L182-L192).

## States
RPCConnWS is a WebSocket connection that is shared and multiplexed
 for all concurrent requests to the same address.  An RPCConnWS instance is [constructed](https://github.com/baum/noobaa-operator/blob/906f04ba735675b26fe7e2090f91b2646aebd41d/pkg/nb/rpc_ws.go#L57-L98) in a connected state and ready for executing RPC Calls.  `nil` is returned if connection refused or within [reconnect delay](https://github.com/baum/noobaa-operator/blob/906f04ba735675b26fe7e2090f91b2646aebd41d/pkg/nb/rpc_ws.go#L20). RPCConnWS detects peer disconnect on send and receives paths enhanced by heartbeat Ping goroutine and closes gracefully.

### [Connection, reconnection map](https://github.com/baum/noobaa-operator/blob/906f04ba735675b26fe7e2090f91b2646aebd41d/pkg/nb/rpc_ws.go#L25-L31)
[NewRPCConnWS](https://github.com/baum/noobaa-operator/blob/906f04ba735675b26fe7e2090f91b2646aebd41d/pkg/nb/rpc_ws.go#L57-L98) uses connection and reconnection tables mapped by the connection's address URL. The state shared by the construction and closing logic is guarded by the mutex, unlocking is insured by deferring unlock.

### Threads
[...]

PR #971, opened 2022-08-15
golangci lint 1.48.0

baum:
Several PRs are failing, see for instance https://github.com/noobaa/noobaa-operator/pull/970, https://github.com/noobaa/noobaa-operator/pull/964

- Bump golangci lint to 1.48.0
- Handle ioutil deprecation

### DIFF
--- a/pkg/admission/validator.go
+++ b/pkg/admission/validator.go
@@ -3,7 +3,7 @@ package admission
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http"
 
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
@@ -28,7 +28,7 @@ func (gs *ServerHandler) serve(w http.ResponseWriter, r *http.Request) {
 	var arResponse admissionv1.AdmissionReview
 	var body []byte
 	if r.Body != nil {
-		if data, err := ioutil.ReadAll(r.Body); err == nil {
+		if data, err := io.ReadAll(r.Body); err == nil {
 			body = data
 		}
 	}
--- a/pkg/nb/rpc_ws.go
+++ b/pkg/nb/rpc_ws.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"io/ioutil"
 	"sync"
 	"time"
 
@@ -112,7 +111,7 @@ func (c *RPCConnWS) ConnectUnderLock() error {
 	}
 
 	logrus.Infof("RPC: Connecting websocket (%p) %+v", c, c)
-	dialCtx, dialCancel := context.WithTimeout(context.Background(), connectTimeout) 
+	dialCtx, dialCancel := context.WithTimeout(context.Background(), connectTimeout)
 	defer dialCancel()
 	ws, _, err := websocket.Dial(dialCtx, c.Address, &websocket.DialOptions{HTTPClient: &c.RPC.HTTPClient})
 	if err != nil {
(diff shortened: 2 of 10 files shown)


pkg/admission/validator.go
```diff
@@ -28,7 +28,7 @@ func (gs *ServerHandler) serve(w http.ResponseWriter, r *http.Request) {
 	var arResponse admissionv1.AdmissionReview
 	var body []byte
 	if r.Body != nil {
-		if data, err := ioutil.ReadAll(r.Body); err == nil {
+		if data, err := io.ReadAll(r.Body); err == nil {
```
utkarsh-pro, line 31: @baum apologies, this isn't related to this PR just caught my attention - I think we are missing `defer r.body.Close()` somewhere here.

pkg/nb/rpc_ws.go
```diff
@@ -321,7 +320,7 @@ func (c *RPCConnWS) ReadMessage() (*RPCMessage, error) {
 
 	// Read message buffers if any
 	if msg.Buffers != nil && len(msg.Buffers) > 0 {
-		buffers, err := ioutil.ReadAll(reader)
+		buffers, err := io.ReadAll(reader)
```
utkarsh-pro, line 323: Not sure but I think we are missing to close the fd here as well 🤔

pkg/util/kms/kms_vault.go
```diff
@@ -153,13 +152,13 @@ func writeCrtsToFile(secretName string, namespace string, secretValue []byte, en
 	}
 
 	// Generate a temp file
-	file, err := ioutil.TempFile("", "")
+	file, err := os.CreateTemp("", "")
```
utkarsh-pro, line 155: I think here also we are missing `defer file.Close()`, does that sound right @baum?

pkg/admission/validator.go
baum: @utkarsh-pro thanks for the review.🖖. 

```
The Server will close the request body.
The ServeHTTP  Handler does not need to.
```
See [lines 176-177 in net/http/request.go](https://go.dev/src/net/http/request.go)

That is why I think we're OK here. @kfir-payne WDYT?

pkg/util/kms/kms_vault.go
baum: Great catch, thank you @utkarsh-pro 😀

[Amended](https://github.com/noobaa/noobaa-operator/pull/971/commits/04dca9c6b64338b6de0279cb8ab55289cb03dfaf).

pkg/admission/validator.go
kfir-payne: I think that @baum is right here and we're OK.

pkg/nb/rpc_ws.go
baum: Looks like we're OK here. The WS  RPC logic [IMHO could be crispier 😉 ](https://github.com/noobaa/noobaa-operator/pull/811) however, we don't necessarily want to close the WebSocket connection after reading each message, since the client might want to send us another one. The WS.connection is closed [here](https://github.com/noobaa/noobaa-operator/blob/60628671e4f05f55b49af1ab6d58ff689d1536d0/pkg/nb/rpc_ws.go#L162-L167). WDYT @utkarsh-pro ?

pkg/admission/validator.go
utkarsh-pro: Ah, I didn't know that. Now I got the rethink every `defer req.Body.Close()` that I have written 😆 . Thank you for pointing to the source @baum 🙌 .

pkg/nb/rpc_ws.go
utkarsh-pro: Oh right, that makes sense. Thank you! 😄

utkarsh-pro (approved):
LGTM 🚀 .

PR #977, opened 2022-08-31
Support OCP injected CA

jackyalbo:

### Explain the changes
following this documentation: https://docs.openshift.com/container-platform/4.11/networking/configuring-a-custom-pki.html
1. Add support for OCP CA Inject in case we have namespace openshift-config in the cluster
2. It will create a config map name noobaa-inject-ca, which OCP will inject with system supported CA
3. It will attach this config map under "/etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem"

1. Fixed https://bugzilla.redhat.com/show_bug.cgi?id=2121104

### Testing Instructions:
1. Add a backingstore to a compatible S3 which uses a self sign cert.
2. Install the self signed certificate to OCP (following above doc)
3. see that there is no issue with self signed certs when creating backingstore, and when trying to write to it (using OBC) 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/internal/configmap-ca-inject.yaml
+++ b/deploy/internal/configmap-ca-inject.yaml
@@ -1,0 +1,7 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  labels:
+    config.openshift.io/inject-trusted-cabundle: "true"
+  name: noobaa-ca-inject
+data: {}
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -3153,6 +3153,17 @@ spec:
     skipServiceCheck: true
 `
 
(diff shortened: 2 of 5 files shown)


jackyalbo:
> Looks nice and tight @jackyalbo
> 
> Could some simple integration test ( see [the admission test](https://github.com/noobaa/noobaa-operator/blob/master/.github/workflows/run_admission_test.yml) for instance ) be added, that verifies that the injected certificate appears in the pod's filesystem?

This feature is specifically used for OCP, so not sure if we want to add it to our general testing. I agree that is needed to be properly tested and QE suppose to take care of it once merged.

jackyalbo:
> This is a good idea to run nighly on a deployed OCP along other tests, during PR tests we run on minik and kind

let's push and add test when we will have this setup ready. There is a costumer waiting for this. QE should check all the procedure of course.

baum (approved):
lgtm 🖖

PR #979, opened 2022-09-12
🪳 Move status account and services instances creation upon first reference

baum:
Move status account and services instances creation upon first referenced

See https://bugzilla.redhat.com/show_bug.cgi?id=2125775 
According to the bug, the status services could be nil upon initial reference. Refactor the code to ensure that the status account and services instances are created upon initial reference.


dannyzaken:
@baum I think that this was fixed in this PR https://github.com/noobaa/noobaa-operator/pull/939
can you check with @kfir-payne if it's the same issue? #939 is not part of 5.11 which the BZ refer to

baum:
Handled by PR https://github.com/noobaa/noobaa-operator/pull/939

PR #987, opened 2022-09-20
Supporting WARN debug level in operator

jackyalbo:

### Explain the changes
1.  We will use same architecture we have for core for operator. 2 levels:
 warn - to print just warning and up log levels
 default_level - to print all of the logs like we did before
2.  fixes to use the correct log print for cli commands and for syslog

1. Fixed https://bugzilla.redhat.com/show_bug.cgi?id=1932846 (last part needed - operator, after already taking care of core and endpoint)

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


liranmauda (approved):
LGTM

PR #996, opened 2022-11-16
Map alternate keys value

baum:
Fix util.CheckForIdenticalSecretsCreds() for alternative key names

This issue was noted at #992, according to [this comment](https://github.com/noobaa/noobaa-operator/pull/992#issuecomment-1315069219), util.CheckForIdenticalSecretsCreds() uses an empty string `""` as a value for the backing store secret key & id.
Refactor an existing functionality.

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -260,7 +260,7 @@ func (r *Reconciler) LoadBackingStoreSecret() error {
 	if util.IsSTSClusterBS(r.BackingStore) {
 		return nil
 	}
-	
+
 	secretRef, err := util.GetBackingStoreSecret(r.BackingStore)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


liranmauda (commented):
There are 2 commits here that seem to add something and then revert it.
If this is not needed, then LGTM.
Please squash the commits.

romayalon:
sounds good, we should add a test :)

romayalon:
I meant for a test that includes 2 secrets, the first with lower case keys and the second with upper case keys and to check if they match
we need to add a lot tests there as we saw many issues in that area, worth documenting it at least

baum:
>  ... the first with lower case keys and the second with upper case keys and to check if they match

Verified with [this procedure](https://github.com/noobaa/noobaa-operator/pull/992#issuecomment-1315069219), that util.CheckForIdenticalSecretsCreds() uses the correct string as a value for the backing store secret key & id while comparing.

romayalon (approved):

PR #998, opened 2022-12-01
Add workflow for automated NooBaa releases

utkarsh-pro:
Signed-off-by: Utkarsh Srivastava <srivastavautkarsh8097@gmail.com>

### Explain the changes
This PR attempts to provide a Github workflow which can be used to automate:
1. Updating `version/version.go`.
2. Updating `options.go`.
3. Update README.md.
4. Release of NooBaa CLI binaries for macos and linux platform (amd64 and arm64).
5. Release of NooBaa operator docker image to dockerhub and quay.
6. Release of NooBaa CLI homebrew formula.
7. Release of krew plugin (disabled in this PR, krew requires a manual first PR and then it can be enabled here).
8. Release of operator on Operator Hub

image

Flow:
1. Workflow will first read the `version/version.go` to figure out which version is supposed to be built and released.
2. README.md is updated accordingly and push the changes to the base branch.
3. Build assets like: CLI binaries, Homebrew file, krew file, OCI images, OLM manifests.
4. Create release on Github.
5. Push the homebrew file to the given homebrew repositories.
6. Create a PR against the `OPERATOR_HUB_REPO` (== "k8s-operatorhub/community-operators") with the new version of `CSV`.
7. Save the .krew.yaml in the current tree to be utilized by the krew-release-bot.
8. Increment the patch version in `version/version.go` and `options.go`
9. Commit the changes and push to the base branch.

Caveats:
- The entire workflow assumes that the version in the `version/version.go` is correct. It does not have additional logic to ensure that. Whatever the version would be in that file, it will be released.
- Rollbacks are manual for now. That is, delete the releases, delete the tag, close the PR, revert the commits. Can be automated in the future though.
- Whenever a new version branch is created, we need to ensure that `version/version.go` reflects that. So when we move to `5.14` branch, `version/version.go` needs to be set to `Version = "5.14.0"`. This can be automated as well but not sure if its worth doing so.

The scripts/workflow will require a few secrets in place like:
1. `GHACTION_GH_PAT` - A github token which has write access to the repositories. This is needed because the workflow updates certain files and then runs `make gen` one last time and commit the changes to the branch. This is also required to create a new release and upload release assets.
11. `DOCKERHUB_USERNAME` - Username in dockerhub where the OCI image is supposed to be pushed.
12. `DOCKERHUB_TOKEN` - Dockerhub token is used to login to the registry and push the image.
13. `QUAY_USERNAME` - Username in quay where the OCI image is supposed to be pushed.
14. `QUAY_TOKEN` - Quay token is used to login to the registry and push the image.
15. `HOMEBREW_CORE_REPO` - This is the repository where we store our hombrew formulas (format: `org/repo`).
16. `OCI_ORG` - This is used internally hence this can be anything. It can be same as `DOCKERHUB_USERNAME` or `QUAY_USERNAME` or a random string.

### DIFF
--- a/.github/workflows/releaser.yaml
+++ b/.github/workflows/releaser.yaml
@@ -1,0 +1,47 @@
+name: Releaser
+
+on:
+  workflow_dispatch:
+    inputs:
+      base_branch:
+        description: 'The base branch to release from'
+        required: true 
+
+permissions:
+  contents: write
+
+jobs:
+  releaser:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Set branch
+        run: echo "BRANCH=${{ github.event.inputs.base_branch }}" >> $GITHUB_ENV
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ env.BRANCH }}
+      - name: Fetch all tags
+        run: git fetch --force --tags
+      - name: Set up Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.17
+      - name: Release
+        env:
+          GITHUB_TOKEN: ${{ secrets.GHACTION_GH_PAT }}
+          DOCKERHUB_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
+          DOCKERHUB_TOKEN: ${{ secrets.DOCKERHUB_TOKEN }}
+          QUAY_USERNAME: ${{ secrets.QUAY_USERNAME }}
+          QUAY_TOKEN: ${{ secrets.QUAY_TOKEN }}
+          HOMEBREW_CORE_REPO: ${{ secrets.HOMEBREW_CORE_REPO }}
+          OCI_ORG: ${{ secrets.OCI_ORG }}
+          OPERATOR_HUB_REPO: "k8s-operatorhub/community-operators"
+          TAG: "1"
+        run: |
+          git config --global user.email "github-action@noobaa.io"
+          git config --global user.name "NooBaa GitHub Action"
+
+          bash build/tools/builder.sh --oci-org $OCI_ORG --gh-org noobaa --gh-repo noobaa-operator || exit 1
+          bash build/tools/releaser.sh --oci-org $OCI_ORG --gh-org noobaa --gh-repo noobaa-operator || exit 1
+      # - name: "Krew Release"
+      #   uses: rajatjindal/krew-release-bot@v0.0.40
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -3,12 +3,13 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 ENV OPERATOR=/usr/local/bin/noobaa-operator \
     USER_UID=1001 \
     USER_NAME=noobaa-operator
+ARG NOOBAA_BIN_PATH=build/_output/bin/noobaa-operator
 
 # tar is needed for kubectl cp
 RUN microdnf install tar
 
 # install operator binary
-COPY build/_output/bin/noobaa-operator ${OPERATOR}
+COPY ${NOOBAA_BIN_PATH} ${OPERATOR}
 
 # copy scripts
 COPY build/bin /usr/local/bin
--- a/build/tools/builder-pre.sh
+++ b/build/tools/builder-pre.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+# NOTE: This script can be run manually but is run anyway by builder.sh before that script starts executing its `main`
+
+# This script will automate the file alteration, which involves
+# 1. Updating the README.md
+# 2. Committing the changes and pushing the changes
+
+set -eux
+set -o pipefail
+
+export PS4='\e[36m+ ${FUNCNAME:-main}@${BASH_SOURCE}:${LINENO} \e[0m'
+
+dir=$(dirname "$0")
+
+# Source the utilities
+source $dir/utils.sh
+
+# Variables
+DRY_RUN=${DRY_RUN:="false"}
+
+# Update the version of noobaa core image, noobaa operator image and CLI version in the README output.
+function update_readme() {
+	local version=$(get_noobaa_version)
+	local core_version="$version" # Assume to be the same as the version of operator
+	local readme_file=README.md
+
+	finline_replace "INFO\[0000\] CLI version: .*" "INFO\[0000\] CLI version: ${version}" $readme_file
+	finline_replace "INFO\[0000\] noobaa-image: .*" "INFO\[0000\] noobaa-image: noobaa\/noobaa-core:${core_version}" $readme_file
+	finline_replace "INFO\[0000\] operator-image: .*" "INFO\[0000\] operator-image: noobaa\/noobaa-operator:${version}" $readme_file
+}
+
+function commit_changes() {
+	if [[ $DRY_RUN == "true" ]]; then
+		echo "DRY_RUN is set to true, skipping commiting changes."
+		return
+	fi
+
+	local version=$(get_noobaa_version 1)
+
+	git add .
+	git commit -m "Automated commit to update README for version: ${version}"
+	git push
+
+	# If TAG=1 is provided, then create a tag
+	if [ -n "$TAG" ]; then
+		git tag -a "${version}" -m "Tag for version ${version}"
+		git push origin "${version}"
+	fi
+}
+
+# Main function
+function main() {
+	update_readme
+	commit_changes
+}
+
+main "$@"
--- a/build/tools/builder.sh
+++ b/build/tools/builder.sh
@@ -1,0 +1,238 @@
+#!/bin/bash
+
+# This script is supposed to automate various types of artifact generation that this project has
+# to do.
+# Following types of artifacts are supported:
+# 1. Build a new version of the CLI
+# 2. Build a new version of OCI image
+# 3. Generate Krew manifest
+# 4. Generate Homebrew formula
+
+set -eux
+set -o pipefail
+
+export PS4='\e[36m+ ${FUNCNAME:-main}@${BASH_SOURCE}:${LINENO} \e[0m'
+
+dir=$(dirname "$0")
+
+# Source the utilities
+source $dir/utils.sh
+
+bin_path=$dir/artifacts/bin
+krew_path=$dir/artifacts/krew
+homebrew_path=$dir/artifacts/homebrew
+operator_bundle_path=$dir/artifacts/operator-bundle
+dependencies=(go docker make sha256sum git)
+GH_ORG="noobaa"
+GH_REPO="noobaa-operator"
+DRY_RUN="false"
+
+function parse_args() {
+  # Parse command-line arguments.
+  while [ "$#" -gt 0 ]; do
+  case "$1" in
+    --gh-org)
+      GH_ORG="$2"
+      shift
+      ;;
+    --gh-repo)
+      GH_REPO="$2"
+      shift
+      ;;
+    --oci-org)
+      OCI_ORG="$2"
+      shift
+      ;;
+    --dry-run)
+      DRY_RUN="true"
+      ;;
+    -h|--help)
+      usage
+      exit 0
+      ;;
+    *)
+      usage
+      exit 1
+      ;;
+  esac
+
+  shift
+  done
+}
+
+function init() {
+  mkdir -p $bin_path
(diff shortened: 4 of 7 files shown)


liranmauda:
@utkarsh-pro before we merge this PR we need to align all the z versions in the branches.

.github/workflows/releaser.yaml
```diff
@@ -0,0 +1,47 @@
...
+          OPERATOR_HUB_REPO: "k8s-operatorhub/community-operators"
+          TAG: "1"
+        run: |
+          git config --global user.email "github-action@noobaa.io"
+          git config --global user.name "NooBaa GitHub Action"
+
+          bash build/tools/builder.sh --oci-org $OCI_ORG --gh-org noobaa --gh-repo noobaa-operator || exit 1
+          bash build/tools/releaser.sh --oci-org $OCI_ORG --gh-org noobaa --gh-repo noobaa-operator || exit 1
+      # - name: "Krew Release"
+      #   uses: rajatjindal/krew-release-bot@v0.0.40
```
liranmauda, line 47: Why do we need those comments? 
If we will not use it, we can remove it

build/Dockerfile
```diff
@@ -3,12 +3,13 @@ FROM registry.access.redhat.com/ubi8/ubi-minimal:latest
 ENV OPERATOR=/usr/local/bin/noobaa-operator \
     USER_UID=1001 \
     USER_NAME=noobaa-operator
+ARG NOOBAA_BIN_PATH=build/_output/bin/noobaa-operator
```
liranmauda, line 6: Why ARG and not ENV? 
Are we intending to provide a different path at some point?

build/tools/builder-pre.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# NOTE: This script can be run manually but is run anyway by builder.sh before that script starts executing its `main`
+
+# This script will automate the file alteration, which involves
+# 1. Updating the README.md
+# 2. Commiting the changes and pushing the changes
```
liranmauda, line 7:
```suggestion
# 2. Committing the changes and pushing the changes
```

build/tools/builder-pre.sh
```diff
@@ -0,0 +1,56 @@
...
+
+dir=$(dirname "$0")
+
+# Source the utilities
+source $dir/utils.sh
+
+# Variables
+DRY_RUN=${DRY_RUN:="false"}
+
+# Update version of noobaa core image, noobaa operator image and CLI version in the README output.
```
liranmauda, line 20:
```suggestion
# Update the version of noobaa core image, noobaa operator image, and CLI version in the README output.
```

build/tools/builder-pre.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+# NOTE: This script can be run manually but is run anyway by builder.sh before that script starts executing its `main`
+
+# This script will automate the file alteration, which involves
+# 1. Updating the README.md
+# 2. Commiting the changes and pushing the changes
+
```
liranmauda, line 8: We might want to add PS4 for this script to see the debugging:
```suggestion
export PS4='\e[36m+ ${FUNCNAME:-main}@${BASH_SOURCE}:${LINENO} \e[0m'
```

build/tools/builder.sh
```diff
@@ -0,0 +1,235 @@
+#!/bin/bash
+
+# This script is supposed to automate various types of artefact generation that this project has
+# to do.
+# Following types of artefacts are supported:
```
liranmauda, line 5:
```suggestion
# This script is supposed to automate various types of artifact generation that this project has
# to do.
# Following types of artifacts are supported:
```

build/tools/builder.sh
```diff
@@ -0,0 +1,235 @@
...
+# 1. Build a new version of the CLI
+# 2. Build a new version of OCI image
+# 3. Generate Krew manifest
+# 4. Generate Homebrew formula
+
```
liranmauda, line 10: We might want to add PS4 for this script to see the debugging:
```suggestion

export PS4='\e[36m+ ${FUNCNAME:-main}@${BASH_SOURCE}:${LINENO} \e[0m'
```

liranmauda, line 10: We might want to add PS4 for this script to see the debugging:
```suggestion

export PS4='\e[36m+ ${FUNCNAME:-main}@${BASH_SOURCE}:${LINENO} \e[0m'
```

build/tools/builder.sh
```diff
@@ -0,0 +1,235 @@
...
+		shift
+	done
+}
+
+function init() {
+	mkdir -p $bin_path
+	mkdir -p $krew_path
+  mkdir -p $homebrew_path
+  mkdir -p $operator_bundle_path
+	parse_args "$@"
```
liranmauda, line 65: Indentation

build/tools/builder.sh
```diff
@@ -0,0 +1,235 @@
...
+				shift
+				;;
+			--gh-repo)
+				GH_REPO="$2"
+				shift
+				;;
+      --oci-org)
+        OCI_ORG="$2"
+        shift
+        ;;
```
liranmauda, line 43: indentation

build/tools/releaser-post.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+# This script will automate the file alteration, which involves
+# 1. Updating version of the CLI in version/version.go
+# 2. Updating version of noobaa core image in pkg/options.go => ContainerImageTag
+# 3. Run make gen
+# 4. Commit the changes with automated commit message
+
```
liranmauda, line 8: We might want to add PS4 for this script to see the debugging:
```suggestion

export PS4='\e[36m+ ${FUNCNAME:-main}@${BASH_SOURCE}:${LINENO} \e[0m'
```

build/tools/releaser-post.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
+
+# This script will automate the file alteration, which involves
+# 1. Updating version of the CLI in version/version.go
+# 2. Updating version of noobaa core image in pkg/options.go => ContainerImageTag
+# 3. Run make gen
+# 4. Commit the changes with automated commit message
```
liranmauda, line 7:
```suggestion
# This script will automate the file alteration, which involves
# 1. Updating the version of the CLI in version/version.go
# 2. Updating version of noobaa core image in pkg/options.go => ContainerImageTag
# 3. Run make gen
# 4. Commit the changes with an automated commit message
```

build/tools/releaser-post.sh
```diff
@@ -0,0 +1,61 @@
...
+source $dir/utils.sh
+
+version=$(bump_semver_patch $(get_noobaa_version))
+DRY_RUN=${DRY_RUN:="false"}
+
+# Update version of the CLI in version/version.go
+function update_version() {
+	local version_file=version/version.go
+	# Replace version line with new version using perl because sed is not compatible with both mac and linux
+	# and awk solution is clumsy
```
liranmauda, line 24:
```suggestion
	# Replace the version line with the new version using perl because sed is not compatible with both mac and linux
	# and awk solution is clumsy
```

build/tools/utils.sh
```diff
@@ -0,0 +1,66 @@
...
+  if [[ $# -eq 1 ]]; then
+	if [[ "$1" == "1" ]]; then
+		version_with_v "$version"
+	else
+		version_without_v "$version"
+	fi
+  else
+    version_without_v "$version"
+  fi
+}
```
liranmauda, line 50: not sure that I like 1 and 0:
0 has no meaning here as we don't even check that so the comment is misleading.
and using $1 directly should be avoided
I suggest using a variable and changing it if we get the argument:
maybe something like: 
```suggestion
# get_noobaa_version returns the NooBaa version by reading the `cmd/version/main.go`
#
# Note: If an argument is presented, it will append "v" to the version
function get_noobaa_version() {
  local version=$(go run cmd/version/main.go)
  if [[ $# -eq 1 ]]
  then
	version_with_v "$version"
  else
    version_without_v "$version"
  fi
}
```

liranmauda (changes requested):
2 general comments:
1. We should probably add PS4 in all the scripts and add `set -x` also.
2. don't forget the newline at the end of each file

build/tools/builder-pre.sh
utkarsh-pro: Yes, will add this.

build/Dockerfile
utkarsh-pro: Actually I added it to make a way for us to build to multi-platform image by mixing `docker manifest` and `docker build`
ARG was chosen over ENV because args can be used to build a conditional multi-stage dockerfile (`if ARG is "aarch64 do ...` kind of) .

.github/workflows/releaser.yaml
utkarsh-pro: We will need it. I have commented it out to because we will need to create first PR in the krew repository manually and only after that this bot can be used (limitation of the krew bot). I left it here as a comment to highlight that this needs to be added in the workflow so that others are aware of it.

build/tools/builder.sh
utkarsh-pro: Ah, US English 😛 . Will change it.

build/tools/utils.sh
utkarsh-pro: I think the comments are stale, previously this function explicitly just supported `0` (false) and `1` (true).

Will commit the above suggestion.

utkarsh-pro: I didn't notice this yesterday but I am not sure if this is a good function signature. It allows the argument to be completely arbitrary which seems a bit odd (it might be that I am just being pedantic).

liranmauda: Yes, that's true, that's the reason for `maybe something like:` 😄 
The thing is that in bash you need to check it, but in this case, ANY flag should do it.
Well, the example above is if you get a flag act this way otherwise act as the default.
If you want to be strict you can get the $1 into a variable and check its value, and prompt an error if that is not what you expected.
might be overkill as this function is internal, but you can do it.

build/tools/utils.sh
```diff
@@ -0,0 +1,68 @@
...
+    if [[ "$1" == "1" ]]; then
+      version_with_v "$version"
+    else
+      version_without_v "$version"
+      echo "[Warn]: Unknown argument passed to get_noobaa_version."
+    fi
+  else
+    version_without_v "$version"
+  fi
+}
```
liranmauda, line 52: It is an internal function so we can be not strict about it, but in general, I don't like the `if [[ $# -eq 1 ]]`
basically, it assumes that there will be a max of one argument.
but in the case of `get_noobaa_version 1 2` the intention is `version_with_v` as we got `1` in the first argument, but it will never get into the if.
If we want to be strict about it then we can add something like:
```bash
if [[ $# -gt 1 ]] 
then
    echo "[Error]: expecting one argument at the most" 
fi
```

build/tools/utils.sh
```diff
@@ -0,0 +1,68 @@
+#!/bin/bash
+
+export PS4='\e[36m+ ${FUNCNAME:-main}@${BASH_SOURCE}:${LINENO} \e[0m'
```
liranmauda, line 3: I see that we added this here, but in general, We should add it to each bash file, as we do with `set -x`

utkarsh-pro: Done

build/tools/utils.sh
utkarsh-pro: Done

liranmauda (approved):
LGTM

liranmauda:
Versions of the branches:
5.8.18, 5.9.14, 5.10.10, 5.11.5

dannyzaken (approved):

build/tools/utils.sh
```diff
@@ -0,0 +1,71 @@
...
+    if [[ "$1" == "1" ]]; then
+      version_with_v "$version"
+    else
+      version_without_v "$version"
+      echo "[Warn]: Unknown argument passed to get_noobaa_version."
+    fi
+  else
+    version_without_v "$version"
+  fi
+}
```
liranmauda, line 55: Same as the core: 

We cant use `echo "[Error]:` or `echo "[Warn]:` here unless we are handling this error on the call.
Currently, upon error or this warning, we will change the version to include this string.

liranmauda (changes requested):

liranmauda (approved):

PR #999, opened 2022-12-01
RBAC for endpoint is handled by OLM

baum:
Avoid race where noobaa-endpoint SA does not exist while the noobaa-db stateful set is transitioning via init container image update. Pass the endpoint RBAC to OLM via CSV, while handling the endpoint's SCC within the operator.

See https://bugzilla.redhat.com/show_bug.cgi?id=2149261

### DIFF
--- a/pkg/operator/operator.go
+++ b/pkg/operator/operator.go
@@ -105,8 +105,11 @@ func RunInstall(cmd *cobra.Command, args []string) {
 	c := LoadOperatorConf(cmd)
 	util.KubeCreateSkipExisting(c.NS)
 	util.KubeCreateSkipExisting(c.SA)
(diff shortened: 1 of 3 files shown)


pkg/operator/operator.go
```diff
@@ -145,7 +148,7 @@ func waitForOperatorPodExit() {
 		// List failure
 		if !listRes {
 			log.Printf("❌ Can not list pods in %v namespace with noobaa-operator=deployment label, try again.", options.Namespace)
-		// Exit condition, list succeded and no operator's pods are found
+			// Exit condition, list succeded and no operator's pods are found
```
dannyzaken, line 151: mistake?

dannyzaken (approved):

PR #1002, opened 2022-12-05
Support feature to disable default backingstore

achouhan09:
…tBackingStoreSpec

Signed-off-by: Aayush Chouhan <achouhan@redhat.com>

### Explain the changes
1. deprecated `DefaultBackingStoreSpec`
2. Added flag `DisableDefaultBackingStore` to allow deletion of default backingstore.

- [x] Tests added

### DIFF
--- a/deploy/crds/noobaa.io_noobaas_crd.yaml
+++ b/deploy/crds/noobaa.io_noobaas_crd.yaml
@@ -1008,7 +1008,8 @@ spec:
                 - default_level
                 type: integer
               defaultBackingStoreSpec:
-                description: BackingStoreSpec defines the desired state of BackingStore
+                description: 'Deprecated: DefaultBackingStoreSpec is not supported
+                  anymore, use DisableDefaultBackingStore instead.'
                 properties:
                   awsS3:
                     description: AWSS3Spec specifies a backing store of type aws-s3
@@ -1216,6 +1217,11 @@ spec:
                 required:
                 - type
                 type: object
+              disableDefaultBackingStore:
+                description: DisableDefaultBackingStore (optional - default value
+                  is false) if true it will allow the user to delete DefaultBackingStore
+                nullable: true
+                type: boolean
               disableLoadBalancerService:
                 description: DisableLoadBalancerService (optional) sets the service
                   type to ClusterIP instead of LoadBalancer
--- a/deploy/crds/noobaa.io_noobaas_crd.yaml
+++ b/deploy/crds/noobaa.io_noobaas_crd.yaml
@@ -1009,7 +1009,7 @@ spec:
                 type: integer
               defaultBackingStoreSpec:
                 description: 'Deprecated: DefaultBackingStoreSpec is not supported
-                  anymore, use DisableDefaultBackingStore instead.'
+                  anymore, use ManualDefaultBackingStore instead.'
                 properties:
                   awsS3:
                     description: AWSS3Spec specifies a backing store of type aws-s3
@@ -1217,11 +1217,6 @@ spec:
                 required:
                 - type
                 type: object
-              disableDefaultBackingStore:
-                description: DisableDefaultBackingStore (optional - default value
-                  is false) if true it will allow the user to delete DefaultBackingStore
-                nullable: true
-                type: boolean
               disableLoadBalancerService:
                 description: DisableLoadBalancerService (optional) sets the service
                   type to ClusterIP instead of LoadBalancer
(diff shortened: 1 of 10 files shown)


liranmauda (changes requested):

pkg/system/phase1_verifying.go
```diff
@@ -132,14 +131,9 @@ func (r *Reconciler) CheckSystemCR() error {
...
+	// nolint: staticcheck
 	if r.NooBaa.Spec.DefaultBackingStoreSpec != nil {
-		bs := nbv1.BackingStore{
-			Spec: *r.NooBaa.Spec.DefaultBackingStoreSpec,
-		}
-		err = validations.ValidateBackingStore(bs)
-		if err != nil {
-			return util.NewPersistentError("InvalidDefaultBackingStoreSpec", fmt.Sprintf(`%s`, err))
-		}
+		return util.NewPersistentError("Invalid, DefaultBackingStoreSpec was deprecated", fmt.Sprintf(`%s`, err))
```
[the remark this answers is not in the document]
achouhan09: @liranmauda yes thats the thought, Do we need any change here?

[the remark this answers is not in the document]
liranmauda: @achouhan09 My actual comment was longer than that but it somehow got deleted when I committed this, not sure why...

So, what Is missing in this comment is that @dannyzaken and I talked and I am aware that this is a flag that no one should use on a real system, But in that case, I wanted to ask, or raise it here also, why should we deprecate the flag and not completely remove it? 
@achouhan09  @dannyzaken WDYT?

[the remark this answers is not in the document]
dannyzaken: we will do it in stages. we can deprecate and remove later

liranmauda (commented):
@achouhan09 and I looked at the installation and saw that there is a `noobaa-default-bucket-class` even when we are using the `disableDefaultBackingStore` flag:
```
#------------------#
#- Bucket Classes -#
#------------------#

NAME                          PLACEMENT                                                        NAMESPACE-POLICY   QUOTA   PHASE      AGE
noobaa-default-bucket-class   {"tiers":[{"backingStores":["noobaa-default-backing-store"]}]}   null               null    Rejected   18s
```
@dannyzaken Why do we need `default-bucket-class` if we dont have `default-backing-store`? 
I think that this flag can also skip creating `default-bucket-class`, WDYT?

liranmauda:
@achouhan09 Found that when we try to delete the default Backingstore we face the following error: 
`Error from server: admission webhook "admissionwebhook.noobaa.io" denied the request: cannot complete because pool "noobaa-default-backing-store" in "DEFAULT_RESOURCE" state`

We should probably take care of it in this function:

https://github.com/noobaa/noobaa-operator/blob/fa46462481c800dfafe374e343741982b107f3ec/pkg/validations/backingstore_validations.go#L260-L278

as the intention is to `allow the user to delete DefaultBackingStore` as described in the NooBaa CRD

@achouhan09 @dannyzaken WDYT?

achouhan09:
@dannyzaken @liranmauda PR is ready for the review, Can you pls have a look

test/cli/test_cli_flow.sh
```diff
@@ -26,6 +26,7 @@ function post_install_tests {
     check_pgdb_config_override
     test_noobaa_cr_deletion
     test_noobaa_loadbalancer_source_subnet
+    check_default_backingstore
```
dannyzaken, line 29: Can you add a comment that specifies the changes this test does - delete bs, delete firs.bucket, etc.

achouhan09: Done. Thanks!

dannyzaken (approved):

deploy/crds/noobaa.io_noobaas_crd.yaml
```diff
@@ -1327,6 +1328,11 @@ spec:
                       type: string
                     type: array
                 type: object
+              manualDefaultBackingStore:
+                description: ManualDefaultBackingStore (optional - default value is
+                  false) if true it will allow the user to delete DefaultBackingStore
```
liranmauda, line 1333: Maybe change the description to describe the manual treatment of the DefaultBackingStore

liranmauda (changes requested):

deploy/crds/noobaa.io_noobaas_crd.yaml
achouhan09: Sure @liranmauda 
I will change it to :
```suggestion
                description: ManualDefaultBackingStore (optional - default value is
                  false) if true it will allow the user to delete DefaultBackingStore, user needs to delete associated buckets 
                  and update the admin account with new BackingStore in order to delete the DefaultBackingStore
```

liranmauda (approved):

PR #1006, opened 2022-12-28
Bumped to GO v1.18 rook v1.10.1 controller-runtime v0.13.1

liranmauda:
### Explain the changes

github.com/sirupsen/logrus v1.8.1 -> v1.9.0
github.com/rook/rook v1.8.7 -> v1.10.1
sigs.k8s.io/controller-runtime v0.10.2 -> v0.13.1
And more dependencies updates.

Added `get install` to the tests yamls:
 - go get -d flag tells go get not to build or install packages.
 - In Go 1.18, -d will always be enabled. (https://go.dev/ref/mod#go-get)
 - We need to install ginkgo


utkarsh-pro:
Do we need to do both `go get` and `go install` for executables?

liranmauda:
> Do we need to do both `go get` and `go install` for executables?

Seems like.
`go get` will not install as In Go 1.18, -d will always be enabled. (https://go.dev/ref/mod#go-get)
and `go install` seems not to get the vendor, well at list at the Github Action it didn't work and we have to add the `go get` before.

utkarsh-pro (approved):

PR #1023, opened 2023-01-18
Integrate with KEDA for custom metrics scaling

naveenpaul1:
### Explain the changes
1. Integrate with KEDA for custom metrics scaling: Initial changes
     This PR will have only Keda related changes, not prometheus-adaptor 

Signed-off-by: Naveen Paul @naveenpaul1 

1. https://issues.redhat.com/browse/MCGI-101

### Testing Instructions:
1. Install noobaa in a cluster after installing keda and Prometheus
2. check for the newly created Keda HPA 

- [ ] Doc added/updated
- [ ] Tests added


pkg/system/azure_utils.go
```diff
@@ -5,7 +5,7 @@ import (
 	"log"
 	"net/url"
 
-	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-06-01/storage"
+	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2022-05-01/storage"
```
shirady, line 8: @naveenpaul1 this line failed the action (you can delete my comment after you'll see it).

> SA1019: "github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2022-05-01/storage" is deprecated: Please note, this package has been deprecated. A replacement package is available [github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage](https://pkg.go.dev/github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/storage/armstorage). We strongly encourage you to upgrade to continue receiving updates. See [Migration Guide](https://aka.ms/azsdk/golang/t2/migration) for guidance on upgrading. Refer to our [deprecation policy](https://azure.github.io/azure-sdk/policies_support.html) for more details. (staticcheck)

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -457,3 +462,16 @@ const (
...
 	DBTypePostgres DBTypes = "postgres"
 )
+
+// AutoscalingTypes is a string enum type for specify the types of autoscalling supported.
+type AutoscalingTypes string
+
+// These are the valid DB types:
+const (
+	// AutoscalingTypeDefault is hpa
+	AutoscalingTypeDefault AutoscalingTypes = "default"
```
dannyzaken, line 472: I think it should be `AutoscalingTypeHPAV1` instead of `AutoscalingTypeDefault `. if we change the default from HPAV1 to something else then `AutoscalingTypeDefault` will be redundant. Also, we might want to have an option to set the autoscaler explicitly to HPAV1, mainly for customer support workarounds

naveenpaul1: code updated

pkg/system/phase4_configuring.go
```diff
@@ -541,16 +545,109 @@ func (r *Reconciler) awaitEndpointDeploymentPods() error {
...
 	}
+	var err error
+	prometheusList := deployedPrometheus(log)
+	switch r.NooBaa.Spec.AutoscalingType {
+	case "keda":
+		err = r.autoscaleKeda(prometheusList)
+	case "hpav2":
+		//TODO
+		log.Infof("HPAV2 implementation pending")
+	default:
```
dannyzaken, line 562: instead of `default` use a specific handling of `hpav1`. if `r.NooBaa.Spec.AutoscalingType` is not specified, use `options.DefaultAutoscalingType`

pkg/system/phase4_configuring.go
```diff
@@ -541,16 +545,109 @@ func (r *Reconciler) awaitEndpointDeploymentPods() error {
...
+		if err := r.ReconcileObject(r.HPAEndpoint, r.SetDesiredHPAEndpoint); err != nil {
+			return err
+		}
+	}
+	if err != nil {
```
dannyzaken, line 567: I think there shouldn't be a fallback in case of an error in deploying the autoscaler. this error should be returned back and treated as any other temp error. the operator will retry in the next reconcile

naveenpaul1: @dannyzaken if we return the error back to the main reconcile method `ReconcilePhaseConfiguring`, then it will skip the rest of the flow in that reconcile method, is that Ok?

pkg/options/options.go
```diff
@@ -130,6 +130,10 @@ var ShowSecrets = false
 // ManualDefaultBackingStore is used for disabling and allow deletion of default backingstore
 var ManualDefaultBackingStore = false
 
+// AutoscalingType is the default noobaa-endpoint autoscaling type
+// it can be overridden for testing or different types. for now default is hpav1 it will replace with hpav2 in future
+var AutoscalingType = "hpav1"
```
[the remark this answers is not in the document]
naveenpaul1: code updated

pkg/system/phase4_configuring.go
```diff
@@ -537,16 +541,156 @@ func (r *Reconciler) awaitEndpointDeploymentPods() error {
...
 	if r.HPAEndpoint.UID == "" {
 		if err := r.awaitEndpointDeploymentPods(); err != nil {
 			return err
 		}
 	}
+	var err error
+	prometheusList := deployedPrometheus(log)
+	switch r.NooBaa.Spec.AutoscalingType {
+	case "keda":
+		if err = r.deleteExistingAutoscaler(string(r.NooBaa.Spec.AutoscalingType)); err != nil {
```
dannyzaken, line 554: `deleteExistingAutoscaler` is relevant for all cases, so maybe call it once before the switch.
Also, the name is a bit confusing. Maybe rename it to something like `cleanPreviousAutoscalingResources`? Something that better state that this is a cleanup code

pkg/system/phase4_configuring.go
```diff
@@ -537,16 +541,156 @@ func (r *Reconciler) awaitEndpointDeploymentPods() error {
...
+	default:
+		log.Infof("❌ Autoscaling type is not defined for noobaa")
+		return errors.New("Autoscaling type is not defined")
+	}
+	return r.updateNoobaaEndpoint()
+
+}
+
+func (r *Reconciler) deleteExistingAutoscaler(autoscalingType string) error {
+	autoscalers := &autoscalingv1.HorizontalPodAutoscalerList{}
```
dannyzaken, line 579: notice that here you only list HPAV1 resources. these are not relevant for KEDA for example. 
I suggest having a dedicated cleanup function for each type of autoscaler and from here call the relevant cleanups according to `autoscalingType`. e.g.:

```go
if autoscalingType == `nbv1.AutoscalingTypeKeda` {
   cleanupHPAV1()
   cleanupHPAV2()
}
if autoscalingType == `nbv1.AutoscalingTypeHPAV1` {
   cleanupKEDA()
   cleanupHPAV2()
}
// ... etc.
```

pkg/system/phase4_configuring.go
```diff
@@ -537,16 +541,156 @@ func (r *Reconciler) awaitEndpointDeploymentPods() error {
...
+	log := r.Logger.WithField("func", "ReconcileHPAEndpoint")
 	if r.HPAEndpoint.UID == "" {
 		if err := r.awaitEndpointDeploymentPods(); err != nil {
 			return err
 		}
 	}
+	var err error
+	prometheusList := deployedPrometheus(log)
+	switch r.NooBaa.Spec.AutoscalingType {
+	case "keda":
```
dannyzaken, line 553: a general comment - instead of referring to the types as string, use the consts 
```suggestion
	case nbv1.AutoscalingTypeKeda:
```

liranmauda:
Small changes, as for the review of the flow we will do it online as discussed

deploy/cluster_role.yaml
```diff
@@ -80,3 +81,17 @@ rules:
       - securitycontextconstraints
     verbs:
       - '*'
+  - apiGroups:
+      - keda.sh
+    resources:
+      - '*'
```
dannyzaken, line 87: Let's avoid using wildcards where we can. we already have a BZ regarding it (https://bugzilla.redhat.com/show_bug.cgi?id=2052928)

pkg/system/autoscaler.go
```diff
@@ -0,0 +1,239 @@
...
+
+	var promethesNamespace string
+	var serviceAccountName string
+	if len(prometheusList.Items) > 1 {
+		for _, prometheus := range prometheusList.Items {
+			promethesNamespace = prometheus.Namespace
+			serviceAccountName = prometheus.Spec.ServiceAccountName
+			if prometheus.Namespace == "openshift-monitoring" {
+				break
+			}
```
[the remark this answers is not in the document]
dannyzaken: Following a discussion with @naveenpaul1 we decided to have an `AutoscalingSpec` property in the noobaa CRD, that will hold both the type and other options, such as the Prometheus NS.
In that case we should not look for all installed Prometheus instances. instead, we should fail the reconcile if the given or default (`openshift-monitoring`) does not exist

pkg/system/autoscaler.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	return nil
+}
+
+func (r *Reconciler) cleanPreviousAutoscalingResources(autoscalingType string) error {
+	autoscalers := &autoscalingv1.HorizontalPodAutoscalerList{}
+	util.KubeList(autoscalers, &client.ListOptions{Namespace: r.Request.Namespace})
+	if len(autoscalers.Items) == 0 {
+		return nil
+	}
```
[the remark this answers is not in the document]
naveenpaul1: @dannyzaken we need to delete the existing HPA even for Keda and HPAV2 for that we need to get the existing autoscaler, correct me if am wrong

pkg/system/autoscaler.go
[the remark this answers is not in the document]
naveenpaul1: @dannyzaken If Prometheus is only in one namespace should we consider the given Prometheus namespace?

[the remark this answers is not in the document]
dannyzaken: I think we should always consider the given Prometheus namespace. if prometheus is not installed in the given NS, then it's an error. 
I wonder now if we should even have a default (`openshift-monitoring`) or maybe it's better to have this field mandatory when the type is KEDA or HPAV2 (we can enforce that in the admission controller). @liranmauda any thoughts?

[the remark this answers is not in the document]
liranmauda: Maybe it should be mandatory...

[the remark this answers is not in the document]
naveenpaul1: ya, It makes sense, it might add small complexity on the user side but if this value is not configured properly it will affect the functionality.

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -197,6 +197,20 @@ type NooBaaSpec struct {
...
+}
+
+// AutoscalerSpec defines different actoscaling spec such as autoscaler type and prometheus namespace
+type AutoscalerSpec struct {
+	// Type of autoscaling (optional) for noobaa-endpoint, hpav1 - default kubernetes based, hpav2 and keda - Prometheus metrics based
+	// +kubebuilder:validation:Enum=hpav1;hpav2;keda
+	AutoscalerType AutoscalerTypes `json:"autoscalerType"`
+
+	// Prometheus namespace that scrap metrics from noobaa
+	PrometheusNamespace string `json:"prometheusNamespace"`
```
dannyzaken, line 213: `PrometheusNamespace` should be optional
```suggestion
	// Prometheus namespace that scrap metrics from noobaa
	// +optional
	PrometheusNamespace string `json:"prometheusNamespace"`
```

pkg/system/autoscaler.go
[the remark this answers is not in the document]
naveenpaul1: @dannyzaken I could see that util.KubeList() is listing all HPAs even with version `autoscaling/v2`. This code can delete all the HPA with  `apiVersion` such as `autoscaling/v2`, `autoscaling/v2beta2` and `autoscaling/v1`.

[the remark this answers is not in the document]
dannyzaken: @naveenpaul1 That's weird since in `KubeList` we list for specific `GroupVersionKind` and also `autoscalers` is of type `autoscalingv1.HorizontalPodAutoscalerList`. can you validate? if this is indeed the behavior then it seems like a bug to me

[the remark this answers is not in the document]
naveenpaul1: @dannyzaken  autoscalingv1 and autoscalingv2 check is added to the code before deleting HPAs, PR updated

pkg/system/autoscaler.go
```diff
@@ -0,0 +1,269 @@
...
+			return err
+		}
+	default:
+		log.Errorf("❌ Autoscaler type is not defined for noobaa")
+		return errors.New("Autoscaler type is not defined")
+	}
+	return nil
+}
+
+func (r *Reconciler) cleanPreviousAutoscalingResources(log *logrus.Entry, autoscalerType nbv1.AutoscalerTypes) error {
```
dannyzaken, line 74: The cleanup code can still be much cleaner in my opinion. maybe I'm missing something, but I don't see the benefit of listing all types of scalers on every reconciliation and aggregating it into a single list (e.g.: if `autoscalerType` is hpav2, why do we need to list it for cleanup?). we should only list and clean for each type if it's not the current `autoscalerType`.
Also, I think we should not continue reconciling the autoscaler before other types are cleaned, so we won't have 2 different autoscalers configuring the endpoints deployment. so if any other autoscalers exist, then the cleanup function should return a temp error.  maybe also rename the function to something that better states that we are blocked for the cleanup, something like `ensurePreviousAutoscalerCleanup`

I suggest something like that:

```golang
func (r *Reconciler) ensurePreviousAutoscalerCleanup(autoscalerType nbv1.AutoscalerTypes) error {
   if autoscalerType != nbv1.AutoscalerTypeHPAV1 {
      if err := r.ensureHPAV1Cleanup(); err != nil {
         return err
      }
   }
   if autoscalerType != nbv1.AutoscalerTypeHPAV2 {
      if err := r.ensureHPAV2Cleanup(); err != nil {
         return err
      }
   }
   if autoscalerType != nbv1.AutoscalerTypeHPAV1 {
      if err := r.ensureKEDACleanup(); err != nil {
         return err
      }
   }

   return nil
}


func (r *Reconciler) ensureHPAV1Cleanup() error {
   noobaaHpaSelector, _ := labels.Parse("app=noobaa")
   //get autoscaler only for HPAV1 based on the label
   autoscalersv1 := &autoscalingv1.HorizontalPodAutoscalerList{}
   util.KubeList(autoscalersv1, &client.ListOptions{Namespace: r.Request.Namespace, LabelSelector: noobaaHpaSelector})
   // ... delete all autoscaresv1. I don't see a reason to look for a specific name, since you already list based on the label.

   
   if len(autoscalersv1) > 0 {
      //As long as we still have HPAV1 running we don't want to continue. return error
      return fmt.Errorf("HPAV1 autoscaler still not deleted ")
   }

}


// same for HPAV2\KEDA
```


notice that I also removed the logger parameter. we already have it on the Reconciler struct

pkg/system/phase4_configuring.go
```diff
@@ -44,6 +43,7 @@ const (
 	ibmEndpoint = "https://s3.direct.%s.cloud-object-storage.appdomain.cloud"
 	ibmLocation = "%s-standard"
 	ibmCOSCred  = "ibm-cloud-cos-creds"
+	schema      = "https://"
```
dannyzaken, line 46: move to `autoscalers.go`

pkg/system/phase4_configuring.go
```diff
@@ -537,16 +537,20 @@ func (r *Reconciler) awaitEndpointDeploymentPods() error {
...
-	if err := r.ReconcileObject(r.HPAEndpoint, r.SetDesiredHPAEndpoint); err != nil {
+	if err := r.reconcileAutoscaler(); err != nil {
 		return err
 	}
+	return r.updateNoobaaEndpoint()
+
+}
 
+func (r *Reconciler) updateNoobaaEndpoint() error {
 	max := r.HPAEndpoint.Spec.MaxReplicas
```
dannyzaken, line 554: This code seems irrelevant now, since `HPAEndpoint` is only relevant for HPAV1. maybe we should get min\max from `endpointsSpec`?

deploy/internal/hpa-keda-scaled-object.yaml
```diff
@@ -0,0 +1,25 @@
...
+  pollingInterval: 30
+  scaleTargetRef:
+    name: noobaa-endpoint
+  triggers:
+    - authenticationRef: 
+        name: keda-prom-creds
+      metadata:
+        authModes: bearer
+        metricName: NooBaa_num_namespace_buckets
+        query: '(sum(irate(container_network_receive_bytes_total{pod=~"noobaa-endpoint-.*", namespace="placeholder"}[5m])) by (pod, namespace))'
```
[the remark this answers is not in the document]
naveenpaul1: I think disabling Keda till we finalize the metrics also should be an option, don't see any value in spending time on something that we will remove in the future. WDYT?

dannyzaken (approved):

deploy/internal/hpa-keda-scaled-object.yaml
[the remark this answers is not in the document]
naveenpaul1: Final decision: Leave the autoscale code, and do not disable Keda changes.

PR #1039, opened 2023-01-31
Add the characters EXAMPLE as a prefix in secrets examples

shirady:
Signed-off-by: shirady <57721533+shirady@users.noreply.github.com>

### Explain the changes
1. Add the characters EXAMPLE as a prefix in secrets examples for lines the infoSec found in its alerts that were false-positive alerts in files with ".md" extension.

### Issues: Gap
1. When I worked on the GH Action on a copy of noobaa-operator repository I got false-positive alerts from infoSec.
2. One of the ways to turn it off is to add EXAMPLE as a prefix in secrets, which is the simplest way, suggested by @bplaxco.

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/doc/install-on-minikube-on-ubuntu.md
+++ b/doc/install-on-minikube-on-ubuntu.md
@@ -214,7 +214,7 @@ PodPorts    : [https://172.17.0.5:6443]
 #------------------#
 
 AWS_ACCESS_KEY_ID     : ku2pBEPGqsTvGd47qfL9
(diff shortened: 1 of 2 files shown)


bplaxco:
@shirady I ran the scanner against the content with "EXAMPLE" included and it ignores it as expected :+1:

dannyzaken (approved):

PR #1062, opened 2023-02-23
Fix Bug BZ 2149872, 1066 | Add Toleration And Affinity to Backingstore Pod after Editing Noobaa CR

shirady:
### Explain the changes
1.  Add toleration and affinity to backingstore pod after editing noobaa CR by adding it as a part of `needUpdate`.
2. Edit the controller of backingstore and add a watch explicitly on changes in noobaa CR.
3. Create a generic function `contains` (will be discussed).
4. Change the condition of toleration and affinity (to be set even if it is `nil`).

### Issues: Fixed [BZ 2149872](https://bugzilla.redhat.com/show_bug.cgi?id=2149872), #1066
1. Title:  Toleration not added to the noobaa-default-backing-store pod after editing storagecluster CR and subs with toleration (to the original BZ about **tolerations** we also added issue #1066 about **affinities**).
2. Before we implemented the change (on adding a change - affinity or toleration): the pods core, db, and endpoint were be terminated and then running, but the operator and backinstore pods were not. 
For the operator it is clear, this is what starts the system, changes in Noobaa CR does not pass to it. 
In the backing store pod, it is a bug.
3. We also solved when the tolerations or affinities were deleted the pods core, db, endpoint and backingstore were not terminated.

#### Toleration (add a new one):
Requisitions: Have noobaa system running.
Test on minikube - only tolerations (without taint, in minikube we have only one node).
1. Change Noobaa CR and add Toleration to it:  `kubectl edit noobaa noobaa`.
2. Paste this inside :
```
tolerations:
  - effect: NoSchedule
     key: xyz
     operator: Equal
     value: "true"
```
It looks like this (after paste):
Screenshot 2023-02-15 at 12 33 02
3. See the changes of the pods:  `kubectl get pods -w`.
would expect to see the pods core, db, endpoint and backingstore terminated. 

#### Affinity (add a new one):
Test on minikube
1. Add affinity on the node: `kubectl label nodes minikube name=app-shira` (I chose `name=app-shira` as a new label).
2. If you want to check that your label was added: `kubectl describe nodes minikube`.
3. Change Noobaa CR and add Affinity to it:  `kubectl edit noobaa noobaa`.
4. Paste this inside :
```
  affinity:
    nodeAffinity:
      requiredDuringSchedulingIgnoredDuringExecution:
        nodeSelectorTerms:
        - matchExpressions:
          - key: name
            operator: In
            values:
            - app-shira
```
It looks like this (after paste):
Screenshot 2023-02-28 at 12 49 07
5. See the changes of the pods:  `kubectl get pods -w`.
would expect to see the pods core, db, endpoint and backingstore terminated. 
- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -1064,15 +1064,6 @@ func (r *Reconciler) reconcilePvPool() error {
 	return r.reconcileExistingPods(podsList)
 }
 
-func contains(slice []string, str string) bool {
-	for _, s := range slice {
-		if s == str {
-			return true
-		}
-	}
-	return false
-}
-
 func (r *Reconciler) reconcileMissingPods(podsList *corev1.PodList, pvcsList *corev1.PersistentVolumeClaimList) error {
 	claimNames := []string{}
 	for _, pod := range podsList.Items {
@@ -1082,7 +1073,7 @@ func (r *Reconciler) reconcileMissingPods(podsList *corev1.PodList, pvcsList *co
 		return err
 	}
 	for _, pvc := range pvcsList.Items {
-		if !contains(claimNames, pvc.Name) {
+		if !util.Contains(claimNames, pvc.Name) {
 			i := strings.LastIndex(pvc.Name, "-")
 			postfix := pvc.Name[i+1:]
 			newPod := r.PodAgentTemplate.DeepCopy()
@@ -1187,6 +1178,45 @@ func (r *Reconciler) needUpdate(pod *corev1.Pod) bool {
 		r.Logger.Warnf("Change in Image Pull Secrets detected: NoobaaSecret(%v) Spec(%v)", noobaaSecret, podSecrets)
 		return true
 	}
+
+	// In case noobaa CR was changed and a new toleration was added
+	// we want that change to pass to the backingstore pod as well.
+	// This change does not support removal of a toleration, in this case
+	// the user will need to manually terminate the backingstore pod.
+	podTolerations := pod.Spec.Tolerations
+	noobaaTolerations := r.NooBaa.Spec.Tolerations
+	for _, noobaaToleration := range noobaaTolerations {
+		if !util.Contains(podTolerations, noobaaToleration) {
+			return true
+		}
+	}
+
+	// In case noobaa CR was changed and a new affinity was added
+	// we want that change to pass to the backingstore pod as well.
+	podAffinitiesPtr := pod.Spec.Affinity
+	noobaaAffinitiesPtr := r.NooBaa.Spec.Affinity
+	// There 4 cases
+	//                                          noobaaAffinitiesPtr value
+	//                  |         | nil                         | not nil                        |
+	//                  | ------- | --------------------------- | ------------------------------ |
+	// podAffinitiesPtr | nil     | false (nothing changed)     | true (affinity was added)      |
+	// value            | not nil | true (affinity was deleted) | depends on the reference value |
+	//                  | ------- | --------------------------- | ------------------------------ |
+	if noobaaAffinitiesPtr != nil {
+		noobaaAffinities := *noobaaAffinitiesPtr
+		if podAffinitiesPtr == nil {
+			return true
+		}
+		podAffinities := *noobaaAffinitiesPtr
+		if podAffinities != noobaaAffinities {
+			return true
+		}
+	} else {
+		if podAffinitiesPtr != nil {
+			return true
+		}
+	}
+
 	return false
 }
 
--- a/pkg/backingstore/backingstore.go
+++ b/pkg/backingstore/backingstore.go
@@ -1073,3 +1073,31 @@ func MapSecretToBackingStores(secret types.NamespacedName) []reconcile.Request {
 
 	return reqs
 }
+
+// MapNoobaaToBackingStores returns a list of backingstores that are inside Noobaa system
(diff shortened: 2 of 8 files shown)


shirady:
### Specific discussion on `contains` function:
1. In the reconciler of the backing store there was a contains function (for internal package use only).
https://github.com/noobaa/noobaa-operator/blob/d821724e89d182e69ef05c7c533398c9c4a5241c/pkg/backingstore/reconciler.go#L1067-L1074
2. I also had to use a contains function for another type (not strings). In Go there is no function overloading, so I could not create another function with the same name that differs only by type, so I changed it to a generic function.
2. Anyway the original contains is a kind of code duplication of `contains` inside `utils` file.
https://github.com/noobaa/noobaa-operator/blob/d821724e89d182e69ef05c7c533398c9c4a5241c/pkg/util/util.go#L1325-L1334
3. But it was changed in one line that is specific to strings and serves some use cases.
https://github.com/noobaa/noobaa-operator/blob/d821724e89d182e69ef05c7c533398c9c4a5241c/pkg/util/util.go#L1328. 

@utkarsh-pro, to continue our discussion, please see the function `contains` I added in this PR.
You also had an idea of making the generic function I suggested even more generic (not only for comparables) if you can give more details.

utkarsh-pro:
> 

That's a good suggestion @shirady. I think we can have something like this:

```go
func ContainsAny[T any](arr []T, item T, eq func(T, T) bool) bool {
	for _, data := range arr {
		if eq(data, item) {
			return true
		}
	}

	return false
}

func Contains[T comparable](arr []T, item T) bool {
	return ContainsAny[T](
		arr,
		item,
		func(a, b T) bool {
			return a == b
		},
	)
}
```

Regarding the current `util.Contains`, I think you are right. It got too specific to strings. Maybe the places which require such functionality can rely on the above `ContainsAny` and in the function can pass the trim while the rest can safely use `Contains` without any code changes.

shirady:
@dannyzaken I added the change that we talked about - the case where we delete the toleration from noobaa CR.
Not supported: In this PR if you remove a Toleration it will not pass to the backing store pod.
### Testing Instructions: 
Test on minikube
1. Change Noobaa CR and add Toleration to it:  `kubectl edit noobaa noobaa`.
2. Paste this inside  (2 tolerations):

```
tolerations:
  - effect: NoSchedule
     key: xyz
     operator: Equal
     value: "true"
  - effect: NoSchedule
     key: abc
     operator: Equal
     value: "true"
```
3. See the changes of the pods:  k`ubectl get pods -w`.
would expect to see the pods core, db, endpoint and backingstore to be terminated.
4. Delete `abc` toleration, by `kubectl edit noobaa noobaa` and deleting it.
5. See the changes of the pods:  `kubectl get pods -w`. backingstore would not be terminated and its toleration would include the `abc` toleration.

I don't know how to recognize the builtin toleration in a pod (I opened a [discussion](https://discuss.kubernetes.io/t/how-can-i-identify-the-default-toleration-of-a-pod/23234) in k8s community). Otherwise, I could iterate over the backing store pod's toleration and if this toleration is not default, check if it exists in noobaa CR (if not it was deleted and we will return true from `needUpdate`).

shirady:
@utkarsh, I added the change you suggested in `util`. You'll need to decide in which cases you use `Contains` and which `ContainsAny` and pass the function that uses `TrimSapces`. In my opinion, it should be in specific cases and not for every use of `Contains` of strings.

shirady:
I had a compilation error, so I separated the change of `Contains` into different commits.

pkg/backingstore/reconciler.go
```diff
@@ -1187,6 +1178,18 @@ func (r *Reconciler) needUpdate(pod *corev1.Pod) bool {
 		r.Logger.Warnf("Change in Image Pull Secrets detected: NoobaaSecret(%v) Spec(%v)", noobaaSecret, podSecrets)
 		return true
 	}
+
+	// In case noobaa CR was changed and a new toleration was added
+	// we want that change to pass to the backingstore pod as well
+	podTolerations := pod.Spec.Tolerations
+	noobaaTolerations := r.NooBaa.Spec.Tolerations
+	if len(noobaaTolerations) > 0 {
```
utkarsh-pro, line 1186: I think we can remove this check. Range does this check for us, even if the value is `nil` `range` will manage it just fine.

utkarsh-pro:
@shirady, I re-visited my multi-namespace bucketclass PR and it seems like we don't actually use `strings.Trim` there as well. The PR went through multiple iterations over the course of 9 - 10 months and at some point I got rid of whatever was relying on `strings.Trim`.

pkg/backingstore/reconciler.go
shirady: removed, thanks!

shirady:
> @shirady, I re-visited my multi-namespace bucketclass PR and it seems like we don't actually use `strings.Trim` there as well. The PR went through multiple iterations over the course of 9 - 10 months and at some point I got rid of whatever was relying on `strings.Trim`.

@utkarsh-pro Great, so we don't plan another change in a separate PR using the existing `Contains` call.

shirady:
@dannyzaken @utkarsh-pro I added a commit with the changes of the affinity issue (see here #1066).
I'm fine with squashing it and leaving it in this PR (rather than cherry-picking the commit to a different PR).

dannyzaken (approved):

utkarsh-pro (approved):

Issue #1040, opened 2023-01-31
DB-Dump Not Generated In Test Namespace Using PG DB


shirady:
### Environment info
NooBaa Operator Version: 5.13.0
Platform: Kubernetes 1.20.2 | minikube 1.18.1

### Actual behavior
1. DB-dump not generated.

### Expected behavior
1. Generate DB-dump, or at least inform the reason for the problem.

### Steps to reproduce
1. Use the steps in the workflow file to reproduce.

### More information - Screenshots / Logs / Other output

1. a [link](https://github.com/noobaa/noobaa-operator/actions/runs/4052201988/jobs/6971350284) to the failed job.
_Note: this link will be available under GH retention for saving the logs._
 
2. a Screenshot:
![Screenshot 2023-01-31 at 15 03 39](https://user-images.githubusercontent.com/57721533/215768547-0349b30d-8581-4258-9a3a-a850ed349413.png)
 
3. logs:
[logs_15156.zip](https://github.com/noobaa/noobaa-operator/files/10546642/logs_15156.zip)
[admission-tests-logs.zip](https://github.com/noobaa/noobaa-operator/files/10546641/admission-tests-logs.zip)

shirady:
This was created because I didn't define the namespace.
I would like to have a more informative message (in this case there are no pods in the namespace).

shirady:
Defining the namespace didn't solve it.
a [link](https://github.com/noobaa/noobaa-operator/actions/runs/4066128158) to the failed job.
_Note: this link will be available under GH retention for saving the logs._

logs:
[logs_15360.zip](https://github.com/noobaa/noobaa-operator/files/10566512/logs_15360.zip)
[admission-tests-logs.zip](https://github.com/noobaa/noobaa-operator/files/10566513/admission-tests-logs.zip)

shirady:
I found the RC and can solve it for PG DB.
It doesn't work for Mongo DB, the file is generated in the CLI output, but it does actually generated in the destination directory.

Issue #1007, opened 2023-01-03
Default Backing Store is in phase ready, but test pool cannot be used.


shirady:
### Environment info
CLI version: 5.13.0
noobaa-image: noobaa/noobaa-core:master-20220913
operator-image: noobaa/noobaa-operator:5.13.0

### Actual behavior
1. Default Backing Store is in phase ready, but test pool cannot be used.

### Expected behavior
1. Default Backing Store is in phase read and we can use test pool.

### Steps to reproduce
See in this [PR](https://github.com/noobaa/noobaa-core/pull/7142), take the changes from this commit, or use the instructions [here](https://gist.github.com/dannyzaken/c85f0006e5d6582c8d8666cafce6ab76). This issue is related to timing, so if every step will take a couple of minutes, it will not be reproduced (all steps must be done immediately or automatically). 
1. Based on the instructions [here](https://gist.github.com/dannyzaken/c85f0006e5d6582c8d8666cafce6ab76) - the steps of ‘Build images’ and ‘Deploy noobaa’.
2. I’ve noticed that after installing noobaa we need to wait before running the job, otherwise all the tests will be failed.
`kubectl wait --for=condition=available backingstore/noobaa-default-backing-store --timeout=3m`
**Do not wait a couple of minutes after this condition is met, immediately move to the next step.**
3. Deploy the tests job (run it from the noobaa-core repo):
` kubectl apply -f src/test/system_tests/ceph_s3_tests/test_ceph_s3_job.yml`
4. View the logs:
`kubectl logs job/noobaa-tests-s3 -f`

### More information - Screenshots / Logs / Other output
The test pool when it will be reproduce (and all the tests will be failed):

[...]

PR #1063, opened 2023-02-23
CI | Update Admission Tests Workflow

shirady:
### Explain the changes
1. Remove the comments about open issues that were closed (#1040 and #1007).
2. Add script to wait until we have the backingstore pod (in k8s we cannot wait for a resource that doesn't exist, more information [here](https://github.com/kubernetes/kubernetes/issues/83242)).
3. Update core repo to the latest commit.

1. none

### Testing Instructions:
1. none, will be tested in the workflow.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/run_admission_test.yml
+++ b/.github/workflows/run_admission_test.yml
@@ -17,7 +17,7 @@ jobs:
           # Freeze the version of core
           # to avoid a failed run due to code changes in the core repo.
           # Need to update the commit once in a while
-          ref: 293fefb9e755fa0d6bf1cc8b31f74259a0001730
+          ref: c97e4ddd2e5fcc110a3242c224eeea97d228c6eb
 
       - name: Checkout noobaa-operator
         uses: actions/checkout@v3
@@ -75,11 +75,12 @@ jobs:
           --endpoint-resources='{ "limits": {"cpu": "100m","memory": "1G"}, "requests": {"cpu": "100m","memory": "1G"}}' \
           --noobaa-image='noobaa-core:admission-test' -n test
           ./build/_output/bin/noobaa-operator status -n test
-          # we added the sleep since the test pool is in phase ready and condition available
-          # but the test pool storage is not ready yet, see issue:
-          # https://github.com/noobaa/noobaa-operator/issues/1007
-          sleep 3m
-          kubectl wait --for=condition=available backingstore/noobaa-default-backing-store --timeout=3m -n test
+
+      - name: Wait for phase Ready in the backingstore pod
+        run: |
+          cd ./noobaa-operator
+          ./.travis/number_of_pods_in_system.sh --namespace test --pods 5
+          kubectl wait --for=condition=available backingstore/noobaa-default-backing-store --timeout=5m -n test
 
       - name: Run Admission test
         run: |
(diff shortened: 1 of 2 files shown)


shirady:
I converted it to a draft Pr because the script runs well on my machine and not in the workflow, need to check it.

number_of_pods_in_system.sh
```diff
@@ -0,0 +1,41 @@
+#!/bin/sh
+
+# This script checks the number of pods in a given namespace
+# It waits untill we have the minimum number of pods to start working on the system
+# 
+# It was created simnce after nb install sometimes the backingstore is not created yet
+# In k8s we cannot wait for a resource that doesn't exist
```
liranmauda, line 7: I think we can remove it.
or rephrase completely.

number_of_pods_in_system.sh
```diff
@@ -0,0 +1,41 @@
...
+NUMBER_OF_ITERATIONS=15 # arbitrary number
+
+if [ $# -eq 0 ]; then
+  echo "Using default settings"
+elif [ $# -eq 1 ]; then
+  NAMESPACE=$1
+else # number of arguments is equal or greater than 2
+  NAMESPACE=$1
+  THESHOLD_PODS_NUMBER=$2
+fi
```
liranmauda, line 22:
```suggestion
while true; do
    case "${1}" in
        --pods) THESHOLD_PODS_NUMBER=${2}
                       shift 2 ;;
        --namespace)  NAMESPACE=${2}
                        shift 2 ;;
        --)             shift 1
                        break ;;
    esac
done
```

number_of_pods_in_system.sh
```diff
@@ -0,0 +1,41 @@
...
+
+echo "Check status of noobaa pods:"
+echo "Namespace: ${NAMESPACE}"
+echo "Number of pods to start using the system is ${THESHOLD_PODS_NUMBER}"
+
+i=1
+while [ "$i" -le ${NUMBER_OF_ITERATIONS} ];
+do
```
liranmauda, line 30:
```suggestion
if [ "$i" -le ${NUMBER_OF_ITERATIONS} ]
then
```

number_of_pods_in_system.sh
```diff
@@ -0,0 +1,41 @@
...
+    PODS_NUMBER=$(kubectl get pods -n ${NAMESPACE} --no-headers | wc -l)
+    echo ${PODS_NUMBER}
+    echo -ne "\033[1A\033[2K\033[1A" # move the cursor up 1 line and clear the line
+    if [ ${PODS_NUMBER} -ge ${THESHOLD_PODS_NUMBER} ]; then
+        echo "All needed pods were created!"
+        break
+    fi
+
+    echo "waiting for noobaa pods ${PODS_NUMBER}/${THESHOLD_PODS_NUMBER}..."
+    sleep 1
```
liranmauda, line 40: sleeping 1 means that when the `NUMBER_OF_ITERATIONS=15` it means we try only for 15 sec

number_of_pods_in_system.sh
```diff
@@ -0,0 +1,41 @@
...
+done
```
liranmauda, line 41: changing the while loop to if, we need to change respectfully:
```suggestion
fi
```

liranmauda (changes requested):

number_of_pods_in_system.sh
shirady: Why do you suggest changing it from loop to condition?

number_of_pods_in_system.sh
shirady: Yes, the cases when the pod is not created are rare, and you can see in the logs it is usually created on the first iteration, so 15 is a buffer.

number_of_pods_in_system.sh
shirady: I tried to add information, please tell me what is stil not clear.

number_of_pods_in_system.sh
shirady: Thanks, I added the change.

liranmauda (approved):

PR #1071, opened 2023-03-13
Master Root Key rotation for K8S secret backend

baum:
# KMS - Key Rotate

This is the interface for sharing keys between the NooBaa operator and the NooBaa core, endpoint pods for the Master Root Key rotation feature. [The Root key rotate - core side #7218](https://github.com/noobaa/noobaa-core/pull/7218)

## Sample rotating Master Root Key secret

The secret will contain a list of keys and the pointer to the current key.

```yaml
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  name: noobaa-root-master-key-volume
  namespace: default
data:
  active_root_key: a2V5LTE2Nzg4MjYyMzE1Mjk3ODk0Mzc=
  key-1678826231529789437: cTJ1SG5pbXJYeHViVTJTUVkwZmZ6eWVhMFNQSHAwLzNUMS9Wbi9vcGNPVT0=
```

## NooBaa core, endpoint pods view

The keys secret volume is mounted into the core, endpoint pods as a volume,
under `/etc/noobaa-server/root_keys` directory

### Pod shell view, 

```shell
bash-4.4$ cd /etc/noobaa-server/root_keys
bash-4.4$ ls -l
total 0
lrwxrwxrwx 1 root root 22 Mar 14 20:37 active_root_key -> ..data/active_root_key
lrwxrwxrwx 1 root root 30 Mar 14 20:37 key-1678826231529789437 -> ..data/key-1678826231529789437
bash-4.4$ for f in *; do echo "$f -> $(cat $f)"; done
active_root_key -> key-1678826231529789437
key-1678826231529789437 -> q2uHnimrXxubU2SQY0ffzyea0SPHp0/3T1/Vn/opcOU=

```

### DIFF
--- a/deploy/crds/noobaa.io_noobaas_crd.yaml
+++ b/deploy/crds/noobaa.io_noobaas_crd.yaml
@@ -1484,6 +1484,10 @@ spec:
                         additionalProperties:
                           type: string
                         type: object
+                      enableKeyRotation:
+                        type: boolean
+                      schedule:
+                        type: string
                       tokenSecretName:
                         type: string
                     type: object
@@ -1605,6 +1609,11 @@ spec:
                 - readyCount
                 - virtualHosts
                 type: object
+              lastKeyRotateTime:
+                description: LastKeyRotateTime is the time system ran an encryption
+                  key rotate
+                format: date-time
+                type: string
               observedGeneration:
                 description: ObservedGeneration is the most recent generation observed
                   for this noobaa system. It corresponds to the CR generation, which
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -30,7 +30,7 @@ const (
 	// ContainerImageRepo is the repo of the default image url
 	ContainerImageRepo = "noobaa-core"
 	// ContainerImageTag is the tag of the default image url
-	ContainerImageTag = "master-20230123"
+	ContainerImageTag = "master-20230426"
 	// ContainerImageSemverLowerBound is the lower bound for supported image versions
 	ContainerImageSemverLowerBound = "5.0.0"
 	// ContainerImageSemverUpperBound is the upper bound for supported image versions
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -16,6 +16,7 @@ import (
 	secv1 "github.com/openshift/api/security/v1"
 	cloudcredsv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
+	"github.com/robfig/cron/v3"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -52,7 +53,6 @@ func (r *Reconciler) ReconcilePhaseCreating() error {
 	if err := r.ReconcileCAInject(); err != nil {
 		return err
 	}
-
 	if err := r.ReconcileObject(r.ServiceAccount, r.SetDesiredServiceAccount); err != nil {
 		return err
 	}
@@ -444,7 +444,6 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 		case "NODE_EXTRA_CA_CERTS":
 			c.Env[j].Value = r.ApplyCAsToPods
 		}
-
 	}
 }
 
@@ -476,6 +475,7 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 			// adding the missing Env variable from default container
 			util.MergeEnvArrays(&c.Env, &r.DefaultCoreApp.Env)
 			r.setDesiredCoreEnv(c)
+			r.setDesiredRootMasterKeyMounts(podSpec, c)
 
 			util.ReflectEnvVariable(&c.Env, "HTTP_PROXY")
 			util.ReflectEnvVariable(&c.Env, "HTTPS_PROXY")
@@ -801,6 +801,24 @@ func (r *Reconciler) setKMSConditionStatus(s corev1.ConditionStatus) {
 	r.Logger.Infof("setKMSConditionStatus %v", s)
 }
 
+func (r *Reconciler) reportKMSConditionStatus() {
+	conditions := &r.NooBaa.Status.Conditions
+	cond := conditionsv1.FindStatusCondition(*conditions, nbv1.ConditionTypeKMSStatus)
+	if cond == nil {
+		util.Logger().Printf("❌ Missing KMS status: %q\n", nbv1.ConditionTypeKMSStatus)
+		return
+	}
+
+	st := cond.Status
+	var kmsStatusBadge string
+	if kms.StatusValid(st) {
+		kmsStatusBadge = "✅"
+	} else {
+		kmsStatusBadge = "❌"
+	}
+	util.Logger().Printf("%v KMS status: %q\n", kmsStatusBadge, st)
+}
+
 func (r *Reconciler) setKMSConditionType(t string) {
 	conditions := &r.NooBaa.Status.Conditions
 	conditionsv1.SetStatusCondition(conditions, conditionsv1.Condition{
@@ -813,7 +831,6 @@ func (r *Reconciler) setKMSConditionType(t string) {
 
 // ReconcileRootSecret choose KMS for root secret key
 func (r *Reconciler) ReconcileRootSecret() error {
-
 	// External KMS Spec
 	connectionDetails := r.NooBaa.Spec.Security.KeyManagementService.ConnectionDetails
 	authTokenSecretName := r.NooBaa.Spec.Security.KeyManagementService.TokenSecretName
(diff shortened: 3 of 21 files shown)


pkg/util/kms/kms.go
```diff
@@ -2,47 +2,60 @@ package kms
...
 }
 
 // Driver is a backend type specific driver interface for libopenstorage/secrets framework
 type Driver interface {
-	Path()          string
-	Name()          string
-	Config(connectionDetails map[string]string, tokenSecretName, namespace string)  (map[string]interface{}, error)
-	GetContext()    map[string]string
-	SetContext()    map[string]string
+	Path() string
```
jackyalbo, line 44: have to say I prefer the beautified version :) but whatever...

pkg/util/kms/kms_version.go
```diff
@@ -0,0 +1,207 @@
...
+	return nil
+}
+
+// Upgrade implements Version interface for the secret map, i.e. rotating master root key
+func (v *VersionRotatingSecret) Upgrade() error {
+	v1 := VersionSingleSecret{v.k, nil}
+	err := v1.Get()
+	if err != nil {
+		if err == secrets.ErrInvalidSecretId {
+			return nil // nothing to upgraxde
```
jackyalbo, line 171:
```suggestion
			return nil // nothing to upgrade
```

.github/workflows/run_admission_test.yml
```diff
@@ -9,15 +9,6 @@ jobs:
...
     steps:
-      - name: Checkout noobaa-core
-        uses: actions/checkout@v3
-        with:
-          repository: 'noobaa/noobaa-core'
-          path: "noobaa-core"
-          # Freeze the version of core
-          # to avoid a failed run due to code changes in the core repo.
-          # Need to update the commit once in a while
-          ref: c97e4ddd2e5fcc110a3242c224eeea97d228c6eb
```
shirady, line 20: @baum, Is it a temporary change?
We freeze the commit of noobaa-core on purpose because we want the CI to check only the Operator changes.

.github/workflows/run_admission_test.yml
```diff
@@ -51,11 +42,6 @@ jobs:
           sudo chown -R $USER $HOME/.kube $HOME/.minikube
           sed "s/root/home\/$USER/g" $HOME/.kube/config > tmp; mv tmp $HOME/.kube/config
 
-      - name: Build noobaa image
-        run: |
-          cd ./noobaa-core
-          make noobaa NOOBAA_TAG=noobaa-core:admission-test
```
shirady, line 57: Why not explicitly build it?
I think It is easier to debug in case we have a problem (step `Build noobaa image` is separated from `Build operator image`).

.github/workflows/run_admission_test.yml
baum: Master Root Key changes interface with the core in regards with sharing the key, that SHA/ref would of core would not able to run. 

Generally, the admission test froze the version of the core in the GitHub action, possibly to ensure stability during testing. However, the comment suggests that it is necessary to update the commit periodically to keep up with changes or fixes in the codebase.

```shell
          # Need to update the commit once in a while
```

Use the operator's default NooBaa core image, which may not be the most recent version of the NooBaa core but is presumably stable enough to use.

.github/workflows/run_admission_test.yml
baum: If using the operator's default core image, building core image is not required. The image is fetched from the docker hub.

.github/workflows/run_admission_test.yml
baum: Extracted admission test change into a separate [PR](https://github.com/noobaa/noobaa-operator/pull/1076).

pkg/options/options.go
```diff
@@ -26,11 +26,11 @@ func RunOptions(cmd *cobra.Command, args []string) {
 
 const (
 	// ContainerImageOrg is the org of the default image url
-	ContainerImageOrg = "noobaa"
+	ContainerImageOrg = "alexanderindenbaum"
 	// ContainerImageRepo is the repo of the default image url
 	ContainerImageRepo = "noobaa-core"
 	// ContainerImageTag is the tag of the default image url
-	ContainerImageTag = "master-20230123"
+	ContainerImageTag = "jacky-root-key-rotate"
```
dannyzaken, line 33: remember to update core image once the core PR is merged

dannyzaken (changes requested):

deploy/crds/noobaa.io_noobaas_crd.yaml
```diff
@@ -1363,6 +1363,8 @@ spec:
                         additionalProperties:
                           type: string
                         type: object
+                      schedule:
```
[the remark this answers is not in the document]
jackyalbo: So about this I think we wanted to have some kind of default - 1 month. so If the user doesn't provide schedule but does set enabled to true we will run it once a month. about none supporting KMS I think the default will be never. and in anyway even if the user set to something else in schedule or in enabled/disabled we will keep with never for now. Does that make senes?

pkg/system/phase2_creating.go
```diff
@@ -812,30 +811,152 @@ func (r *Reconciler) ReconcileRootSecret() error {
...
+func (r *Reconciler) ReconcileKeyRotation() error {
+	r.Logger.Infof("ReconcileKeyRotation, KMS Starting")
+
+	if len(r.SecretRootMasterKey) > 0 {
+		// key rotation not supported by the KMS backend
+		r.Logger.Infof("ReconcileKeyRotation, KMS skip reconcile single master root mode")
+		return nil
+	}
+
+	keyManagementServiceSpec := r.NooBaa.Spec.Security.KeyManagementService
```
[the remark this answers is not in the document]
jackyalbo: I think we need both flags in case the user wants to set a schedule, but wants to disable it after some time. And for the option to stop key rotation totally. I think what I want to say is:
if enabled and schedule not set => default should be used.
if disable and schedule set or not set => the rotate will never run until will be set to enabled.
the default will be enabled, and schedule of once a month.

pkg/system/reconciler.go
```diff
@@ -307,6 +311,9 @@ func (r *Reconciler) CheckAll() {
 	util.KubeCheck(r.SecretServer)
 	util.KubeCheck(r.SecretOp)
 	util.KubeCheck(r.SecretEndpoints)
+	if len(r.SecretRootMasterKey) == 0 {
```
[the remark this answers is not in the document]
jackyalbo: but where do we set it? we are setting it only in phase2_creating ... am I missing something?

pkg/system/phase2_creating.go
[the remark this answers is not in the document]
baum: I have two concerns with `EnableKeyRotation`:
- The possibility of inconsistent state if rotation is enabled by configuration but not supported by the backend.
- The fact that it is a new mandatory field, which would render existing CRs invalid and cause issues with upgrades.
wdyt?

pkg/system/reconciler.go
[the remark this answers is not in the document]
baum: There is a `SecretReconciler` interface:
https://github.com/baum/noobaa-operator/blob/0cceb836c8692b9127fafaa6701df35eac56c66b/pkg/util/kms/kms.go#L33-L40
Implemented by the system reconciler, the string value is set here:
https://github.com/baum/noobaa-operator/blob/0cceb836c8692b9127fafaa6701df35eac56c66b/pkg/system/phase2_creating.go#L956-L963
Invoked from here, by the "single string" KMS backends :
https://github.com/baum/noobaa-operator/blob/0cceb836c8692b9127fafaa6701df35eac56c66b/pkg/util/kms/kms_version.go#L34-L38

pkg/system/phase2_creating.go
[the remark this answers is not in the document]
baum: Re-added optional EnableKeyRotation, per [OCS operator PR]( https://github.com/red-hat-storage/ocs-operator/pull/1926)

pkg/options/options.go
baum: ✅ [here](https://github.com/baum/noobaa-operator/blob/ace19b95668f5acc89288657104b81b57bf041b9/pkg/options/options.go#L28-L33)

pkg/system/reconciler.go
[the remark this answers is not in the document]
jackyalbo: again about this as r.SecretRootMasterKey is empty when creating a new reconciler - it will only be set when getting to phase2 (take a look we don't run the reconcile - we just create an instance of it in system.go:590). I don't think this check will help. Maybe we should use KubeCheckOptional and remove this if? WDYT? or check the KMS type somehow?

jackyalbo (approved):

pkg/system/reconciler.go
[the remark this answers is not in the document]
baum: Great point. I see no great value on reporting K8S secret existence, [exposed](https://github.com/noobaa/noobaa-operator/compare/83c199048e730e8ff771cf0b9f7d044829da4787..9c38cac0e8ffcca50c3a320deb09e2335eeeb3ca) KMS status condition instead. Thank you, Jacky 🖖

PR #1076, opened 2023-03-27
Admission test - release NooBaa core image

baum:
# Admission test - release NooBaa core image

The admission test froze the version of the core in the GitHub action, possibly to ensure stability during testing. However, the comment suggests that it is necessary to update the commit periodically to keep up with changes or fixes in the codebase.

```shell
          # Need to update the commit once in a while
```

Use the operator's default NooBaa core image, defined by [options](https://github.com/noobaa/noobaa-operator/blob/f7358501716816250702d9a4c96f2526f98534e7/pkg/options/options.go#L33), which may not be the most recent version of the NooBaa core but is presumably stable enough to use.

### DIFF
--- a/.github/workflows/run_admission_test.yml
+++ b/.github/workflows/run_admission_test.yml
@@ -9,15 +9,6 @@ jobs:
       group: ${{ github.workflow }}-${{ github.ref }}
       cancel-in-progress: true
     steps:
-      - name: Checkout noobaa-core
-        uses: actions/checkout@v3
-        with:
-          repository: 'noobaa/noobaa-core'
-          path: "noobaa-core"
-          # Freeze the version of core
-          # to avoid a failed run due to code changes in the core repo.
-          # Need to update the commit once in a while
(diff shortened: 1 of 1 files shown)


.github/workflows/run_admission_test.yml
```diff
@@ -9,15 +9,6 @@ jobs:
...
     steps:
-      - name: Checkout noobaa-core
-        uses: actions/checkout@v3
-        with:
-          repository: 'noobaa/noobaa-core'
-          path: "noobaa-core"
-          # Freeze the version of core
-          # to avoid a failed run due to code changes in the core repo.
-          # Need to update the commit once in a while
-          ref: c97e4ddd2e5fcc110a3242c224eeea97d228c6eb
```
shirady, line 20: @baum, Why not simply update the commit?

dannyzaken: I agree with removing it. I think it's better to maintain one common place (currently the default image in `options.go`) to determine the test image.

PR #1078, opened 2023-03-29
Replace ownership change init container with K8s FSGroup

utkarsh-pro:
### Explain the changes

This PR attempts to replace FS ownership change init container with K8s FSGroup in an attempt to:
1. Fix https://github.com/noobaa/noobaa-core/issues/7030.

FSGroup unfortunately is not a silver bullet:
- Its usage helps us gets rid of init container but that just delegates the responsibility of changing permissions and setting owner to kubelet. In theory, this should be more performant because CRI doesn't have to spin up a container just so that we can do our permission adjustment rather kubelet does that automatically on the mounted volume.
- Kubelet will perform 2 actions on the mounted directory
    - Change the group of the owner to the `fsGroup` passed in pod spec. Kubelet does this simply by running `os.Lchown(file, -1, int(*fsGroup))` which is run as a `root` user. This should work most of the time HOWEVER in the issue linked above, it WILL fail because any operation performed by the root user will be executed as `nobody:nogroup` by NFS. This does NOT causes any issues however because kubelet just logs the error instead of erroring out. This works out in our favor in THIS case.
    - Change the directory permissions to `0660 | os.ModeSetgid | 0110` which ensures that we get `drwxdrwx--` permission on the mounted directory which is good enough for us because we run group ID `0` while `os.ModeSetgid` set the first octal to `4` implying that directories and files created within this directory will inherit the group ownership. This suffers the same problem in NFS (root_squash) mounts but because kubelet will not error out, it should work just fine.
 
So this does NOT exactly fixes out problems with NFS mounts but makes sure that our DB pod doesn't ends up in CLBO because of permission issues but will crash if our second init container fails to do its thing (which it will if NFS mount does not provide enough permissions - which I think are available in the above issue's environment)

Reference: [Kubernetes FSGroup Source Code](https://github.com/kubernetes/kubernetes/blob/ad18954259eae3db51bac2274ed4ca7304b923c4/pkg/volume/volume_linux.go#L43)

#### Tests Performed
1. Deploy and run in fresh Minikube Cluster.
2. Deploy and run in fresh OpenShift Cluster (AWS gp2 storageclass).
3. Deploy and run in fresh OpenShift Cluster (AWS NFS mount with `drwxdrwxdrwx` mount permissions and `root_squash` enabled).

Fixes https://github.com/noobaa/noobaa-core/issues/7030.

### Testing Instructions:
Passing pre-existing CLI tests should be good enough indication as only deployment has changed.

### DIFF
--- a/deploy/internal/statefulset-db.yaml
+++ b/deploy/internal/statefulset-db.yaml
@@ -20,25 +20,6 @@ spec:
     spec:
       serviceAccountName: noobaa-db
       terminationGracePeriodSeconds: 60
-      initContainers:
-      #----------------#
-      # INIT CONTAINER #
-      #----------------#
-      - name: init
-        image: NOOBAA_CORE_IMAGE
-        command:
-        - /noobaa_init_files/noobaa_init.sh
-        - init_mongo
-        resources:
-          requests:
-            cpu: "500m"
-            memory: "500Mi"
-          limits:
-            cpu: "500m"
-            memory: "500Mi"
-        volumeMounts:
-        - name: db
-          mountPath: /mongo_data
       containers:
       #--------------------#
       # DATABASE CONTAINER #
--- a/deploy/internal/statefulset-postgres-db.yaml
+++ b/deploy/internal/statefulset-postgres-db.yaml
@@ -23,21 +23,6 @@ spec:
       #-----------------#
       # INIT CONTAINERS #
       #-----------------#
-      - name: init
-        image: NOOBAA_CORE_IMAGE
-        command:
-        - /noobaa_init_files/noobaa_init.sh
-        - init_postgres
-        resources:
-          requests:
-            cpu: "500m"
-            memory: "500Mi"
-          limits:
-            cpu: "500m"
-            memory: "500Mi"
-        volumeMounts:
-        - name: db
-          mountPath: /var/lib/pgsql
       - name: initialize-database
         image: NOOBAA_DB_IMAGE
         env:
@@ -122,6 +107,8 @@ spec:
       securityContext: 
         runAsUser: 10001
         runAsGroup: 0
+        fsGroup: 0
+        fsGroupChangePolicy: "OnRootMismatch"
   volumeClaimTemplates:
     - metadata:
         name: db
--- a/deploy/scc_db.yaml
+++ b/deploy/scc_db.yaml
@@ -14,7 +14,7 @@ allowedCapabilities:
 - SETUID
 - SETGID
 fsGroup:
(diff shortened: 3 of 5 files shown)


baum:
Hello 👋 Ukarsh,

You have described test scenarios verbosely, however all reuse the existing PV. I'm wondering about clean install scenario in OpenShift environment. Is DB pod without ownership change init container able to use a fresh PV which was just provisioned for it?

For instance, for the "2. Normal NooBaa deployment in OpenShift" what happens if we change "ii. Delete the DB Pod." to "ii. Delete the DB Pod **and PVC**" making sure DB gets a fresh PV?

utkarsh-pro:
Hi @baum! That's a good test! I haven't performed it yet. Lemme get back to you once I am done testing this. Please lemme know if there are more tests that I should perform here.

deploy/internal/statefulset-db.yaml
```diff
@@ -20,25 +20,6 @@ spec:
...
       terminationGracePeriodSeconds: 60
-      initContainers:
-      #----------------#
-      # INIT CONTAINER #
-      #----------------#
-      - name: init
-        image: NOOBAA_CORE_IMAGE
-        command:
-        - /noobaa_init_files/noobaa_init.sh
-        - init_mongo
```
shirady, line 31: In case we don't use the `init_mongo` argument (and the function it calls `prepare_mongo_pv` iniside `noobaa_init.sh` file) - maybe edit it in this PR.

utkarsh-pro:
Under development - ignore for review ATM.

utkarsh-pro:
I don't think that golangci-lint is upset with the PR rather something related to https://github.com/actions/runner-images/discussions/7188? 🤔

deploy/internal/statefulset-db.yaml
utkarsh-pro: I am not sure if I completely understand.

shirady: Do we still use `init_mongo` function somewhere?

utkarsh-pro: No, we don't. If this PR gets in, I will remove it from the core as well.

baum:
> I don't think that golangci-lint is upset with the PR rather something related to [actions/runner-images#7188](https://github.com/actions/runner-images/discussions/7188)? 🤔

Please rebase, hope #1088 fixes the golangci-lint issue

utkarsh-pro:
Thank you @baum! It fixed the test.

dannyzaken (approved):

PR #1086, opened 2023-04-04
Fix Bug BZ 2183480 | Update AWS Regions

shirady:
### Explain the changes
1. Update AWS regions.

### Issues: Fixed [BZ 2183480](https://bugzilla.redhat.com/show_bug.cgi?id=2183480)
1. Update AWS regions.

### Testing Instructions:
1. none.

- [ ] Doc added/updated
- [ ] Tests added


dannyzaken (approved):

dannyzaken:
> @dannyzaken, I didn't know if we would also like to update the General endpoint (like we have `ec2)

I think we should consider returning a default zone if we can't map the nodes zone to a valid one. In any case it should be in a different PR, so we can backport this one if necessary

Issue #1017, opened 2023-01-15
Add a flag for developer resource


shirady:
**Is your feature request related to a problem? Please describe.**
No, it will only be more comfortable for developers to use.
We suggest adding flags for `nb install` in addition to `mini`:
1. **m1-dev resources**: today I'm working with `mini` flag by changing the values of the resource in the file `system.go`.
```
	if options.MiniEnv {
		coreResourceList := corev1.ResourceList{
			corev1.ResourceCPU:    *resource.NewScaledQuantity(int64(500), resource.Milli),
			corev1.ResourceMemory: *resource.NewScaledQuantity(int64(1), resource.Giga),
		}
		dbResourceList := corev1.ResourceList{
			corev1.ResourceCPU:    *resource.NewScaledQuantity(int64(800), resource.Milli),
			corev1.ResourceMemory: *resource.NewScaledQuantity(int64(1), resource.Giga),
		}
		endpointResourceList := corev1.ResourceList{
			corev1.ResourceCPU:    *resource.NewScaledQuantity(int64(500), resource.Milli),
			corev1.ResourceMemory: *resource.NewScaledQuantity(int64(1), resource.Giga),
		}
```
3. **VM resources**: today in our VM (which we use in GH Action) we install noobaa using this long command: 
```
         ./build/_output/bin/noobaa-operator crd create
         ./build/_output/bin/noobaa-operator operator install
         ./build/_output/bin/noobaa-operator system create  \
         --db-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
         --core-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
         --endpoint-resources='{ "limits": {"cpu": "200m","memory": "1G"}, "requests": {"cpu": "200m","memory": "1G"}}' \
         --noobaa-image='noobaa-core:s3-tests'
         ./build/_output/bin/noobaa-operator status
```

**Describe the solution you'd like**
I would like to have 2 new flags that will describe the needed resources \(in addition to `mini`), and that I could simply run them instead of the bypasses I'm using today.

**Describe alternatives you've considered**
As mentioned before there are 2 options:

1. If you can change the code locally - change the resources of `mini` flag and then run `nb install --mini `.
2. Otherwise, run the commands inside the install and change the resources as we did above (VM resource example).

PR #1094, opened 2023-04-19
added --dev flag to the cli

dannyzaken:
### Explain the changes
1. added `--dev` flag to noobaa CLI
2. same as `--mini` but with higher resources requests tuned for a dev environment (mainly M1 macs)

1. Fixes https://github.com/noobaa/noobaa-operator/issues/1017

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -110,6 +110,11 @@ var ImagePullSecret = ""
 // pod)
 var MiniEnv = false
 
+// DevEnv setting this option indicates to the operator that it is deployed on development environment
+// This info is used by the operator for environment based decisions (e.g. number of resources to request per
+// pod)
+var DevEnv = false
+
 // DisableLoadBalancerService is used for setting the service type to ClusterIP instead of LoadBalancer
 var DisableLoadBalancerService = false
 
@@ -208,6 +213,10 @@ func init() {
 	FlagSet.BoolVar(
 		&MiniEnv, "mini",
 		false, "Signal the operator that it is running in a low resource environment",
+	)
+	FlagSet.BoolVar(
+		&DevEnv, "dev",
+		false, "Set sufficient resources for dev env",
 	)
 	FlagSet.BoolVar(
 		&DisableLoadBalancerService, "disable-load-balancer",
--- a/pkg/system/system.go
+++ b/pkg/system/system.go
@@ -213,6 +213,36 @@ func LoadSystemDefaults() *nbv1.NooBaa {
 				Limits:   endpointResourceList,
 			}}
 	}
+	if options.DevEnv {
+		coreResourceList := corev1.ResourceList{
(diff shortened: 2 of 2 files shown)


v-harihar (approved):
just the comment in wrong place. apart from that lgtm

pkg/options/options.go
```diff
@@ -110,6 +110,11 @@ var ImagePullSecret = ""
 // pod)
 var MiniEnv = false
 
+// MiniEnv setting this option indicates to the operator that it is deployed on low resource environment
```
liranmauda, line 113:
```suggestion
// DevEnv setting this option indicates to the operator that it is deployed on low resource environment
```

liranmauda (commented):
@dannyzaken Why should we keep the MiniEnv (--mini) flag?

pkg/options/options.go
shirady: Please update the text (and not only `MiniEnv` to `DevEnv`).

pkg/options/options.go
```diff
@@ -209,6 +214,10 @@ func init() {
 		&MiniEnv, "mini",
 		false, "Signal the operator that it is running in a low resource environment",
 	)
+	FlagSet.BoolVar(
+		&DevEnv, "dev",
+		false, "Set sufficient resoureces for dev env (Arm Macs)",
```
shirady, line 219: Please change `Arm` to `ARM`

pkg/options/options.go
```diff
@@ -110,6 +110,11 @@ var ImagePullSecret = ""
 // pod)
 var MiniEnv = false
 
+// MiniEnv setting this option indicates to the operator that it is deployed on low resource environment
+// This info is used by the operator for environment based decisions (e.g. number of resources to request per
+// pod)
+var DevEnv = false
```
shirady, line 116: Why is it a global flag?
I thought we use it only in `install`...

shirady:
@dannyzaken, would you please add this issue #1017 as fixed in the description?

dannyzaken:
> @dannyzaken Why should we keep the MiniEnv (--mini) flag?

it is used by our tests. at first I tried to increase the resources for `--mini`, but the images in the actions don't have enough resources

pkg/options/options.go
dannyzaken: decided to drop the reference to arm altogether

dannyzaken:
> General comment - I think [it](https://github.com/noobaa/noobaa-operator/actions/runs/4740342204/jobs/8419275980?pr=1094) is an example of why we should run linters (`make lint` & `golangci-lint `) separated from the tests themselves (we could see that it doesn't break anything and we have just a lint error about a comment).

actually, for the operator we do have a separate action for the linter, but it passed. don't know what are the differences from the makefile one

PR #1093, opened 2023-04-18
CI | Exclude 'docs' Directory From GitHub Workflows 

shirady:
### Explain the changes
1. Add `workflow_dispatch` option where we didn't have it.
2. Exclude `doc` directory from the CI (for more information you can see [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths)).

### Issues: Fixed #xxx / Gap #xxx
1. none

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

PR #1103, opened 2023-04-24
CI | Undo Exclude Doc Directory from CI

shirady:
### Explain the changes
1. Undo exclude doc directory from CI. After this change all workflows will run on every change (even if it is only documentation).
This was added in #1093, but it doesn't work well with required tests. We decided not to use the workaround to solve it (as described [here](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/managing-protected-branches/troubleshooting-required-status-checks#handling-skipped-but-required-checks)). 

1. none

### Testing Instructions:
1. none

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,12 +1,5 @@
 name: golangci-lint
-on:
-  push:
-    paths-ignore:
-      - 'doc/**'
-  pull_request:
-    paths-ignore:
-      - 'doc/**'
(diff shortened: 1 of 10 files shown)


shirady:
@nimrod If you don't see a problem please merge without waiting for the checks to run.

liranmauda (approved):

nimrod-becker (approved):

PR #1104, opened 2023-04-24
Fix init container cleanup, Rollback lib-bucket-provisioner, Fix autoscaler

utkarsh-pro:
### Explain the changes
This PR attempts to fix 3 items:
1. Init container cleanup - In the previous PR I missed the cleanup of the unused init container. This PR fixes that **but** ramification of the fix is that now a user cannot create an initcontainer to the DB pod whose name is "init". Not a big deal though, IMO. This PR also reconciles the FSGroups and FSChangePolicy as we do not expect those to change either.
2. Rollback lib-bucket-provisioner: lib-bucket-provisioner made a breaking change to the library, this is being rollback as a temporary solution.
3. Autoscaler encounters a RPC error due to invalid endpoint count - it is being defaulted to min: 1, max: 2 (previous values).

- [ ] Doc added/updated
- [ ] Tests added


naveenpaul1:
LGTM

dannyzaken (approved):

PR #1109, opened 2023-04-25
Fixing issue with operator log level

jackyalbo:
### Explain the changes
1. moved the code to after we reconcile the noobaa-config config map, before it was always empty

1. Fixed BZ#1932846

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -18,6 +18,7 @@ import (
 	conditionsv1 "github.com/openshift/custom-resource-status/conditions/v1"
 	"github.com/robfig/cron/v3"
 	cephv1 "github.com/rook/rook/pkg/apis/ceph.rook.io/v1"
+	"github.com/sirupsen/logrus"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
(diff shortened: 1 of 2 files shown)


pkg/system/phase2_creating.go
```diff
@@ -93,6 +94,13 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 		return err
 	}
 
+	if r.CoreAppConfig.Data["NOOBAA_LOG_LEVEL"] == "warn" {
```
dannyzaken, line 97: This will only be running on noobaa CR reconcile. I think it should work because AFAIR we reconcile noobaa CR on every system store update notification, but maybe we should think of a better solution

pkg/system/phase2_creating.go
```diff
@@ -93,6 +94,13 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
+		r.Logger.Infof("Setting operator log level to Warn")
+		util.InitLogger(logrus.WarnLevel)
+	} else {
+		util.InitLogger(logrus.DebugLevel)
```
utkarsh-pro, line 101: The default is going to be debug level?

pkg/system/phase2_creating.go
utkarsh-pro: I think we do enqueue a reconcile if noobaa owned configmap is updated.

jackyalbo: this time it works - its a fix for my previous PR which wasn't actually running. For sure we can have a better solution - I think we need to completely redesign our logging for both operator and core.

dannyzaken (approved):
LGTM

utkarsh-pro (approved):

liranmauda (approved):

Issue #1035, opened 2023-01-25
Could not delete backing store in phase Rejected


shirady:
### Environment info
CLI version: 5.13.0
noobaa-image: noobaa/noobaa-core:master-20220913
operator-image: noobaa/noobaa-operator:5.13.0

### Actual behavior
1. Could not delete backing store with phase Rejected.

### Expected behavior
1. Delete the backing store with phase Rejected successfully.

### Steps to reproduce
1. `nb backingstore create aws-s3 help` (help is the name of the backing store).
2. Enter a wrong access-key (for example a link to a web).
3. See the new backing store (help) in phase `Rejected` using `nb backingstore list`
4. Try to delete it using `nb backingstore delete help` (as mentioned, help is the name of the backing store).

### More information - Screenshots / Logs / Other output
`nb backingstore create aws-s3 help`
You can see that I accidentally pasted a wrong access key (you can see the length is 129 instead of 20.
![Screenshot 2023-01-25 at 15 54 44](https://user-images.githubusercontent.com/57721533/214581851-7c8edd41-7a7a-40e0-9b82-b4aeec8d8dd3.png)

`nb backingstore list` - see the new backing store (help) in phase `Rejected`
![Screenshot 2023-01-25 at 15 55 24](https://user-images.githubusercontent.com/57721533/214582143-ebe02a4c-81e9-422c-b061-1265cb7d84d7.png)

`nb backingstore delete help` and then again `nb backingstore list`
![Screenshot 2023-01-25 at 16 00 47](https://user-images.githubusercontent.com/57721533/214583301-53a0cc38-fa75-4028-8845-3d2799c3947b.png)
It doesn't find the pool to delete it, but it could list it.

shirady:
I saw that we do not support deletion in case of credentials are invalid, but it seems that we want to suggest a manual resolution (not in the case above).
From our [docs](https://github.com/noobaa/noobaa-operator/blob/master/doc/backing-store-crd.md):

> "There are cases where the decommissioning cannot be complete due to inability to read the data from the backing-store that is already not serving - for example... the credentials were invalidated ... In such cases the system status will be used to report these issues and suggest a manual resolution..."

nadavMiz:
I seems the issue doesn't for any wrong access-key. the access should contain no graphical characters. for example I used a secret key containing Cntrl+d Cntrl+f Cntrl+g

so the steps should be as followed:
1. nb backingstore create aws-s3 help (help is the name of the backing store).
[...]

PR #1115, opened 2023-05-01
ignore ReadSystemInfo persistent error in deletion

nadavMiz:
### Explain the changes
1. this issue is caused by ReadSystemInfo failing on deletion. this should not cause the deletion to fail. added a check to continue deleting if the error is persistent 

1. https://github.com/noobaa/noobaa-operator/issues/1035 

### Testing Instructions:
1. testing instructions in issue

- [ ] Doc added/updated
- [ ] Tests added


baum (approved):
lgtm 🖖

PR #1117, opened 2023-05-02
Adding quantity quota alerts

v-harihar:
`NooBaaBucketExceedingSizeQuotaState` and `NooBaaBucketReachingQuantityQuotaState` are missing in the deployment and the result which the alerts are not getting triggered. Adding the above alerts to this patch.

Ref Patch: https://github.com/noobaa/noobaa-mixins/pull/23

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2154250

### DIFF
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -3726,7 +3726,7 @@ spec:
         claimName: noobaa-pv-claim
 `
 
-const Sha256_deploy_internal_prometheus_rules_yaml = "9fc9d30ce1f9ca180867255b363943b672689bb98139d6dcb12b63a95529ae0f"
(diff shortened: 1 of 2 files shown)


pkg/bundle/deploy.go
```diff
@@ -3857,6 +3857,30 @@ spec:
       for: 5m
       labels:
         severity: warning
+    - alert: NooBaaBucketReachingQuantityQuotaState
```
dannyzaken, line 3860: @v-harihar notice that deploy.go is generated, don't update it directly. you need to update `File_deploy_internal_prometheus_rules_yaml`

v-harihar: My bad

Issue #872, opened 2022-03-03
`ListAccountsAPI()` Call is incorrect and will lead to invalid schema


liranmauda:
`ListAccountsAPI()` Call is incorrect and will lead to invalid schema

```
	ListAccountsAPI() (ListAccountsReply, error)

        func (c *RPCClient) ListAccountsAPI() (ListAccountsReply, error) {
	        req := &RPCMessage{API: "account_api", Method: "list_accounts"}
```

It should get params:
```
	ListAccountsAPI(ListAccountsParams) (ListAccountsReply, error)

        func (c *RPCClient) ListAccountsAPI(params ListAccountsParams) (ListAccountsReply, error) {
	        req := &RPCMessage{API: "account_api", Method: "list_accounts", Params: params}
```

nimrod-becker:
This is not being called in the operator, so can either fix or remove

PR #1120, opened 2023-05-04
Fix incorrect function ListAccountsAPI()

nadavMiz:
### Explain the changes
1. fix incorrect function ListAccountsAPI(): add missing parameters

1. https://github.com/noobaa/noobaa-operator/issues/872

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


liranmauda (commented):
What is the motivation to remove it? why not do the simple fix?
how many unused API functions do we have? Can you list all of them? 
I think the simple way is just to fix it.

PR #1138, opened 2023-06-15
Endpoints  topology-spread-constraints added

naveenpaul1:
### Explain the changes
1. a new [topology-spread-constraints/](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/) is added for noobaa endpoint pod to make sure the endpoint pod spread across multiple nodes with a skew difference of 1.

1. https://issues.redhat.com/browse/MCGI-116

### Testing Instructions:
1. increase the number of endpoint pods either through HPA or manually through  `noobaa-endpoint` Deployment. Each pod should be spread in different nodes with a maximum skew difference of 1. One point to remember is `whenUnsatisfiable` is configured as `ScheduleAnyway` so if there is any CPU/Memory constraint in nodes that node will be skipped and the pod added to another node without considering topology-spread-constraints.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/doc/noobaa-crd.md
+++ b/doc/noobaa-crd.md
@@ -240,6 +240,22 @@ spec:
   dbConf: |+
     max_connections = 1000
 ```
+## Pod Topology Spread Constraints for Noobaa endpoint
+
+Noobaa operator will add the Topology Spread Constraints to Noobaa endpoint deployment to make sure endpoint pods are evenly spread across nodes. Cluster based on Kubernetes 1.25 and previous versions missing `nodeTaintsPolicy` in `topologySpreadConstraints` and pods are not evenly spread between nodes for these clusters. Noobaa operator checks for the Kubernates version and adds `topologySpreadConstraints` only if the Kubernates version is 1.26+. Operator also skips `topologySpreadConstraints` if this is already set on the deployment, or the CR annotation `noobaa.io/skip_topology_spread_constraints` is set to true.
+
+```yaml
+topologySpreadConstraints:
+  - maxSkew: 1
+    topologyKey: kubernetes.io/hostname
+    whenUnsatisfiable: ScheduleAnyway
+    labelSelector:
+      matchLabels:
+        noobaa-s3: noobaa
+    nodeTaintsPolicy: Honor
+
+```
+Users can make changes to `topologySpreadConstraints` configuration after the operator creates it and the changes will not override it. But once user remove the custome `topologySpreadConstraints` default value is resotored. 
 
 ## Notes
 1. `dbConf` field will have no effect if `dbType` is not "postgres".
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -14,8 +14,7 @@ import (
 	"encoding/json"
 
 	"cloud.google.com/go/storage"
-	"google.golang.org/api/option"
-
+	semver "github.com/coreos/go-semver/semver"
 	"github.com/marstr/randname"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
@@ -25,6 +24,7 @@ import (
 	secv1 "github.com/openshift/api/security/v1"
 	monitoringv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 	"github.com/sirupsen/logrus"
+	"google.golang.org/api/option"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
@@ -40,9 +40,10 @@ import (
 )
 
 const (
-	ibmEndpoint = "https://s3.direct.%s.cloud-object-storage.appdomain.cloud"
-	ibmLocation = "%s-standard"
-	ibmCOSCred  = "ibm-cloud-cos-creds"
+	ibmEndpoint                           = "https://s3.direct.%s.cloud-object-storage.appdomain.cloud"
+	ibmLocation                           = "%s-standard"
+	ibmCOSCred                            = "ibm-cloud-cos-creds"
+	topologyConstraintsEnabledKubeVersion = "1.26.0"
 )
 
 type gcpAuthJSON struct {
@@ -270,6 +271,31 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 	podSpec.SecurityContext.RunAsGroup = &rootUIDGid
 	podSpec.ServiceAccountName = "noobaa-endpoint"
 
+	honor := corev1.NodeInclusionPolicyHonor
+	disableDefaultTopologyConstraints, found := r.NooBaa.ObjectMeta.Annotations[nbv1.SkipTopologyConstraints]
+	if podSpec.TopologySpreadConstraints != nil {
+		r.Logger.Debugf("deployment %s TopologySpreadConstraints already exists, leaving as is", r.DeploymentEndpoint.Name)
+	} else if !r.hasNodeInclusionPolicyInPodTopologySpread() {
+		r.Logger.Debugf("deployment %s TopologySpreadConstraints cannot be set because feature gate NodeInclusionPolicyInPodTopologySpread is not supported on this cluster version",
+			r.DeploymentEndpoint.Name)
+	} else if found && disableDefaultTopologyConstraints == "true" {
+		r.Logger.Debugf("deployment %s TopologySpreadConstraints will not be set because annotation %s was set on noobaa CR",
+			r.DeploymentEndpoint.Name, nbv1.SkipTopologyConstraints)
+	} else {
+		r.Logger.Debugf("default TopologySpreadConstraints is added to %s deployment", r.DeploymentEndpoint.Name)
+		topologySpreadConstraint := corev1.TopologySpreadConstraint{
+			MaxSkew:           1,
+			TopologyKey:       "kubernetes.io/hostname",
+			WhenUnsatisfiable: corev1.ScheduleAnyway,
+			NodeTaintsPolicy:  &honor,
+			LabelSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"noobaa-s3": r.Request.Name,
+				},
+			},
+		}
+		podSpec.TopologySpreadConstraints = []corev1.TopologySpreadConstraint{topologySpreadConstraint}
+	}
 	for i := range podSpec.Containers {
 		c := &podSpec.Containers[i]
 		switch c.Name {
@@ -388,6 +414,23 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 		}
 	}
 	return nil
+}
+
+func (r *Reconciler) hasNodeInclusionPolicyInPodTopologySpread() bool {
+	kubeVersion, err := util.GetKubeVersion()
+	if err != nil {
+		r.Logger.Printf("❌ Failed to get kube version %s", err)
+		return false
+	}
+	enabledKubeVersion, err := semver.NewVersion(topologyConstraintsEnabledKubeVersion)
+	if err != nil {
+		util.Panic(err)
+		return false
+	}
+	if kubeVersion.LessThan(*enabledKubeVersion) {
+		return false
+	}
+	return true
 }
 
 func (r *Reconciler) setDesiredRootMasterKeyMounts(podSpec *corev1.PodSpec, container *corev1.Container) {
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -26,6 +26,7 @@ import (
 	"time"
 	"unicode"
 
+	semver "github.com/coreos/go-semver/semver"
 	kedav1alpha1 "github.com/kedacore/keda/v2/apis/keda/v1alpha1"
 	obv1 "github.com/kube-object-storage/lib-bucket-provisioner/pkg/apis/objectbucket.io/v1alpha1"
 	nbapis "github.com/noobaa/noobaa-operator/v5/pkg/apis"
@@ -53,6 +54,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/apimachinery/pkg/version"
 	"k8s.io/cli-runtime/pkg/printers"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
@@ -183,6 +185,22 @@ func KubeConfig() *rest.Config {
 		}
 	}
 	return lazyConfig
+}
+
+// GetKubeVersion will fetch the kubernates minor version
+func GetKubeVersion() (*semver.Version, error) {
+	var err error
+	var discClient *discovery.DiscoveryClient
+	var kubeVersionInfo *version.Info
+	if discClient, err = discovery.NewDiscoveryClientForConfig(KubeConfig()); err != nil {
+		return nil, err
+	}
+	if kubeVersionInfo, err = discClient.ServerVersion(); err != nil {
+		return nil, err
+	}
+	kubeVersion := fmt.Sprintf("%s.%s.0", kubeVersionInfo.Major, kubeVersionInfo.Minor)
+	version, err := semver.NewVersion(kubeVersion)
+	return version, err
 }
 
 // MapperProvider creates RESTMapper
--- a/pkg/apis/noobaa/v1alpha1/noobaa_types.go
+++ b/pkg/apis/noobaa/v1alpha1/noobaa_types.go
@@ -478,6 +478,9 @@ const (
 
 	// DeleteOBCConfirmation represents the validation to destry obc
 	DeleteOBCConfirmation CleanupConfirmationProperty = "yes-really-destroy-obc"
+
+	// SkipTopologyConstraints is Annotation name for disabling default topology Constraints
+	SkipTopologyConstraints = "noobaa.io/skip_topology_spread_constraints"
 )
 
 // DBTypes is a string enum type for specify the types of DB that are supported.


deploy/internal/deployment-endpoint.yaml
```diff
@@ -23,6 +23,16 @@ spec:
         noobaa.io/configmap-hash: ""
     spec:
       serviceAccountName: noobaa-endpoint
+      topologySpreadConstraints:
+      - maxSkew: 1
+        topologyKey: kubernetes.io/hostname
+        whenUnsatisfiable: ScheduleAnyway
+        # nodeTaintsPolicy available from kubernetes 1.26 or latest, 
+        # without this master node also will use for calculating pod topology spread skew
+        nodeTaintsPolicy: Honor
```
liranmauda, line 32: Could not find `nodeTaintsPolicy` in [openshift documentation](https://docs.openshift.com/container-platform/4.13/nodes/scheduling/nodes-scheduler-pod-topology-spread-constraints.html) , only in the [kubernetes documentation](https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/) 

Does Openshift support this? 
if yes, in what version, if not what will happen?

dannyzaken: this is supported since [kuberentes 1.19](https://kubernetes.io/blog/2023/04/17/fine-grained-pod-topology-spread-features-beta/)

deploy/internal/deployment-endpoint.yaml
```diff
@@ -23,6 +23,16 @@ spec:
         noobaa.io/configmap-hash: ""
     spec:
       serviceAccountName: noobaa-endpoint
+      topologySpreadConstraints:
+      - maxSkew: 1
+        topologyKey: kubernetes.io/hostname
+        whenUnsatisfiable: ScheduleAnyway
+        # nodeTaintsPolicy available from kubernetes 1.26 or latest,
```
dannyzaken, line 30: That might be a problem in ODF. 
I see that in 1.26, it is available as beta in Kubernetes, but I don't know the availability of k8s beta features in openshift. @naveenpaul1, any idea?

In any case, as long as this is supported by OCP 4.14, we're good. Otherwise, we can't use it since we should honor the taints policy.
@nimrod-becker WDYT?

naveenpaul1: @dannyzaken I did my testing in openshift 4.13.0 and nodeTaintsPolicy was considered for topology-spread-constraints, need to get a confirmation. 
@liranmauda without nodeTaintsPolicy master node also consider for skew calculation, since our skew value is maxSkew is 1 it might give unexpected behavior. It will not affect the pod deployment because of `whenUnsatisfiable: ScheduleAnyway`

liranmauda: Seems like openshift 4.13 is based on Kubernetes 1.12 [see this link](https://access.redhat.com/solutions/4870701)
So I guess we are good.

naveenpaul1: I could find the `nodeTaintsPolicy` in the Deployment creation page in openshift 4.13, So `nodeTaintsPolicy`  should be available.
image

naveenpaul1: Did testing with or without `nodeTaintsPolicy : Honor` in openshift 4.13(2 master nodes, 4 worker nodes), and could find the following behavior
- without `nodeTaintsPolicy : Honor`, the master node is also considered for `Skew` calculation, and if `whenUnsatisfiable: DoNotSchedule`(just for testing purposes) we can deploy only 4 pods in 4 worker nodes. This is because if we try to deploy one more pod it will violate topologySpreadConstraints that include the master node.
- with `nodeTaintsPolicy : Honor`, the master node is not considered for `Skew` calculation, and if `whenUnsatisfiable: DoNotSchedule`(just for testing purposes) we can deploy as any pod in 4 worker nodes and the only constrain is CPU/Memory.
with this behavior, I am certain that `nodeTaintsPolicy : Honor` is valid for openshift 4.13 and the latest releases.

guymguym: @naveenpaul1 Just wanted to say that I also stumbled this before, but I had a lower openshift version so couldn't use the `nodeTaintsPolicy: Honor` option, so my solution was to label the worker nodes with a label `noobaa-endpoint-domain: <node-name>` (uniq values) and used that label for the topologyKey. 

I also saw on 4.10 and 4.12 that the deployment yaml doesn't even store this new option. So what will happen for deployments that upgrade from 4.12 or lower? I will never be set... Perhaps we should reconcile this to the deployment if missing (but not override it if exists). WDYT @dannyzaken

naveenpaul1: @guymguym Yes it makes sense for clusters migrating from 4.12 and older versions of openshift because there wont be `nodeTaintsPolicy` in the endpoint deployment yaml file when these clusters migrate.

guymguym: @naveenpaul1 But our operator can be installed on clusters with lower version. We are not syncing the version of the noobaa operator with the cluster version. So if we deploy the new operator on a 4.12 version, that setting will get lost.

guymguym: The thing is that upstream does not have the same notion of OCP -1 for k8s.
We could reconcile it, but then I worry that the operator will keep creating the topologySpreadConstraints in case a user is interested in removing it or modifying it...
@dannyzaken what would you do?

dannyzaken: We should provide a flag to disable it under endpointSpec. I usually prefer to avoid adding configurations in our CR, but I don't see a way around it for older k8s versions. @guymguym @naveenpaul1 WDYT?

dannyzaken (commented):
@naveenpaul1 I think that this change is only good for new deployments but will not take effect on upgraded deployments. we should handle it also in [phase 4 `SetDesiredDeploymentEndpoint` ](https://github.com/noobaa/noobaa-operator/blob/ab64999012486f1cfefc6eb3f0e471104234e131/pkg/system/phase4_configuring.go#L252C1-L252C1)

deploy/internal/deployment-endpoint.yaml
naveenpaul1: I think if we are planning to give users an option to edit/delete topology constraints we do not have another option, especially when we add the code in noobaa endpoint reconcile loop.

dannyzaken: @naveenpaul1, following our discussion, let's add an env variable to disable this feature. It's better than adding a flag to the CRD

dannyzaken (approved):

pkg/system/phase4_configuring.go
```diff
@@ -269,7 +271,24 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
...
+		r.Logger.Info("TopologyConstraints for noobaa endpoint is disabled")
+	} else {
+		topologySpreadConstraint := corev1.TopologySpreadConstraint{
+			MaxSkew:           1,
+			TopologyKey:       "kubernetes.io/hostname",
+			WhenUnsatisfiable: corev1.ScheduleAnyway,
+			NodeTaintsPolicy:  &honor,
+			LabelSelector: &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"noobaa-s3": "noobaa",
```
guymguym, line 286:
```suggestion
					"noobaa-s3": r.Request.Name,
```

pkg/system/phase4_configuring.go
```diff
@@ -269,7 +271,24 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
...
+				},
+			},
+		}
+		podSpec.TopologySpreadConstraints = []corev1.TopologySpreadConstraint{topologySpreadConstraint}
```
guymguym, line 290: why are we always overriding it? shouldn't we only set it if null?

deploy/operator.yaml
```diff
@@ -45,6 +45,7 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
+            - name: DISABLE_TOPOLOGY_SPREAD_CONSTRAINTS
```
guymguym, line 48: I don't know about this. There are no other configurations that we use to set on the operator as env. Nobody will expect to find this option here.

guymguym (changes requested):

pkg/system/phase4_configuring.go
naveenpaul1: @guymguym we will override it only when `DISABLE_TOPOLOGY_SPREAD_CONSTRAINTS`(will rename to `DISABLE_DEFAULT_TOPOLOGY_SPREAD_CONSTRAINTS`) is empty or false, If the user set it to true and adds their `TopologySpreadConstraint` it will not be overridden. Correct me if am wrong.

deploy/operator.yaml
naveenpaul1: @guymguym another option would be adding a ConfigMap for env, but now noobaa operator does not have such a configuration we need to create one for that.

pkg/system/phase4_configuring.go
guymguym: why override it at all? if it's set I would just leave it as is.

deploy/operator.yaml
guymguym: I would prefer a label on the noobaa CR than env on the operator deployment.

pkg/system/phase4_configuring.go
```diff
@@ -269,7 +269,22 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 	podSpec.SecurityContext.RunAsUser = &rootUIDGid
 	podSpec.SecurityContext.RunAsGroup = &rootUIDGid
 	podSpec.ServiceAccountName = "noobaa-endpoint"
-
+	honor := corev1.NodeInclusionPolicyHonor
+	disableDefaultTopologyConstraints, found := r.NooBaa.ObjectMeta.Annotations[util.DisableDefaultTopologyConstraints]
+	if (found && disableDefaultTopologyConstraints == "false") && podSpec.TopologySpreadConstraints == nil {
```
guymguym, line 274: I think we better also check here if the specific feature-gate for NodeInclusionPolicyInPodTopologySpread is enabled on this cluster. 

Also I think the conditions can be written with a bit more information:
```suggestion
	honor := corev1.NodeInclusionPolicyHonor
	hasNodeInclusionPolicyInPodTopologySpread := // ... how to get feature gate ...
	disableDefaultTopologyConstraints, found := r.NooBaa.ObjectMeta.Annotations[util.DisableDefaultTopologyConstraints]
	if podSpec.TopologySpreadConstraints != nil {
		log.Infof("deployment %s TopologySpreadConstraints already exists, leaving as is", r.DeploymentEndpoint.Name)
	} else if hasNodeInclusionPolicyInPodTopologySpread {
		log.Infof("deployment %s TopologySpreadConstraints cannot be set because feature gate NodeInclusionPolicyInPodTopologySpread is not supported on this cluster version", r.DeploymentEndpoint.Name)
	} else if found && disableDefaultTopologyConstraints == "true" {
		log.Infof("deployment %s TopologySpreadConstraints will not be set because annotation %s was set on noobaa CR", r.DeploymentEndpoint.Name, util.DisableDefaultTopologyConstraints)
	} else {
		// ... set the spec
	}
```

pkg/util/util.go
```diff
@@ -75,6 +75,8 @@ const (
 	gigabyte             = 1024 * 1024 * 1024
 	petabyte             = gigabyte * 1024 * 1024
 	obcMaxSizeUpperLimit = petabyte * 1023
+	// DisableDefaultTopologyConstraints is Annotation name for disabling default topology Constraints
+	DisableDefaultTopologyConstraints = "disableDefaultTopologyConstraints"
```
guymguym, line 79: Why here? It should be in the noobaa_types.go as it is part of our yaml api.
Also for best practices of annotations I believe we want to set it with a prefix to something like `noobaa.io/skip_topology_spread_constraints`

I prefer to say skip vs disable so that it's clear that we are not disabling the feature, but just not operating it from our operator.

guymguym (changes requested):

pkg/util/util.go
```diff
@@ -1512,6 +1513,26 @@ func IsStringArrayUnorderedEqual(stringsArrayA, stringsArrayB []string) bool {
...
 }
 
+// EnsureNoobaaAnnotations ensures that the resource has all mandatory meta fields
+func EnsureNoobaaAnnotations(object metav1.Object) bool {
+	updated := false
+
+	annotations := object.GetAnnotations()
+	if annotations == nil {
+		object.SetAnnotations(map[string]string{
+			DisableDefaultTopologyConstraints: "false",
```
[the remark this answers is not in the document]
naveenpaul1: @guymguym I didn't get what you meant by `leave this out.`, I added this and set it to default `true` only because the user should get a chance to know this configuration we have in Noobaa as you suggested.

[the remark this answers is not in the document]
guymguym: no need to set it on the CR though, it should be documented instead.

pkg/system/phase4_configuring.go
naveenpaul1: @guymguym I checked the 4.14 cluster and  NodeInclusionPolicyInPodTopologySpread is not explicitly mentioned in FeatureGate CR.  Openshift team mentioned that Instead of checking the `NodeInclusionPolicyInPodTopologySpread` we can check the OCP version and every OCP/Kubernetes version until 4.12 has the feature disabled. 4.13+ have it enabled. WDYT?

guymguym: sounds ok to me

pkg/system/phase4_configuring.go
```diff
@@ -390,6 +415,20 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 	return nil
 }
 
+func (r *Reconciler) hasNodeInclusionPolicyInPodTopologySpread() bool {
```
guymguym, line 418: This is reversed - it returns true if it doesn't support it...

pkg/system/phase4_configuring.go
```diff
@@ -390,6 +415,20 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
...
+	kubeMinorVersion, err := util.GetKubeVersion()
+	if err != nil {
+		r.Logger.Infof("❌ Failed to get kube version %s", err)
+		return true
```
guymguym, line 422: Return true - why? Also this logs twice.

pkg/util/util.go
```diff
@@ -185,6 +186,20 @@ func KubeConfig() *rest.Config {
 	return lazyConfig
 }
 
+// GetKubeVersion will fetch the kubernates minor version
+func GetKubeVersion() (string, error) {
```
guymguym, line 190: Why only minor? And if so, why not name it GetKubeMinorVersion...

doc/noobaa-crd.md
```diff
@@ -240,6 +240,22 @@ spec:
   dbConf: |+
     max_connections = 1000
 ```
+## Pod Topology Spread Constraints for Noobaa endpoint
+
+Pod Topology Spread Constraints is enabled by default in Noobaa.
```
guymguym, line 245: This short sentence is not clear enough to me. I would explain that the operator will set this on the noobaa endpoint deployment in order to force spreading endpoint pods evenly between nodes, and that this will not work before kubernetes 1.26 so the operator checks for it, and also skips if this is already set on the deployment, or the CR annotation is set. I would also explain that users can make changes to this after the operator creates it and it will not override it.

pkg/system/phase4_configuring.go
```diff
@@ -390,6 +415,19 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 	return nil
 }
 
+func (r *Reconciler) hasNodeInclusionPolicyInPodTopologySpread() bool {
+	kubeVersion, err := util.GetKubeVersion()
+	if err != nil {
+		return true
+	}
+	for _, topologyConstraintsSkipKubeVersion := range topologyConstraintsSkipKubeVersions {
```
dannyzaken, line 423: you can use `semver` package to compare the versions. see usage here:
https://github.com/naveenpaul1/noobaa-operator/blob/ca4a2513254fea3f8093fa10a63d0939b37caaf8/pkg/system/phase1_verifying.go#L87-L100

pkg/system/phase4_configuring.go
```diff
@@ -270,6 +271,30 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 	podSpec.SecurityContext.RunAsGroup = &rootUIDGid
 	podSpec.ServiceAccountName = "noobaa-endpoint"
 
+	honor := corev1.NodeInclusionPolicyHonor
+	disableDefaultTopologyConstraints, found := r.NooBaa.ObjectMeta.Annotations[nbv1.SkipTopologyConstraints]
+	if podSpec.TopologySpreadConstraints != nil {
+		r.Logger.Infof("deployment %s TopologySpreadConstraints already exists, leaving as is", r.DeploymentEndpoint.Name)
```
dannyzaken, line 277: this might be too verbose, since it is expected to be the case for most reconciles.

doc/noobaa-crd.md
```diff
@@ -240,6 +240,22 @@ spec:
...
+  - maxSkew: 1
+    topologyKey: kubernetes.io/hostname
+    whenUnsatisfiable: ScheduleAnyway
+    labelSelector:
+      matchLabels:
+        noobaa-s3: noobaa
+    nodeTaintsPolicy: Honor
+
+```
+Users can make changes to `topologySpreadConstraints` configuration after the operator creates it and the changes will not override it.
```
dannyzaken, line 258: worth clarifying that the default value is resotored if `topologySpreadConstraints` is removed completely

pkg/system/phase4_configuring.go
naveenpaul1: Hi @dannyzaken, Guy sugested this changes  : https://github.com/noobaa/noobaa-operator/pull/1138#discussion_r1257818370

guymguym (approved):
Looks good! added couple of comments on implementation details ^ Thanks

PR #1142, opened 2023-06-28
update noobaa endpoint replication

naveenpaul1:
### Explain the changes
1.  update noobaa endpoint pod with respect to noobaa CR spec. 

1. https://bugzilla.redhat.com/show_bug.cgi?id=2217904

### Testing Instructions:

1. install Prometheus(only if Promethue missing) and noobaa in cluster 
eg: ``` helm install prometheus  prometheus-community/kube-prometheus-stack -n monitoring ```
nb install --operator-image='{image}' --noobaa-image='{image}' -n {namespace} --autoscaler-type hpav2 --prometheus-namespace {namespace}
2. update noobaa CR with autoscalerType and prometheusNamespace
``` autoscaler: ```
    ``` autoscalerType: hpav2 ```
     ``` prometheusNamespace: monitoring ```
    
3.update noobaa endpoint min and max count from noobaa CR and check the noobaa endpoint scale up/down.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/autoscaler.go
+++ b/pkg/system/autoscaler.go
@@ -109,7 +109,7 @@ func (r *Reconciler) ensureHPAV1Cleanup(log *logrus.Entry) error {
 
 func (r *Reconciler) ensureKedaCleanup(log *logrus.Entry) error {
 	noobaaHpaSelector, _ := labels.Parse("app=noobaa")
-	noobaaKedaFieldSelector, _ := fields.ParseSelector("metadata.name=keda-hpa-noobaa-endpoint")
+	noobaaKedaFieldSelector, _ := fields.ParseSelector("metadata.name=keda-hpa-noobaa")
 	// List autoscalers based on the label and name for keda ,
 	// Added name beause autoscalingv2 KubeList will fetch autoscaler with version v1 and v2
 	autoscalersv2 := &autoscalingv2.HorizontalPodAutoscalerList{}
@@ -139,7 +139,7 @@ func (r *Reconciler) ensureKedaCleanup(log *logrus.Entry) error {
 func (r *Reconciler) ensureHPAV2Cleanup(log *logrus.Entry) error {
 
 	noobaaHpav2Selector, _ := labels.Parse("app=noobaa")
-	noobaaHPAV2FieldSelector, _ := fields.ParseSelector("metadata.name=hpav2-noobaa-endpoint")
+	noobaaHPAV2FieldSelector, _ := fields.ParseSelector("metadata.name=noobaa-hpav2")
 	// List autoscalers based on the label and name for HPAV2 ,
 	// Added name beause autoscalingv2 KubeList will fetch autoscaler with version v1 and v2
 	autoscalersv2 := &autoscalingv2.HorizontalPodAutoscalerList{}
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: pkg/system/reconciler.go
Comment (shirady) [inline pkg/system/reconciler.go:1]: How those changes in the `AdapterHPA` are related to the bug fix (endpoint minCount and maxCount are ignored)?

File: pkg/system/autoscaler.go [hunk 1]
```diff
@@ -273,20 +274,23 @@ func (r *Reconciler) autoscaleKeda(prometheus *monitoringv1.Prometheus) error {
...
-		r.KedaScaled.Spec.MinReplicaCount = &minReplicas
-		r.KedaScaled.Spec.MaxReplicaCount = &maxReplicas
-	}
-	if !util.KubeCreateSkipExisting(r.KedaScaled) {
-		log.Errorf("❌ Failed to create KedaScaledObject")
-		return fmt.Errorf("Failed to create KedaScaledObject")
+		minReplicas = endpointsSpec.MinCount
+		maxReplicas = endpointsSpec.MaxCount
 	}
+	fmt.Println("r.KedaScaled.Spec.MinReplicaCount ", minReplicas, maxReplicas)
```
shirady, line 291: Is this printing supposed to be for the user or internal log printing?

pkg/system/reconciler.go
naveenpaul1: @shirady These new changes will make sure the operator will fetch the new changed value from noobaa CR, before it was skipping the update if HPA has already been created.

pkg/system/autoscaler.go
naveenpaul1: yes, will remove it

pkg/system/autoscaler.go
```diff
@@ -493,22 +496,23 @@ func (r *Reconciler) reconcilePrometheusAdapterResources(prometheus *monitoringv
...
-	adapterHPA.Namespace = r.Request.Namespace
-	if err := r.ReconcileObject(adapterHPA, nil); err != nil {
-		return err
+		minReplicas = endpointsSpec.MinCount
+		maxReplicas = endpointsSpec.MaxCount
 	}
-
+	r.AdapterHPA.Spec.MinReplicas = &minReplicas
+	r.AdapterHPA.Spec.MaxReplicas = maxReplicas
+	r.AdapterHPA.Spec.Metrics[0].Object.Target.Value = resource.NewQuantity(1, resource.DecimalSI)
```
shirady, line 515: Would you please add a comment about this line?

naveenpaul1: @shirady I was getting an error when trying to update the HPA without `target.value`, So added this and it will not have any specific use since  `type`  is `AverageValue`.

dannyzaken (approved):

pkg/system/autoscaler.go
shirady: It can be easier to understand if you add a comment about the purpose of this line.

jackyalbo: I agree, let's explain this a bit... 
What do we expect as Metrics in place 0? shouldn't we build the whole spec?
In the file it hpav2-autoscaling.yaml it looks like we will create averageValue and not Value, I think they should look the same here and there

pkg/system/reconciler.go
```diff
@@ -284,6 +287,7 @@ func NewReconciler(
 
 	r.KedaTriggerAuthentication.Namespace = r.Request.Namespace
 	r.KedaScaled.Namespace = r.Request.Namespace
+	r.AdapterHPA.Namespace = r.Request.Namespace
```
jackyalbo, line 290: move namespace set to the namespace area for all of this resources - line 220
also add names to the names set area line 263, following our format: r.Request.Name + <type>
```
	r.AdapterHPA.Name = r.Request.Name + "-adapter-hpa"
	r.KedaScaled.Name = r.Request.Name + "-keda-scaled"
```

jackyalbo (changes requested):

pkg/system/autoscaler.go
naveenpaul1: Hi @shirady and @jackyalbo I managed to dive deep into this issue and find out that instead of setting a value I can set `nil` to this, the code is updated 
https://github.com/kubernetes/kubernetes/blob/afff019fbc90373127020693cdb056a968468f04/pkg/apis/autoscaling/validation/validation.go#L445C5-L445C7

pkg/system/reconciler.go
naveenpaul1: @jackyalbo code updated,

pkg/system/reconciler.go
```diff
@@ -258,6 +264,8 @@ func NewReconciler(
 	r.DeploymentEndpoint.Name = r.Request.Name + "-endpoint"
 	r.UpgradeJob.Name = r.Request.Name + "-upgrade-job"
 	r.CaBundleConf.Name = r.Request.Name + "-ca-inject"
+	r.KedaScaled.Name = r.Request.Name
```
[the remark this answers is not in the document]
naveenpaul1: @shirady keda automatically adds a prefix for the hpa `keda-hpa` so it will make the name long and repeating words

pkg/system/autoscaler.go
```diff
@@ -493,22 +495,23 @@ func (r *Reconciler) reconcilePrometheusAdapterResources(prometheus *monitoringv
...
-	adapterHPA.Namespace = r.Request.Namespace
-	if err := r.ReconcileObject(adapterHPA, nil); err != nil {
-		return err
+		minReplicas = endpointsSpec.MinCount
+		maxReplicas = endpointsSpec.MaxCount
 	}
-
+	r.AdapterHPA.Spec.MinReplicas = &minReplicas
+	r.AdapterHPA.Spec.MaxReplicas = maxReplicas
+	r.AdapterHPA.Spec.Metrics[0].Object.Target.Value = nil
```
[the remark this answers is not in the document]
naveenpaul1: @shirady without this change operator will simply create the HPA only once and do not update
`target.value` should set to nil otherwise HPA reconciler (in Kubernetes code) will try to validate the code

jackyalbo (approved):

pkg/system/autoscaler.go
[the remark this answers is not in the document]
naveenpaul1: comment added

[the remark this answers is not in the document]
shirady: Thanks

PR #1147, opened 2023-07-03
COSI | CLI | bucketclass & bucketclaim support

romayalon:
### Explain the changes
1. Added cosi bucket claim and bucket class commands to noobaa cli.
2. Supported create, list, status and delete for both COSI bucketclaim and bucketclass
3. Refactored bucketclass cli in order to use the same populate functions for cosi bucketclass.
4. Added COSI types 
5. Added COSI bucket crd yaml to be used on COSI bucket claim creation.
6. Updated docs
7. Added cli integration tests

### Testing Instructions:
1. make test-cosi

- [x] Doc added/updated
- [x] Tests added

### DIFF
--- a/pkg/bucketclass/bucketclass.go
+++ b/pkg/bucketclass/bucketclass.go
@@ -199,102 +199,26 @@ func CmdReconcile() *cobra.Command {
 
 // RunCreateSingleNamespaceBucketClass runs a CLI command
 func RunCreateSingleNamespaceBucketClass(cmd *cobra.Command, args []string) {
-	log := util.Logger()
-	createCommonBucketclass(cmd, args, nbv1.NSBucketClassTypeSingle, func(bucketClass *nbv1.BucketClass) ([]string, []string) {
-		resource, _ := cmd.Flags().GetString("resource")
-		if resource == "" {
-			log.Fatalf(`❌ Must provide one namespace store`)
-		}
-		bucketClass.Spec.NamespacePolicy.Single = &nbv1.SingleNamespacePolicy{
-			Resource: resource,
-		}
-		var namespaceStoresArr []string
-		return append(namespaceStoresArr, resource), []string{}
-	})
+	createCommonBucketclass(cmd, args, nbv1.NSBucketClassTypeSingle, PopulateSingleNamespaceBucketClass)
 }
 
 // RunCreateMultiNamespaceBucketClass runs a CLI command
 func RunCreateMultiNamespaceBucketClass(cmd *cobra.Command, args []string) {
-	log := util.Logger()
-	createCommonBucketclass(cmd, args, nbv1.NSBucketClassTypeMulti, func(bucketClass *nbv1.BucketClass) ([]string, []string) {
-		writeResource, _ := cmd.Flags().GetString("write-resource")
-		readResources, _ := cmd.Flags().GetStringSlice("read-resources")
-		if len(readResources) == 0 {
-			log.Fatalf(`❌ Must provide at least one read resource`)
-		}
-		bucketClass.Spec.NamespacePolicy.Multi = &nbv1.MultiNamespacePolicy{
-			WriteResource: writeResource,
-			ReadResources: readResources,
-		}
-		if writeResource == "" {
-			return readResources, []string{}
-		}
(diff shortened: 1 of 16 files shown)


pkg/cosi/cosi_bucket_claim.go
```diff
@@ -0,0 +1,346 @@
...
+}
+
+// WaitReady waits until the cosi bucket claim status bucket ready changes to true
+func WaitReady(cosiBucketClaim *nbv1.COSIBucketClaim) bool {
+	log := util.Logger()
+	klient := util.KubeClient()
+
+	intervalSec := time.Duration(3)
+
+	err := wait.PollImmediateInfinite(intervalSec*time.Second, func() (bool, error) {
```
jackyalbo, line 292: let's try to limit the time since we don't have a failed status like in obc.
in future PR please add a recording of error events so the user will know the failure reason.

pkg/cosi/cosi_bucket_class.go
```diff
@@ -0,0 +1,435 @@
...
+
+	// placement policy flags
+	cmd.Flags().String("placement", "",
+		"Set first tier placement policy - Mirror | Spread | \"\" (empty defaults to single backing store)")
+	cmd.Flags().StringSlice("backingstores", nil,
+		"Set first tier backing stores (use commas or multiple flags)")
+	cmd.Flags().String("replication-policy", "",
+		"Set the json file name that contains the replication rules")
+	cmd.Flags().String("deletion-policy", "delete",
+		"Set the deletion policy of the buckets delete | retain - the default deletion policy is delete")
```
jackyalbo, line 65: maybe rephrase this so user will better understand the meaning of the options and the default.

pkg/cosi/cosi_bucket_class.go
```diff
@@ -0,0 +1,435 @@
...
+	deletionPolicy, _ := cmd.Flags().GetString("deletion-policy")
+	if deletionPolicy == "delete" {
+		bucketClass.DeletionPolicy = nbv1.COSIDeletionPolicyDelete
+	} else if deletionPolicy == "retain" {
+		bucketClass.DeletionPolicy = nbv1.COSIDeletionPolicyRetain
+	} else {
+		log.Fatalf(`❌ Invalid deletion policy, valid values are delete or retain %s`, deletionPolicy)
+	}
+
+	bucketClassSpec := &nbv1.BucketClassSpec{}
```
jackyalbo, line 238: move next to populate function

pkg/cosi/cosi_bucket_class.go
```diff
@@ -0,0 +1,435 @@
...
+		pp := bc.Parameters["placementPolicy"]
+		np := bc.Parameters["namespacePolicy"]
+		quota := bc.Parameters["quota"]
+		rp := bc.Parameters["replicationPolicy"]
+		table.AddRow(
+			bc.Name,
+			pp,
+			np,
+			quota,
+			rp,
```
jackyalbo, line 430: also here limit the size for replication policy

pkg/bucketclass/bucketclass.go
```diff
@@ -522,11 +535,13 @@ func RunList(cmd *cobra.Command, args []string) {
 		pp, _ := json.Marshal(bc.Spec.PlacementPolicy)
 		np, _ := json.Marshal(bc.Spec.NamespacePolicy)
 		quota, _ := json.Marshal(bc.Spec.Quota)
+		rp, _ := json.Marshal(bc.Spec.ReplicationPolicy)
```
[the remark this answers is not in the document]
romayalon: I decided to remove it from this PR as it doesn't look good even with a limited number of chars, we need to think about how to show it in a better way and it's not related to this PR so we can move it out

pkg/cosi/cosi_bucket_claim.go
romayalon: added retries

pkg/cosi/cosi_bucket_class.go
romayalon: I also changed the default to retain as COSI's default

jackyalbo (approved):

PR #1153, opened 2023-07-04
Support Azure log-based replication optimization credentials

Neon-White:
### Explain the changes
1. Support the new secret fields that are required for https://github.com/noobaa/noobaa-core/pull/7308

The changes were also inserted into the backingstores' `reconciler.go` for cases in which the same secret is used for Azure backingstores and namespacestores


dannyzaken (commented):
we will probably want to add tests for the additional secret fields. in that case we will want to merge the core PR first and reference it in options.go

dannyzaken (commented):
looks good to me. once the [core PR](https://github.com/noobaa/noobaa-core/pull/7308) is merged we can change the core image ref in `options.go`

dannyzaken (approved):

PR #1183, opened 2023-07-31
Remove operator-sdk dependency

v-harihar:
list of changes:
Replaced operator-sdk command to generate
1. crds with controller-gen cmd
2. runtime object with deepcopy-gen
3. noobaa-operator and noobaa-operator-local directly with go build cmds
4. noobaa images with docker cmd


liranmauda:
@v-harihar  Do we need go 1.20 for this change? 
if changing to go 1.20 is not required, then we should do it in a different PR (that can be a draft till this one is merged).

v-harihar:
> @v-harihar Do we need go 1.20 for this change? if changing to go 1.20 is not required, then we should do it in a different PR (that can be a draft till this one is merged).

Yes Liran. The basic idea of the PR is to go for 1.20. After upgrading to 1.20, we faced issue related to `operator-sdk` (need to change directory structure and other things forcefully). Then we arrived at the solution to go independent of `operator-sdk`. I understand the PR description is bit misleading but take this comment as correction to it :)

liranmauda:
@v-harihar I don't think this change needs go 1.20
Please divide it. 
Yes, eventually this is the goal, but it should not be in the same PR.
Unless making the changes required to remove our dependency on operator-sdk requires go 1.20 specifically, we should not do it in the same PR. 
Doing so in the same PR will make it harder to backport if we want to backport only the removal of our dependency on operator-sdk.

build/hack/boilerplate.go.txt
dannyzaken, line 1: why do we need it?

v-harihar: `Deepcopy-gen` tool asks for the boiler-plate text. I remember it was not working without that and I added it.

deploy/crds/noobaa.io_backingstores.yaml
```diff
@@ -64,6 +67,7 @@ spec:
                           secret name must be unique.
                         type: string
                     type: object
+                    x-kubernetes-map-type: atomic
```
dannyzaken, line 70: Make sure this should not affect our current behavior.

liranmauda (commented):
LGTM

dannyzaken (approved):

deploy/crds/noobaa.io_backingstores.yaml
v-harihar: As per below description and default generation, going with atomic type

```
//+mapType | x-kubernetes-map-type | atomic/granular | Applicable to maps. atomic means that the map can only be entirely replaced by a single manager. granular means that the map supports separate managers updating individual fields.
-- | -- | -- | --
```

PR #1192, opened 2023-08-08
Correcting misaligned .PHONY in Makefile

v-harihar:
Correcting the syntax error introduced in https://github.com/noobaa/noobaa-operator/pull/1183

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ vendor:
 run: gen
 	go build -o $(NOOBAA_OPERATOR_LOCAL) -mod=vendor $(CMD_MANAGER)
 	$(BIN)/noobaa-operator-local operator run
-	.PHONY: run
+.PHONY: run
 
 clean:
 	rm -rf $(OUTPUT)


Makefile
```diff
@@ -86,7 +86,7 @@ vendor:
 run: gen
 	go build -o $(NOOBAA_OPERATOR_LOCAL) -mod=vendor $(CMD_MANAGER)
 	$(BIN)/noobaa-operator-local operator run
-	.PHONY: run
+.PHONY: run
```
liranmauda, line 89:
```suggestion
.PHONY: gen
```

Makefile
```diff
@@ -86,7 +86,7 @@ vendor:
 run: gen
 	go build -o $(NOOBAA_OPERATOR_LOCAL) -mod=vendor $(CMD_MANAGER)
 	$(BIN)/noobaa-operator-local operator run
-	.PHONY: run
+.PHONY: gen
```
dannyzaken, line 89: why is it `.PHONY: gen` and not `.PHONY: run`?

v-harihar: I misread the job . I correct it

liranmauda: @v-harihar
Sorry, you are correct, my mistake 😄
I thought that the target is gen... and the target is run...
it should actually be as you wrote, `.PHONY: run`

liranmauda (approved):

PR #1195, opened 2023-08-10
Add support for config file

tangledbytes:
### Explain the changes
This PR adds support for config file. The following things are supported right now:
1. All the persistent/global flags can be specified in the config file instead of passing as a flag.
2. `resources` field is supported in the config file which allows overriding the default resources allocated to NooBaa components. This field can be used in conjunction with pre-existing flags like `--dev` and `--mini` in which case the values from the config file will override the values of the flag.

The config file can be placed at `/etc/noobaa/noobaa.cfg.yaml`, `~/.noobaa/noobaa.cfg.yaml` or at `./noobaa.cfg.yaml`.

Example config file
```yaml
# This will override the namespace of installation
namespace: noobaa-test

# This will override the operator image
operator-image: noobaa/noobaa-operator:5.15.0

# This will override the show-secrets which defaults to false
show-secrets: true

# This is equivalent to using `--dev`
dev: true

# This is a special field which is unavailable in the CLI via flags
resources:
  core:
    memoryMB: 512
  db:
    cpuMilli: 500
    memoryMB: 1024
  endpoints:
    cpuMilli: 500
    memoryMB: 1024
    minCount: 1
    maxCount: 3 
```

- [ ] Doc added/updated
- [ ] Tests added


tangledbytes:
Moving to draft, tests are failing...

liranmauda (approved):
LGTM, just one comment/question

pkg/cli/cli.go
```diff
@@ -154,3 +158,41 @@ Load noobaa completion to bash:
...
+	viper.SetConfigName("noobaa.cfg")
+	viper.SetConfigType("yaml")
+	viper.AddConfigPath("/etc/noobaa/")
+	viper.AddConfigPath("$HOME/.noobaa")
+	viper.AddConfigPath(".")
+
+	if err := viper.ReadInConfig(); err != nil {
+		if _, ok := err.(viper.ConfigFileNotFoundError); !ok {
+			logrus.Warn("failed to read config:", err)
+		}
```
liranmauda, line 172: I couldn't figure out when this can happen.

tangledbytes: I think this can happen when say the config file is not valid (valid as in not adhering to yaml spec) `yaml`. In such cases we would want just log that there was some issue in reading the file.

PR #1198, opened 2023-08-16
Fix the operator image make target

Neon-White:
### Explain the changes
1. Restore [Operator SDK behavior](https://github.com/operator-framework/operator-sdk/blob/581265d8448c75bdbae619238c4aa0fdcc50765f/cmd/operator-sdk/build/cmd.go#L99-L104) when building the operator image to have `GOOS` set to `linux` and to be open to receive `CGO_ENABLED` from the user

1. Fixes the issue where the operator pod would fail to go up due to `exec format error` since the binary was compiled for the host machine's OS instead of `linux` when compiled on machines not running Linux

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ cli: gen
 .PHONY: cli
 
 image: $(docker) gen
-	go build -o $(BIN)/noobaa-operator -gcflags all=-trimpath="$(MK_PARENT)" -asmflags all=-trimpath="$(MK_PARENT)" -mod=vendor $(CMD_MANAGER)
+	GOOS=linux CGO_ENABLED=$(or ${CGO_ENABLED},0) go build -o $(BIN)/noobaa-operator -gcflags all=-trimpath="$(MK_PARENT)" -asmflags all=-trimpath="$(MK_PARENT)" -mod=vendor $(CMD_MANAGER)
 	docker build -f build/Dockerfile -t $(IMAGE) .
 	@echo "✅ image"
 .PHONY: image


tangledbytes:
For some reason on running `GOOS=linux CGO_ENABLED=0 make image`, I am getting:
```
unknown flag: --version
Usage of deepcopy-gen:
      --add_dir_header                   If true, adds the file directory to the header of the log messages
      --alsologtostderr                  log to standard error as well as files (no effect when -logtostderr=true)
      --bounding-dirs strings            Comma-separated list of import paths which bound the types for which deep-copies will be generated.
      --build-tag string                 A Go build tag to use to identify files generated by this command. Should be unique. (default "ignore_autogenerated")
  -h, --go-header-file string            File containing boilerplate header text. The string YEAR will be replaced with the current 4-digit year. (default "/Users/usrivast/go/src/k8s.io/gengo/boilerplate/boilerplate.go.txt")
  -i, --input-dirs strings               Comma-separated list of import paths to get input types from.
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory (no effect when -logtostderr=true)
      --log_file string                  If non-empty, use this log file (no effect when -logtostderr=true)
      --log_file_max_size uint           Defines the maximum size a log file can grow to (no effect when -logtostderr=true). Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
      --logtostderr                      log to standard error instead of files (default true)
      --one_output                       If true, only write logs to their native severity level (vs also writing to each lower severity level; no effect when -logtostderr=true)
  -o, --output-base string               Output base; defaults to $GOPATH/src/ or ./ if $GOPATH is not set. (default "/Users/usrivast/go/src")
  -O, --output-file-base string          Base name (without .go suffix) for output files. (default "deepcopy_generated")
  -p, --output-package string            Base package path.
      --skip_headers                     If true, avoid header prefixes in the log messages
      --skip_log_headers                 If true, avoid headers when opening log files (no effect when -logtostderr=true)
      --stderrthreshold severity         logs at or above this threshold go to stderr when writing to files and stderr (no effect when -logtostderr=true or -alsologtostderr=false) (default 2)
      --trim-path-prefix string          If set, trim the specified prefix from --output-package when generating files.
  -v, --v Level                          number for the log level verbosity
      --verify-only                      If true, only verify existing output, do not write anything.
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
go mod tidy
go mod vendor
✅ vendor
✅ gen
fork/exec /var/folders/q1/hdfs8drn5q3d3d4_6kgshs0w0000gn/T/go-build4238045178/b001/exe/main: exec format error
go build -o build/_output/bin/noobaa-operator -gcflags all=-trimpath="/Users/usrivast/dev/noobaa" -asmflags all=-trimpath="/Users/usrivast/dev/noobaa" -mod=vendor cmd/manager/main.go
docker build -f build/Dockerfile -t noobaa/noobaa-operator: .
[+] Building 0.0s (0/0)
ERROR: invalid tag "noobaa/noobaa-operator:": invalid reference format
make: *** [image] Error 1
```

Is this only me?

tangledbytes:
Whoops, my bad. GOOS and CGO needs to be set only on the `go build`.

PR #1202, opened 2023-08-21
Pass Region in Analyze Resource

shirady:
### Explain the changes
1. pass region in analyze resource.

1. none

### Testing Instructions:
Note: to actually run the test you'll need to add the changes from this noobaa/noobaa-core#7452.
1. Deploy noobaa on MInikube or Rancher Desktop (see [guide](https://github.com/noobaa/noobaa-operator/blob/master/doc/dev_guide/deply_noobaa_on_minikube_or_rancher_desktop.md)).
_Note: `nb` is an alias that runs the local operator from `build/_output/bin` (alias created by `devenv`)._
2. For backingstore, Run:
`nb backingstore create aws-s3 <backing-store-name>` (optional: add the `--region` flag).
`nb diagnostics analyze backingstore <backing-store-name>`
3. For namespacestore, Run:
`nb namespacestore create aws-s3 <namespace-store-name>`  (optional: add the `--region` flag).
`nb diagnostics analyze namespacestore <namespace-store-name>`
3. For all resources (backingstore and namespacestore), Run:
`nb diagnostics analyze resources`
* It creates the log files and then deletes the created job.
4. Cases:
1 - s3-compatible - Noobaa bucket (with noobaa system as a server, the endpoint was taken from the field `InternalDNS` under `S3 Addresses`).
2 - aws-s3 without flag region on a bucket in a region that is not 'us-east-1' (for example 'us-west-1') and aws-s3 with flag region.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/diagnostics/analyze.go
+++ b/pkg/diagnostics/analyze.go
@@ -270,6 +270,16 @@ func setBackingStoreDetailsInJob(backingStore *nbv1.BackingStore, cmd *cobra.Com
 			backingStore.Kind, backingStore.Name, backingStore.Namespace)
 	}
 
+	// in case it is aws endpoint, we would pass the region
+	if backingStore.Spec.Type == nbv1.StoreTypeAWSS3 {
+		region := util.GetEnvVariable(&analyzeResourceJob.Spec.Template.Spec.Containers[0].Env, "REGION")
+		if region == nil {
+			log.Fatalf("❌ Could not get region in %s %q in Namespace %q",
+				analyzeResourceJob.Kind, analyzeResourceJob.Name, analyzeResourceJob.Namespace)
+		}
+		region.Value = backingStore.Spec.AWSS3.Region
+	}
+
 	// signature version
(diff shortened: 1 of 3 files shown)


pkg/diagnostics/analyze.go
```diff
@@ -338,6 +348,16 @@ func setNamespacetoreDetailsInJob(namespaceStore *nbv1.NamespaceStore, cmd *cobr
 			namespaceStore.Kind, namespaceStore.Name, namespaceStore.Namespace)
 	}
 
+	// in case it is aws endpoint, we would pass the region
```
naveenpaul1, line 351: @shirady I see repeated code, can we move this into a method and refer to that method?

shirady: It's 2 different types One part is for namespacestore and another part is for backingstore as we have two files with similar functions. The mutual code that was not related to types was moved to different functions.

naveenpaul1: Using the interface you can pass any type and check for type and convert to the underlying type, but it might make the changes complicated,

naveenpaul1 (approved):
LGTM

PR #1213, opened 2023-08-31
Supporting core connection string in operator

jackyalbo:
### Explain the changes
1. After adding support of connection-string to noobaa-core. We adapting some of the options to support external DB to work with connection strings.
2. Now we won't create 2 secrets, the same secret created/given to the operator will also be used in 'core'
3. Added some more verifications to the connection string. Once SSL is introduced we will switch it with the operator actually trying to connect to the DB itself and these temp checks won't be needed.
4. Changing core version to be after the push of connection string to master.
5. Adding new test case to cli flow to install and uninstall noobaa connected to external DB (the test starting a pod with postgres:15 as the external DB) 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/reconciler.go
+++ b/pkg/system/reconciler.go
@@ -3,8 +3,6 @@ package system
 import (
 	"context"
 	"fmt"
-	"net"
-	"net/url"
 	"os"
 	goruntime "runtime"
 	"strings"
@@ -121,8 +119,6 @@ type Reconciler struct {
 	KedaTriggerAuthentication *kedav1alpha1.TriggerAuthentication
 	KedaScaled                *kedav1alpha1.ScaledObject
 	AdapterHPA                *autoscalingv2.HorizontalPodAutoscaler
-	PgExternalHost            string
-	PgExternalPort            string
 }
 
 // NewReconciler initializes a reconciler to be used for loading or reconciling a noobaa system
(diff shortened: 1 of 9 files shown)


pkg/system/reconciler.go
```diff
@@ -284,9 +280,10 @@ func NewReconciler(
 	r.SecretServer.StringData["jwt"] = util.RandomBase64(16)
 	r.SecretServer.StringData["server_secret"] = util.RandomHex(4)
 
-	r.SecretDB.StringData["user"] = "noobaa"
-	r.SecretDB.StringData["password"] = util.RandomBase64(10)
-	r.SecretDB.StringData["db_name"] = "nbcore"
+	if r.NooBaa.Spec.ExternalPgSecret == nil {
+		r.SecretDB.StringData["user"] = "noobaa"
+		r.SecretDB.StringData["password"] = util.RandomBase64(10)
+	}
```
liranmauda, line 286: why we don't need db name in postgres secret when we dont have the external string?

test/cli/test_cli_flow.sh
```diff
@@ -39,6 +39,9 @@ function main {
         post_install_tests
     fi
     noobaa_uninstall
+    noobaa_install_external
+    noobaa_uninstall
+    delete_external_postgres
```
liranmauda, line 44: maybe add a new test file, or add it as a different flow?
I feel that this will excceed the time we can run the GH actions, or prolong it too mach.
WDYT?

liranmauda (commented):
Maybe when using the CLI flag of external pg db, we need to verify that the option to use mongo is not set and fail if it is.

pkg/system/reconciler.go
jackyalbo: it is provided inside the connection string. something like: `postgres://<user>:<password>@<host>:<port>/<<dbname>>`

liranmauda: still not sure, when we Don't have ExternalPgSecret, don't we need it?

jackyalbo: ah no, we will use always nbcore

test/cli/test_cli_flow.sh
jackyalbo: moved to random - hope that will make it short enough

test/cli/test_cli_functions.sh
```diff
@@ -196,6 +217,15 @@ function noobaa_install {
     test_admission_deployment
 }
 
+function noobaa_install_external {
+    #noobaa timeout install # Maybe when creating server we can use local PV
```
liranmauda, line 221: We might add some printing here to make it clear that we are in external postgres mode.

jackyalbo: added print in install_external

liranmauda (approved):

PR #1214, opened 2023-09-04
Implement upgrade via CLI

Neon-White:
### Explain the changes
1. Implement the ability to upgrade the product from the CLI

### DIFF
--- a/pkg/crd/crd.go
+++ b/pkg/crd/crd.go
@@ -110,6 +110,11 @@ func RunCreate(cmd *cobra.Command, args []string) {
 	ForEachCRD(CreateCRD)
 }
 
+// RunUpgrade runs a CLI command
+func RunUpgrade(cmd *cobra.Command, args []string) {
+	ForEachCRD(UpgradeCRD)
+}
+
 // RunDelete runs a CLI command
 func RunDelete(cmd *cobra.Command, args []string) {
 	ForEachCRD(DeleteCRD)
@@ -173,6 +178,11 @@ func ForEachCRD(fn func(*CRD)) {
 // CreateCRD creates a CRD
 func CreateCRD(crd *CRD) {
 	util.KubeCreateSkipExisting(crd)
+}
+
+// UpgradeCRD Kubernetesically applies a CRD (create if doesn't exist, update otherwise)
+func UpgradeCRD(crd *CRD) {
+	util.KubeApply(crd)
 }
 
 // DeleteCRD deletes a CRD
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -2,6 +2,7 @@ package install
 
 import (
 	"fmt"
+	"time"
 
 	"github.com/noobaa/noobaa-operator/v5/pkg/backingstore"
 	"github.com/noobaa/noobaa-operator/v5/pkg/bucketclass"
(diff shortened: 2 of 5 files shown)


pkg/install/install.go
```diff
@@ -106,6 +117,25 @@ func RunInstall(cmd *cobra.Command, args []string) {
 	}
 }
 
+func RunUpgrade(cmd *cobra.Command, args []string) {
+	log := util.Logger()
+	system.RunSystemVersionsStatus(cmd, args)
```
liranmauda, line 122: This might be confusing, printing the previous version.
We might want to add a printing, something like "current versions:"
and after the upgrade run `system.RunSystemVersionsStatus(cmd, args)` again

pkg/crd/crd.go
```diff
@@ -175,6 +180,11 @@ func CreateCRD(crd *CRD) {
 	util.KubeCreateSkipExisting(crd)
 }
 
+// ApplyCRD creates a CRD
```
liranmauda, line 183:
```suggestion
// ApplyCRD will check if the CRD exists and will create/update accordingly
// and report the CRD status.
```

pkg/system/system.go
```diff
@@ -404,6 +404,71 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
+	}
+	if dbResourcesJSON != "" {
+		util.Panic(json.Unmarshal([]byte(dbResourcesJSON), &sys.Spec.DBResources))
+	}
+	if endpointResourcesJSON != "" {
+		if sys.Spec.Endpoints == nil {
+			sys.Spec.Endpoints = &nbv1.EndpointsSpec{
+				MinCount: 1,
+				MaxCount: 1,
+			}
```
liranmauda, line 433: We cannot assume this, we need to get the actual values that are currently written

pkg/system/system.go
```diff
@@ -404,6 +404,71 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
+func RunApply(cmd *cobra.Command, args []string) {
+	log := util.Logger()
+	sys := LoadSystemDefaults()
+	ns := util.KubeObject(bundle.File_deploy_namespace_yaml).(*corev1.Namespace)
+	ns.Name = sys.Namespace
+
+	coreResourcesJSON, _ := cmd.Flags().GetString("core-resources")
+	dbResourcesJSON, _ := cmd.Flags().GetString("db-resources")
+	endpointResourcesJSON, _ := cmd.Flags().GetString("endpoint-resources")
+	useOBCCleanupPolicy, _ := cmd.Flags().GetBool("use-obc-cleanup-policy")
```
liranmauda, line 417: Why do we need this in an upgrade?

pkg/system/system.go
```diff
@@ -404,6 +404,71 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
+		sys.Spec.ExternalPgSecret = &corev1.SecretReference{
+			Name:      secret.Name,
+			Namespace: secret.Namespace,
+		}
+	}
+
+	// TODO check PVC if exist and the system does not exist -
+	// fail and suggest to delete them first with cli system delete.
+	util.KubeApply(ns)
+	util.KubeApply(sys)
```
liranmauda, line 469: I don't this it is relevent for the upgrade

liranmauda (changes requested):
Some comments:
Once it will not be a draft We can have a closer look at it.
+ 
We should not forget to hide it on the downstream once it is merged.

liranmauda (approved):

PR #1216, opened 2023-09-14
Adding SSL support to External Postgres

jackyalbo:
### Explain the changes
Commit 1:
1.  Adding support to SSL with external postgres DB both for server side (ssl-required will validate the certificate of the server), and also for client side (user can use ssl-key and ssl-cert to connect with the server)

Commit 2:
1. Adding an options to run noobaa install with ssl options (both sides)
2. Adding the options to run test-cli with --dev option to make it possible to run on M1 macs

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/apis/noobaa/v1alpha1/noobaa_types.go
+++ b/pkg/apis/noobaa/v1alpha1/noobaa_types.go
@@ -121,6 +121,18 @@ type NooBaaSpec struct {
 	// +optional
 	ExternalPgSecret *corev1.SecretReference `json:"externalPgSecret,omitempty"`
 
+	// ExternalPgSSLRequired (optional) holds an optional boolean to force ssl connections to the external Postgres DB
+	// +optional
+	ExternalPgSSLRequired bool `json:"externalPgSSLRequired,omitempty"`
+
+	// ExternalPgSSLUnauthorized (optional) holds an optional boolean to allow unauthorized connections to external Postgres DB
+	// +optional
+	ExternalPgSSLUnauthorized bool `json:"externalPgSSLUnauthorized,omitempty"`
+
+	// ExternalPgSSLSecret (optional) holds an optional secret with client key and cert used for connecting to external Postgres DB
+	// +optional
+	ExternalPgSSLSecret *corev1.SecretReference `json:"externalPgSSLSecret,omitempty"`
+
 	// DebugLevel (optional) sets the debug level
 	// +optional
 	// +kubebuilder:validation:Enum=all;nsfs;warn;default_level
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -101,6 +101,20 @@ var MongoDbURL = ""
 // it can be overridden for testing or different url.
 var PostgresDbURL = ""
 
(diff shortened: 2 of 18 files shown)


pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -121,6 +121,18 @@ type NooBaaSpec struct {
...
 
+	// ExternalPgSSLRequired (optional) holds an optional boolean to force ssl connections to the external Postgres DB
+	// +optional
+	ExternalPgSSLRequired bool `json:"externalPgSSLRequired,omitempty"`
+
+	// ExternalPgSSLUnauthorized (optional) holds an optional boolean to allow unauthorized connections to external Postgres DB
+	// +optional
+	ExternalPgSSLUnauthorized bool `json:"externalPgSSLUnauthorized,omitempty"`
+
+	// ExternalPgSSLSecret (optional) holds an optional secret with client key and cert used for connecting to extrenal Postgres DB
```
liranmauda, line 132:
```suggestion
	// ExternalPgSSLSecret (optional) holds an optional secret with client key and cert used for connecting to external Postgres DB
```

pkg/options/options.go
```diff
@@ -101,6 +101,22 @@ var MongoDbURL = ""
...
+// when using an external postgres DB.
+var PostgresSSLSelfSigned = false
+
+// PostgresSSLKey is used for providing postgres url
+// it can be overridden for testing or different url.
+var PostgresSSLKey = ""
+
+// PostgresSSLCert is used for providing postgres url
+// it can be overridden for testing or different url.
+var PostgresSSLCert = ""
```
liranmauda, line 118:
```suggestion
// PostgresSSLRequired is used to force the client to work with SSL
// when using an external postgres DB.
var PostgresSSLRequired = false

// PostgresSSLSelfSigned is used to force the client to work with self-signed SSL
// when using an external postgres DB.
var PostgresSSLSelfSigned = false

// PostgresSSLKey is used for providing postgres url
// it can be overridden for testing or different url.
var PostgresSSLKey = ""

// PostgresSSLCert is used for providing postgres url
// it can be overridden for testing or a different url.
var PostgresSSLCert = ""
```

pkg/options/options.go
```diff
@@ -232,6 +248,22 @@ func init() {
 		&PostgresDbURL, "postgres-url",
 		PostgresDbURL, "url for postgresql",
 	)
+	FlagSet.BoolVar(
+		&PostgresSSLRequired, "pg-ssl-required",
+		false, "Force noobaa to work with ssl (external postgres - server-side) [if server cert is self signed, needs to add --ssl-unauthorized]",
```
liranmauda, line 253:
```suggestion
		false, "Force noobaa to work with ssl (external postgres - server-side) [if server cert is self-signed, needs to add --ssl-unauthorized]",
```

pkg/options/options.go
jackyalbo: also, most of it is incorrect. thanks, fixed

liranmauda (approved):

PR #1220, opened 2023-09-21
Adding operator checks for External DB w/o SSL

jackyalbo:
### Explain the changes
1. adding external DB validator using pq driver for Golang fix
- Will if we can connect to the DB
- Will validate the version is 15
- Will validate the collation is C (no collation)
- Will start the same connection type as we expect the core to do - so SSL (self-signed support or not and w/o client side certs) or no SSL at all.

2. Fix - adapting secret to the one in Noobaa spec

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


pkg/system/phase1_verifying.go
```diff
@@ -171,3 +201,60 @@ func (r *Reconciler) CheckJoinSecret() error {
 	}
 	return nil
 }
+
+func (r *Reconciler) CheckExternalPg(postgresDbURL string) error {
+	if r.ExternalPgSecret.StringData["db_url"] == "" {
+		return util.NewPersistentError("InvalidExternalPgSecret",
+			"ExternalPgSecret is missing db_url")
+	}
```
liranmauda, line 209: I think this is redundent

pkg/system/phase1_verifying.go
```diff
@@ -171,3 +201,60 @@ func (r *Reconciler) CheckJoinSecret() error {
...
+	// Query the PostgreSQL version
+	var version string
+	err = db.QueryRow("SELECT current_setting('server_version_num')::integer / 10000").Scan(&version)
+	if err != nil {
+		return util.NewPersistentError("InvalidExternalPgVersion",
+			fmt.Sprintf("failed getting version of external DB url: %q, error: %s",
+				db_url, err))
+	}
+	// Check if the version is 15
+	if version != "15" {
```
liranmauda, line 240: Why do we want to limit it only to this specific version? 
Can we use a min version (maybe 12?)

liranmauda (approved):
Approved, with few comments.

pkg/system/phase1_verifying.go
jackyalbo: indeed. thanks

PR #1223, opened 2023-10-10
Bug Fix BZ 2242854 | Fallback to PV Pool (CCO)

shirady:
### Explain the changes
1. Fallback to PV pool when cloud credential (CCO) was sent but didn't receive Secrets by the limited time that we defined. In the current implementation, it is 10 minutes (the number was chosen arbitrarily).
2. Record it as an event (using `recorder.Event`).
3. Save variable `secretName` for readability.

### Issues: Fixed [BZ 2242854](https://bugzilla.redhat.com/show_bug.cgi?id=2242854)
1. The option to use CCO with Manual mode was already introduced in 4.13 (as you can see in the [doc of CCO](https://docs.openshift.com/container-platform/4.13/authentication/managing_cloud_provider_credentials/cco-mode-sts.html)) and this change would be backport.

### Testing Instructions:
* Tests were on the cluster bot using `launch 4.14 aws`.

**Before this change:**
1. Install noobaa with default CCO mode - noobaa install should be Ready and we would have default backingstore type will be AWS.
2. Install noobaa with CCO mode as Manual - noobaa is stuck in phase Configuring (you can see in the logs of the operator that the Secret was not created yet (_"Secret \"noobaa-aws-cloud-creds-secret\" was not created yet by cloud-credentials operator. retry on next reconcile.."_).

**After this change:**
1. Install noobaa with default CCO mode (verify same behaviour) - noobaa install should be Ready and we would have default backingstore type will be AWS.
2. Install noobaa with CCO mode as Manual - noobaa install should be Ready and we would have default backingstore type will be PV Pool.  In the logs of the operator you'll see that is writes that it would create pv pool. In addition to that you need to see the new event.

Every step is described in detail in the [comment below](https://github.com/noobaa/noobaa-operator/pull/1223#issuecomment-1755265600).

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -44,6 +44,7 @@ const (
 	ibmLocation                           = "%s-standard"
 	ibmCOSCred                            = "ibm-cloud-cos-creds"
 	topologyConstraintsEnabledKubeVersion = "1.26.0"
+	minutesToWaitForDefaultBSCreation     = 10
 )
 
 type gcpAuthJSON struct {
@@ -763,11 +764,30 @@ func (r *Reconciler) preparePVPoolBackingStore() error {
 	return nil
 }
 
+func (r *Reconciler) defaultBSCreationTimedout(timestampCreation time.Time) bool {
+	minutesSinceCreation := time.Since(timestampCreation).Minutes()
+	return minutesSinceCreation > float64(minutesToWaitForDefaultBSCreation)
+}
+
+func (r *Reconciler) fallbackToPVPoolWithEvent(backingStoreType nbv1.StoreType, secretName string) error {
+	message := fmt.Sprintf("Failed to create default backingstore with type %s by %d minutes, "+
+		"fallback to create PV Pool backingstore",
+		backingStoreType, minutesToWaitForDefaultBSCreation)
+	additionalInfoForLogs := fmt.Sprintf(" (could not get Secret %s).", secretName)
+	r.Logger.Info(message + additionalInfoForLogs)
+	r.Recorder.Event(r.NooBaa, corev1.EventTypeWarning, "DefaultBackingStoreFailure", message)
+	if err := r.preparePVPoolBackingStore(); err != nil {
+		return err
+	}
+	return nil
+}
+
 func (r *Reconciler) prepareAWSBackingStore() error {
 	// after we have cloud credential request, wait for credentials secret
+	secretName := r.AWSCloudCreds.Spec.SecretRef.Name
 	cloudCredsSecret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      r.AWSCloudCreds.Spec.SecretRef.Name,
+			Name:      secretName,
 			Namespace: r.AWSCloudCreds.Spec.SecretRef.Namespace,
 		},
 	}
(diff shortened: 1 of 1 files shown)


shirady:
### Detailed Testing Instructions:

#### Image:
1. I have MacOS, for using the cluster-bot I need the operator image for linux/amd64 (when I run make all it builds it for arm64), hence for building the image I run: `DOCKER_DEFAULT_PLATFORM=linux/amd64 GOARCH=amd64 make all`.
2. Then tag it to my Quay account and push the image:
`docker tag noobaa/noobaa-operator:5.15.0 quay.io/<my-user>/noobaa-operator:<my-tag>`
`docker push quay.io/<my-user>/noobaa-operator:<my-tag>`

#### Install noobaa:
After I set the cluster with the configuration I've got from the cluster bot and I'm on the operator repo in my terminal I run:
`. ./devenv.sh`
Note: nb is an alias that runs the local operator from build/_output/bin (alias created by devenv).
`nb install --dev --noobaa-image='<noobaa-image-from-docker-hub>'  --operator-image='quay.io/<my-user>/noobaa-operator:<my-tag>' -n <my-namespace>`

#### Check logs of the operator:
`kubectl logs noobaa-operator-<rest-of-the-numbers-in-the-name> -n <my-namespace>`

#### Check the default backingstore type:
`kubectl get backingstore -n <my-namespace>`

Examples:
Regular AWS cluster in default mode:
```bash
NAME                           TYPE     PHASE   AGE
noobaa-default-backing-store   aws-s3   Ready   68s
```
Regular AWS cluster in manual mode:
```bash
NAME                           TYPE      PHASE      AGE
noobaa-default-backing-store   pv-pool   Creating   34s
```

#### Check the events on noobaa:
`kubectl describe noobaa noobaa -n t<namespace-name> `.

Examples:
Regular AWS cluster in default mode:
```bash
Events:
  Type    Reason       Age   From             Message
  ----    ------       ----  ----             -------
  Normal  NooBaaImage  14m   noobaa-operator  Using NooBaa image "noobaa/noobaa-core:master-20231007" for the creation of "noobaa"
```

Regular AWS cluster in manual mode:
```bash
Events:
  Type     Reason                      Age    From             Message
  ----     ------                      ----   ----             -------
  Normal   NooBaaImage                 5m48s  noobaa-operator  Using NooBaa image "noobaa/noobaa-core:master-20231007" for the creation of "noobaa"
  Warning  DefaultBackingStoreFailure  2m50s  noobaa-operator  Failed to get Secret "noobaa-aws-cloud-creds-secret" in 3 minutes, fallback to create PV Pool backingstore
```

pkg/system/phase4_configuring.go
```diff
@@ -777,6 +777,22 @@ func (r *Reconciler) prepareAWSBackingStore() error {
 		// TODO: we need to figure out why secret is not created, and react accordingly
 		// e.g. maybe we are running on azure but our CredentialsRequest is for AWS
 		r.Logger.Infof("Secret %q was not created yet by cloud-credentials operator. retry on next reconcile..", r.AWSCloudCreds.Spec.SecretRef.Name)
+
+		// in case we have a cred request but we do not get a secret
+		minutesSinceCreation := time.Since(r.NooBaa.CreationTimestamp.Time).Minutes()
+		minutesToWait := 3 // this number was chosen arbitrarily
+		if minutesSinceCreation > float64(minutesToWait) {
+			r.Logger.Infof("at least %d minutes passed since noobaa was created, would fallback to create PV Pool backingstore",
```
achouhan09, line 785: I think we can update the log info something like this
```suggestion
			r.Logger.Infof("at least %d minutes passed since noobaa was created and the secret has not been found yet, would fallback to create PV Pool backingstore",
```

achouhan09 (changes requested):

pkg/system/phase4_configuring.go
shirady: Sure, thanks.

pkg/system/phase4_configuring.go
dannyzaken, line 1: 1. we should have the fallback for all platforms and not just AWS.
2. maybe we should rely on a the creation date of the credentialRequest\objectstoreUser and not noobaa. this is a better indication to when we requested credentials.
3. I think that the timeout should be more than 3 minutes. If I remember correctly it can sometimes take some time until the CCO provide the secret

pkg/system/phase4_configuring.go
```diff
@@ -777,6 +777,22 @@ func (r *Reconciler) prepareAWSBackingStore() error {
 		// TODO: we need to figure out why secret is not created, and react accordingly
 		// e.g. maybe we are running on azure but our CredentialsRequest is for AWS
 		r.Logger.Infof("Secret %q was not created yet by cloud-credentials operator. retry on next reconcile..", r.AWSCloudCreds.Spec.SecretRef.Name)
+
+		// in case we have a cred request but we do not get a secret
+		minutesSinceCreation := time.Since(r.AWSCloudCreds.CreationTimestamp.Time).Minutes()
```
dannyzaken, line 782: I suggest refactoring a bit with some common functions. something like

```go 
if defaultBSCreationTimedout(minutesSinceCreation) {
    fallbackToPVpoolWithEvent(message)
}
```

achouhan09 (approved):

pkg/system/phase4_configuring.go
```diff
@@ -763,6 +763,20 @@ func (r *Reconciler) preparePVPoolBackingStore() error {
 	return nil
 }
 
+func (r *Reconciler) defaultBSCreationTimedout(minutesSinceCreation float64) bool {
```
dannyzaken, line 766: any reason not to pass the creation time and calculate `time.Since` inside this function?

shirady: Yes, I'm using the `minutesSinceCreation` in the message. I can change the message and not pass this information and then change the code as you suggested.

dannyzaken (approved):

dannyzaken:
@shirady please squash before merging

dannyzaken (approved):

PR #1221, opened 2023-09-21
Add upgrade process for postgres DB version 15

nadavMiz:
### Explain the changes
1. change the Postgres version from 12 to 15
2. added upgrade code for ODF

### Issues: Fixed #xxx / Gap #xxx
1. https://issues.redhat.com/browse/RHSTOR-4788

### Testing Instructions:
1. on regular image change database version in noobaa CR. should automatically upgrade and migrate the data to the new database
2. on ODF should happen automatically on version upgrade

- [ ] Doc added/updated
- [ ] Tests added

PR #1264, opened 2023-12-07
[Clone] Add upgrade process for postgres DB version 15

jackyalbo:
### Explain the changes
1. Moving efforts started in PR #1221 by @nadavMiz  to here in order to merge the code.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/internal/configmap-postgres-initdb.yaml
+++ b/deploy/internal/configmap-postgres-initdb.yaml
@@ -37,3 +37,51 @@ data:
                  -e 's/^pg_ctl\sstart.*/pg_ctl start || true/'                                   \
                     /usr/bin/run-postgresql
           su postgres -c "bash -x /usr/bin/run-postgresql"
+
+            
+  dumpdb.sh: |
+          set -e
+          sed -i -e 's/^\(postgres:[^:]\):[0-9]*:[0-9]*:/\1:10001:0:/' /etc/passwd
+          su postgres -c "bash -x /usr/bin/run-postgresql" &
+          THRESHOLD=33
+          USE=$(df -h --output=pcent "/$HOME/data" | tail -n 1 | tr -d '[:space:]%')
+          # Check if the used space is more than the threshold
+          if [ "$USE" -gt "$THRESHOLD" ]; then
+            echo "Warning: Free space $USE% is above $THRESHOLD% threshold. Can't start upgrade!"
+            exit 1
+          fi
+          echo "Info: Free space $USE% is below $THRESHOLD% threshold. Starting upgrade!"
+          until pg_isready; do sleep 1; done;
+            pg_dumpall -U postgres > /$HOME/data/dump.sql
+          exit 0
+
+  upgradedb.sh: |
+          set -e
+          PGDATA=$HOME/data/userdata
+          PGDATA_12=$HOME/data/userdata-12
+          THRESHOLD=33
+          USE=$(df -h --output=pcent "/$HOME/data" | tail -n 1 | tr -d '[:space:]%')
+          # Check if the used space is more than the threshold
+          if [ "$USE" -gt "$THRESHOLD" ]; then
+            echo "Warning: Free space $USE% is above $THRESHOLD% threshold. Can't start upgrade!"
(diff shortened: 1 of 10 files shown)


pkg/options/options.go
```diff
@@ -76,7 +76,7 @@ var DBImage = "centos/mongodb-36-centos7"
 
 // DBPostgresImage is the default postgres db image url
 // currently it can not be overridden.
-var DBPostgresImage = "centos/postgresql-12-centos7"
+var DBPostgresImage = "quay.io/sclorg/postgresql-15-c8s"
```
dannyzaken, line 79: we need to check with Boris that the downstream image is compatible with this one (from ENVs perspective, paths, etc.), it was the case for postgres12, but we need to make sure it is for 15

jackyalbo: I asked him, let's see what will be his reply

deploy/internal/configmap-postgres-initdb.yaml
```diff
@@ -37,3 +37,34 @@ data:
...
                     /usr/bin/run-postgresql
           su postgres -c "bash -x /usr/bin/run-postgresql"
+
+            
+  dumpdb.sh: |
+          set -e
+          sed -i -e 's/^\(postgres:[^:]\):[0-9]*:[0-9]*:/\1:10001:0:/' /etc/passwd
+          su postgres -c "bash -x /usr/bin/run-postgresql" &
+          until pg_isready; do sleep 1; done;
+            pg_dumpall -U postgres > /$HOME/data/dump.sql
```
dannyzaken, line 47: We need to handle the case here of not having enough space

deploy/internal/configmap-postgres-initdb.yaml
```diff
@@ -37,3 +37,34 @@ data:
...
+
+  upgradedb.sh: |
+          set -e
+          PGDATA=$HOME/data/userdata
+          PGDATA_new=$HOME/data/userdata-new
+          mv $PGDATA $PGDATA_new
+          sed -i -e 's/^\(postgres:[^:]\):[0-9]*:[0-9]*:/\1:10001:0:/' /etc/passwd
+          su postgres -c "bash -x /usr/bin/run-postgresql" &
+          until pg_isready; do sleep 1; done;
+          psql -U postgres < /$HOME/data/dump.sql
```
dannyzaken, line 58: Also here - we need to handle the case of not having enough space

pkg/system/phase2_creating.go
```diff
@@ -1482,6 +1493,187 @@ func (r *Reconciler) UpgradeMigrateDB() error {
...
+		}
+		sts.Namespace = options.Namespace
+		//postgres database doesn't exists, no need to upgrade.
+		if !util.KubeCheckQuiet(sts) {
+			phase = nbv1.UpgradePhaseNone
+			break
+		}
+
+		oldImage := sts.Spec.Template.Spec.Containers[0].Image
+		if oldImage == GetDesiredDBImage(r.NooBaa) {
```
dannyzaken, line 1583: If the upgrade will break after phase `UpgradePhasePrepare`, we will not continue the upgrade in the next `Reconcile`, since the STS image will be the same as desired

jackyalbo: @dannyzaken, maybe I missed something, but as we discussed and if there is no failthorough, we will first set the phase to prepare before we fail on updating the DB STS, no?

pkg/options/options.go
jackyalbo: @dannyzaken I discussed both with Boris and with @liranmauda and I think we should use this image for the upstream as it makes more sense: `[sclorgpostgresql-15-c9s](https://quay.io/repository/sclorg/postgresql-15-c9s)` (centos 9 like we will use RHEL 9) 
I will also re-test the happy path with the downstream image to validate it works like we expect

dannyzaken (approved):

dannyzaken:
@jackyalbo we need to add a doc explaining the changes and the flows of the upgrade

PR #1287, opened 2024-01-11
Fix - don't run db 15 upgrade for mongoDB

jackyalbo:
### Explain the changes
1. in case of MongoDB we still ran the upgrade process for Postgres - we shouldn't 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


nimrod-becker (approved):

liranmauda (approved):

liranmauda:
I think we can remove the whole mongo part from the operator, but that is another task...

Issue #1282, opened 2024-01-04
2000 reconciles every hour


vinzent:
We have NooBaa Operator 5.13.4 on OpenShift 4.13 (part of ODF). I see that the operator logs ~2000 reconciles every hour.

I counted these loglines:

* Start BackingStore Reconcile ... 
* Start BucketClass Reconcile...
* Start NooBaa system Reconcile ...

Each of them has 1/3 of the 2000 reconciles.

If watching for example the BackingStore or BucketClass resource, I can see that the `.status.conditions[].lastHearbeatTimestmap` gets updated frequently, but nothing else. 

It seems to me, 2000 reconciles every hour - when nothing changes  - is a bit too much. 

Is this expected?

liranmauda:
BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2258681

liranmauda:
Hi Tomas,
It works as expected, we can remove the log entry if it overloads.

Please take a look at the README in the PR: https://github.com/noobaa/noobaa-operator/pull/1257

Please let us know if you have any more questions (or If the README is unclear).

PR #1291, opened 2024-01-21
Remove upgradeBackingStore

liranmauda:
### Explain the changes
Remove upgradeBackingStore

1.  Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2255586
2. Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2258681
3. Fixes: https://github.com/noobaa/noobaa-operator/issues/1282

### DIFF
--- a/pkg/backingstore/backingstore.go
+++ b/pkg/backingstore/backingstore.go
@@ -1059,16 +1059,15 @@ func RunReconcile(cmd *cobra.Command, args []string) {
 	}))
 }
 
-// MapSecretToBackingStores returns a list of backingstores that uses the secret in their secretRefernce
-// used by backingstore_contorller to watch secrets changes
+// MapSecretToBackingStores returns a list of backingstores that uses the secret in their secretReference
+// used by backingstore_controller to watch secrets changes
 func MapSecretToBackingStores(secret types.NamespacedName) []reconcile.Request {
 	log := util.Logger()
-	log.Infof("checking which backingstore to reconcile. mapping secret %v to backingstores", secret)
 	bsList := &nbv1.BackingStoreList{
 		TypeMeta: metav1.TypeMeta{Kind: "BackingStoreList"},
 	}
 	if !util.KubeList(bsList, &client.ListOptions{Namespace: secret.Namespace}) {
-		log.Infof("Cloud not found backingStores in namespace %q, while trying to find Backingstore that uses %s secrte", secret.Namespace, secret.Name)
+		log.Infof("Could not found backingStores in namespace %q, while trying to find Backingstore that uses %s secret", secret.Namespace, secret.Name)
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/backingstore/backingstore.go
```diff
@@ -1088,7 +1087,6 @@ func MapSecretToBackingStores(secret types.NamespacedName) []reconcile.Request {
 			})
 		}
 	}
-	log.Infof("will reconcile these backingstores: %v", reqs)
```
jackyalbo, line 1091: I think this is helpful information for us going over the logs... maybe we can just join it together with the first one? something like: 
`log.Infof("will reconcile these backingstores: %v due to changes in the secret %v", reqs, secret)`
WDYT?

pkg/backingstore/backingstore.go
```diff
@@ -1116,7 +1113,6 @@ func MapNoobaaToBackingStores(noobaa types.NamespacedName) []reconcile.Request {
 			},
 		})
 	}
-	log.Infof("will reconcile all backingstores: %v", reqs)
```
jackyalbo, line 1119: same here: 
`log.Infof("will reconcile these backingstores: %v due to changes with noobaa system %v", reqs, noobaa)`
also can you fix the typos in line 1105 and 1071 to Could (and not Cloud) and to secret and not secrte

pkg/backingstore/reconciler.go
```diff
@@ -177,16 +176,6 @@ func (r *Reconciler) Reconcile() (reconcile.Result, error) {
 		}
 	}
 
-	oldStatefulSet := &appsv1.StatefulSet{}
```
jackyalbo, line 180: Are we sure there is no option that old systems are still out there? If only the print is the issue maybe we should change to print in case we do find the old stateful set format instead? like to run kubecheckquiet...

liranmauda: I do not see an option that we have that old of STS structure.

pkg/backingstore/backingstore.go
liranmauda: We should not log it, look at https://github.com/noobaa/noobaa-operator/issues/1282

pkg/backingstore/reconciler.go
jackyalbo: ok, as long as we have this documented

jackyalbo (approved):

pkg/backingstore/backingstore.go
vinzent: @jackyalbo @liranmauda  the issue of #1282 is not the fact that the message is logged. 2000 reconciles with a multitude of api requests with each reconcile is the issue. Currently, the noobaa ServiceAccount is the account with most API requests overall on each of our clusters with ODF. IMHO the operator should react on changes/events not constantly reconcile regardless of changes. Just my 2cents. ;-)

liranmauda: @vinzent 
I will assume this, and correct me if I am wrong:
You are only aware of noobaa reconciliation as it was reported in the logs.
I guess that if you will add printing to other components you will see the same.

The reconcile is watching for changes, which means that it needs to be aware of when (or as close as can be) to the change. 

If you are using an upstream version on your upstream ODF (is there something like that?), you can add a lot more printing to the reconcilers of all components, build that, and replace the images. 

again I will assume that it will be the same.

P.S. If you are using a downstream ODF of RedHat/IBM, do not use upstream images... (just making sure)

vinzent: > @vinzent I will assume this, and correct me if I am wrong: You are only aware of noobaa reconciliation as it was reported in the logs. I guess that if you will add printing to other components you will see the same.

Of course there are reconciles with each operator. But I would expect that most of them don't do 2'000 each hour.

Some numbers from 2024-03-12 to compare:

* Noobaa: 1 ObjectBucketClaim, no changes all day long to this ObjectBucketClaim: 890'151 api requests, 1'813'784 lines of log
* Cert-Manager: 1243 Certificates managing 1243 secrets, 24 certs created/renewed: 16'169 api requests, 295 lines of log

(Cert-Manager is probably "hiding" api requests due to caching, where noobaa doesn't cache at all?)

> 
> If you are using an upstream version on your upstream ODF (is there something like that?), you can add a lot more printing to the reconcilers of all components, build that, and replace the images.
> again I will assume that it will be the same.
> 
> P.S. If you are using a downstream ODF of RedHat/IBM, do not use upstream images... (just making sure)

Of course, we are using official ODF images. We're facing some weird issues on 2 of our ODF/Noobaa installations. I just thought that reporting suspicious behavior upstream makes sense.

liranmauda: @vinzent 
> Of course, we are using official ODF images. We're facing some weird issues on 2 of our ODF/Noobaa installations. I just thought that reporting suspicious behavior upstream makes sense.

Thank you for reporting upstream, it is appreciated. 
Upstream resolving an issue, does not guarantee which official version will be released with this fix.
Currently, our master is targeted to 5.16 (roughly equal to the 4.16 RedHat image).

Another Point is that Upstream suggestions do not consider officially supported flows. 

Any issue/question is always welcome.
I wanted to raise awareness for the differences between upstream and downstream 🙂

PR #1292, opened 2024-01-21
Moving COSI temporarily out of the code

jackyalbo:
### Explain the changes
1. Moving COSI temporarily out of the code - currently moved to comments

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/.github/workflows/run_cosi_test.yaml
+++ b/.github/workflows/run_cosi_test.yaml
@@ -1,5 +1,7 @@
 name: COSI integration Tests
-on: [push, pull_request, workflow_dispatch]
+# SHOULD BE RETURNED ONCE COSI IS BACK
+# on: [push, pull_request, workflow_dispatch]
+on: []
 
 jobs:
   run-cosi-test:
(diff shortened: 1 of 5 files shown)


liranmauda (approved):
LGTM, should we remove also the test?

jackyalbo:
> LGTM, should we remove also the test?

Still thinking about it. But probably I will also remove the test and return when we will need it.

.github/workflows/run_cosi_test.yaml
```diff
@@ -1,5 +1,7 @@
 name: COSI integration Tests
-on: [push, pull_request, workflow_dispatch]
+# SHOULD BE RETURNED ONCE COSI IS BACK
+# on: [push, pull_request, workflow_dispatch]
+on: []
```
shirady, line 4: @jackyalbo May we change it to `workflow_dispatch` meanwhile?
I receive errors in my email "No jobs were run".

PR #1303, opened 2024-02-20
bucket_logging: Bucket log processing in side car container

aspandey:
Adding a sidecar container which will run rsyslod and logrotate It will also be running a log segregate script.

### DIFF
--- a/deploy/internal/statefulset-core.yaml
+++ b/deploy/internal/statefulset-core.yaml
@@ -141,3 +141,36 @@ spec:
               valueFrom:
                 resourceFieldRef:
                   resource: limits.memory
+        - name: noobaa-log-processor
+          image: NOOBAA_CORE_IMAGE
+          command: ["/root/node_modules/noobaa-core/src/deploy/NVA_build/noobaa_logs.sh"]
+          volumeMounts:
+            - name: logs
+              mountPath: /log
+          resources:
+            requests:
+              cpu: "200m"
+              memory: "500Mi"
+            limits:
+              cpu: "200m"
+              memory: "500Mi"
+          ports:
+            - containerPort: 5140
+            - containerPort: 6514
+          env:
+            - name: CONTAINER_CPU_REQUEST
+              valueFrom:
+                resourceFieldRef:
+                  resource: requests.cpu
+            - name: CONTAINER_MEM_REQUEST
+              valueFrom:
+                resourceFieldRef:
+                  resource: requests.memory
+            - name: CONTAINER_CPU_LIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.cpu
+            - name: CONTAINER_MEM_LIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.memory
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -27,6 +27,7 @@ import (
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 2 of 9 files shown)


deploy/internal/statefulset-core.yaml
```diff
@@ -141,3 +141,23 @@ spec:
...
+            - name: logs
+              mountPath: /log
+          resources:
+            requests:
+              cpu: "900m"
+              memory: "1Gi"
+            limits:
+              cpu: "900m"
+              memory: "1Gi"
+          ports:
```
dannyzaken, line 157: you don't need all of these ports. only the syslog port

deploy/internal/statefulset-core.yaml
```diff
@@ -141,3 +141,23 @@ spec:
...
                   resource: limits.memory
+        - name: noobaa-log-processor
+          image: noobaa-core:s3-tests
+          command: ["/root/node_modules/noobaa-core/src/deploy/NVA_build/noobaa_logs.sh"]
+          volumeMounts:
+            - name: logs
+              mountPath: /log
+          resources:
+            requests:
+              cpu: "900m"
```
dannyzaken, line 152: let's use less cpu\mem for now. maybe 200m CPU and 500MB.

dannyzaken (commented):
changing the core sts yaml only handles new installations. you should also handle it programmatically in `phase2_creating.go`, `SetDesiredCoreApp`. basically you need to handle the case where there is no syslog container under `podSpec.Containers` and add it.

deploy/internal/statefulset-core.yaml
```diff
@@ -141,3 +141,23 @@ spec:
               valueFrom:
                 resourceFieldRef:
                   resource: limits.memory
+        - name: noobaa-log-processor
+          image: noobaa-core:s3-tests
```
dannyzaken, line 145: notice that this is just a place holder, and it has to be changed to the real image. you can use `NOOBAA_CORE_IMAGE` place holder as we do for core container

aspandey: Done

deploy/internal/statefulset-core.yaml
aspandey: done

deploy/internal/statefulset-core.yaml
aspandey: done

deploy/internal/statefulset-core.yaml
```diff
@@ -141,3 +141,36 @@ spec:
               valueFrom:
                 resourceFieldRef:
                   resource: limits.memory
+        - name: noobaa-log-processor
+          image: NOOBAA_CORE_IMAGE
+          # command: ["/root/node_modules/noobaa-core/src/deploy/NVA_build/noobaa_logs.sh"]
```
dannyzaken, line 146: uncomment

pkg/system/phase2_creating.go
```diff
@@ -560,6 +560,28 @@ func (r *Reconciler) SetDesiredCoreApp() error {
...
+				coreImageChanged = true
+				c.Image = r.NooBaa.Status.ActualImage
+			}
+			// adding the missing Env variable from default container
+			util.MergeEnvArrays(&c.Env, &r.DefaultCoreApp.Env)
+			r.setDesiredCoreEnv(c)
+			r.setDesiredRootMasterKeyMounts(podSpec, c)
+
+			if r.NooBaa.Spec.CoreResources != nil {
+				c.Resources = *r.NooBaa.Spec.CoreResources
```
dannyzaken, line 574: Don't use core resources for the logs container. we should probably make it configurable through noobaa CR (optionally param, same as `CoreResources`), but for now you can just use a constant amount (200m, 500MB)

dannyzaken (changes requested):
see comment regarding the resources of the logs container

pkg/system/phase2_creating.go
aspandey: Did you mean we should use hard code values 200m and 500MI in these lines?I have made changes. Please check if this is what you meant.

dannyzaken: @aspandey, yes this is what I meant. I see now that the tests are failing due to resources, because with `--mini` we use `100m` for core.  let's add `LogResources`, to noobaa spec, the same as we have [for core](https://github.com/dannyzaken/noobaa-operator/blob/76c01ce9669a6206ddf7224342e3a26e88423b23/pkg/apis/noobaa/v1alpha1/noobaa_types.go#L93-L96), where the default is `200m`. for [mini env](https://github.com/dannyzaken/noobaa-operator/blob/76c01ce9669a6206ddf7224342e3a26e88423b23/pkg/system/system.go#L215-L243) let's set it to `100m` or some other value that will make the tests pass

pkg/system/phase2_creating.go
```diff
@@ -569,6 +569,27 @@ func (r *Reconciler) SetDesiredCoreApp() error {
...
+			if c.Image != r.NooBaa.Status.ActualImage {
+				coreImageChanged = true
+				c.Image = r.NooBaa.Status.ActualImage
+			}
+			// adding the missing Env variable from default container
+			util.MergeEnvArrays(&c.Env, &r.DefaultCoreApp.Env)
+			r.setDesiredCoreEnv(c)
+
+			if r.NooBaa.Spec.LogResources != nil {
+				c.Resources = *r.NooBaa.Spec.LogResources
```
dannyzaken, line 582: `r.NooBaa.Spec.LogResources` is optional and might be `nil`. in that case use the defaults

dannyzaken (approved):

PR #1317, opened 2024-03-10
removed init container for Postgres DB pod

dannyzaken:
### Explain the changes
* The init container was used to workaround an issue with hugepages which seems to be solved (see #661).
* After removing it, there is no need for a privileged container in Postgres DB scc

1. 

### Testing Instructions:
1. See testing instructions [here](https://bugzilla.redhat.com/show_bug.cgi?id=1946792)

- [ ] Doc added/updated
- [ ] Tests added


deploy/internal/statefulset-postgres-db.yaml
```diff
@@ -19,92 +19,47 @@ spec:
...
-            memory: "500Mi"
-          limits:
-            cpu: "500m"
-            memory: "500Mi"
-        volumeMounts:
-        - name: db
-          mountPath: /var/lib/pgsql
-        - name: noobaa-postgres-initdb-sh-volume
-          mountPath: /init
       containers:
```
jackyalbo, line 22: I think removing all the init containers will result in this code failing as it expects init_containers[0] to exist:
https://github.com/noobaa/noobaa-operator/blob/6ddc25e52832366289685c4d813c35642b920c09/pkg/system/phase2_creating.go#L1537

dannyzaken:
Converted back to draft. Still have some work to do with respect to DB upgrades

liranmauda (approved):
LGTM

pkg/system/phase2_creating.go
```diff
@@ -244,20 +244,9 @@ func (r *Reconciler) SetDesiredNooBaaDB() error {
...
 	podSpec.SecurityContext.FSGroupChangePolicy = &defaultFSGroupChangePolicy
-	podSpec.InitContainers = util.FilterSlice(
-		podSpec.InitContainers,
-		func(c corev1.Container) bool {
-			// remove the init container that is not needed
-			return c.Name != "init"
-		},
-	)
+	// remove the init conatainer. It was used to workaround a hugepages issue, which was resolved in Postgres
+	podSpec.InitContainers = nil
```
jackyalbo, line 248: Do we need this?

PR #1328, opened 2024-03-27
Modify the handling of injected OCP CA bundles

Neon-White:
### Explain the changes
1.  Create the CA bundle injection configmap in the OLM stage so the operator can require it to start up, thus preventing a race condition
2. Change the CA bundle mountPath to not override system certs
3. Revert the Azure cert bundle injection (unneeded since the system certs aren't overridden anymore)
4. Change the OCP injected CA bundle configmap name
5. Change `AddToRootCAs` to append to the system pool instead of a new one
6. Remove the configmap creation logic from the operator since it's now being done by the OLM

1.  https://bugzilla.redhat.com/show_bug.cgi?id=2271580
[Note: later we moved to Jira and the link is: [DFBUGS-627](https://issues.redhat.com/browse/DFBUGS-627)]

### Testing Instructions:
1. Deploy NooBaa as part of ODF (preferably once over AWS, GCP, Azure)
2. Verify the system reaches a healthy status and finishes to install
3. Verify that the system can also use RGW backingstores

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -35,7 +35,7 @@ spec:
           name: ocp-injected-ca-bundle
           items:
           - key: ca-bundle.crt
-            path: tls-ca-bundle.pem
+            path: ca-bundle.crt
           optional: true
       containers:
         - name: noobaa-operator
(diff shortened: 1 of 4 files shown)


pkg/olm/olm.go
```diff
@@ -208,6 +208,10 @@ func RunCatalog(cmd *cobra.Command, args []string) {
 		util.Panic(os.WriteFile(dir+"noobaa-operator.package.yaml", pkgBytes, 0644))
 	}
 	util.Panic(util.WriteYamlFile(csvFileName, GenerateCSV(opConf, csvParams)))
+	// The CA configmap is needed prior to the operator startup to prevent a certificate injection race condition
+	util.Panic(util.WriteYamlFile(
```
jackyalbo, line 212: Probably needs to be only for ODF no? no ca inject outside of ODF as far as I know... 
The general question is what happens if we don't have anyone injecting the CAs to the operator? Does this mount delete the default file at /etc/pki/ca-trust/extracted/pem/tls-ca-bundle.pem? the one coming from the OS?

dannyzaken (approved):

PR #1332, opened 2024-04-02
Adding disable_cookies annotation to S3 route

### LINKED ISSUE #1327: S3 route should disable cookies otherwise Cyberduck fails on bad signature
### Environment info
- NooBaa Operator Version: **5.14**
- Platform: **ODF 4.14.5 OCP 4.14.13**

### Actual behavior
1. With Openshift 4.14 we saw Cyberduck fails on bad signature, while awscli works with same creds.
2. We saw that set-cookie was added to the replies by the Openshift haproxy (maybe depends on some cluster wide configuration, or maybe it's a new default for OCP 4.14, not sure).
3. Cyberduck sendt first connect request without cookie, S3 server response ok but haproxy injects set-cookie header to the client, Cyberduck saved the cookie and sent every next request, and when it does it also calculated the S3 signature including that cookie header (maybe a cyberduck bug, but not sure).
4. But when Cyberduck added the cookie header, HAproxy stripped it off, because it uses it for sticky sessions, and the S3 server receives the request without the cookie header, and calculated a different signature than the client.

### Expected behavior
1. The S3 route should disable cookies - 
```
oc annotate routes myroute haproxy.router.openshift.io/disable_cookies='true'
```
2. Perhaps also need to set also haproxy.router.openshift.io/balance='roundrobin' ? (not sure)
3. See docs - 
- https://docs.openshift.com/container-platform/4.14/networking/routes/route-configuration.html#nw-using-cookies-keep-route-statefulness_route-configuration
- https://docs.openshift.com/container-platform/4.14/networking/routes/route-configuration.html#nw-route-specific-annotations_route-configuration

### Steps to reproduce
1. Use OCP 4.14 and make sure sticky sessions are enabled.
2. Use Cyberduck to access the route.
3. See it fail with bad signature.
5. Enable debug log - in Cyberduck Settings -> Connection.
6. See the set-cookie and cookie headers in the log.
7. In the S3 endpoint logs see that the authorization header shows SignedHeader=cookie;host;...

jackyalbo:
### Explain the changes
1. Disable the cookies when using s3 routes as some s3 clients like Cyberduck don't handle them correctly. 

1. Fixed #1327

### Testing Instructions:
1. install a noobaa system on top of OCP cluster
2. try to connect to the system with default access keys using Cyberduck
3. see that you can connect and list the buckets and not get issues about bad signature

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/internal/route-s3.yaml
+++ b/deploy/internal/route-s3.yaml
@@ -1,6 +1,8 @@
 apiVersion: route.openshift.io/v1
 kind: Route
 metadata:
+  annotations:
+    haproxy.router.openshift.io/disable_cookies: 'true'
   labels:
     app: noobaa
   name: s3
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -4666,11 +4666,13 @@ spec:
   wildcardPolicy: None
 `
 
-const Sha256_deploy_internal_route_s3_yaml = "16050267fd5cb0a34ff7b4d849a601d2583da1a11394a94f38c4f066c1613f34"
+const Sha256_deploy_internal_route_s3_yaml = "51a2eeee88436d97847f2911a4d05077885dd94bc69c537ad67417bc823a8e20"
 
 const File_deploy_internal_route_s3_yaml = `apiVersion: route.openshift.io/v1
 kind: Route
 metadata:
+  annotations:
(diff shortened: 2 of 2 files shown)


liranmauda (approved):
LGTM, 
Could you edit the description of this PR and explain the reason?

PR #1342, opened 2024-04-12
Support user-provided NSFSAccountConfig when creating OBCs

### LINKED ISSUE #958: adding nsfs_account_config when using noobaa obc to create the claim in namespace type claim
I have a namespace type resource called say xxxx backed bucketclass

run:
noobaa obc create obc-test2 --bucketclass xxxx

it creates the bucket nicely but what I can see is how to set the 'nsfs_account_config' on the auto account like when I create this manually. The noobaa cli obc generates the account name automatically but I can see how to call the noobaa cli to nfs_account_config details.

The api endpoint don't seem to have a way and update account won't let me.

my manual account with my added config.
nsfs_account_config:
gid: 1003
new_buckets_path: bucket-path/
nsfs_only: true
uid: 1003

Neon-White:
### Explain the changes
- Refactor `LoadBucketReplicationJSON` to a generic `LoadConfigurationJSON` function
- Add CLI support for passing an NSFS configuration
- Pass NsfsAccountConfig to `CreateAccountAPI` when creating a new OBC
- Add NSFS account config validation
- Add `distinguished_name` support
- Convert `NsfsAccountConfig`'s GID and UID properties to be pointers to support a 0-value without `omitempty` getting rid of them
- Fix bug that wouldn't allow setting UID/GID to 0

All the changes are needed to support a flow where the user creates a new OBC on top of an NSFS namespacestore, and would like to use the OBC's credentials to interact with the bucket

Note: `path` is already supported

Fixes #958 

### Testing Instructions:
1. Create an NSFS namespacestore
2. Create a bucketclass that utilizes the NS from the previous step
3. Create an OBC *with NSFS account config* that utilizes the bucketclass from the previous step
4. Use the credentials provided after the OBC's creation to write to it
5. Access the filesystem, verify the object was written successfully

- [x] Doc added/updated
- [x] Tests added

### DIFF
--- a/test/cli/test_cli_functions.sh
+++ b/test/cli/test_cli_functions.sh
@@ -1420,7 +1420,7 @@ function test_create_obc_with_nsfs_acc_cfg_uid_gid_logic {
     local uid=$1
     local gid=$2
     local obc_name="giduidtestobc"
-    test_noobaa obc create ${obc_name} --uid ${uid} --gid ${gid} --bucketclass noobaa-default-backing-store
+    test_noobaa obc create ${obc_name} --uid ${uid} --gid ${gid} --bucketclass noobaa-default-bucket-class
     local obc_account_nsfs_account=$(test_noobaa api account list_accounts {} -ojson | jq '.accounts[] | select(.bucket_claim_owner | test("'${obc_name}'"))?' | jq '.nsfs_account_config')
     local account_gid=$(echo $obc_account_nsfs_account | jq '.gid')
     local account_uid=$(echo $obc_account_nsfs_account | jq '.uid')
@@ -1443,7 +1443,7 @@ function test_create_obc_with_nsfs_acc_cfg_uid_gid {
 function test_create_obc_with_nsfs_acc_distinguished_name {
     local distinguished_name="someuser"
     local obc_name="distinguishednameobc"
-    test_noobaa obc create ${obc_name} --distinguished-name $distinguished_name --bucketclass noobaa-default-backing-store
+    test_noobaa obc create ${obc_name} --distinguished-name $distinguished_name --bucketclass noobaa-default-bucket-class
     local obc_account_nsfs_account=$(test_noobaa api account list_accounts {} -ojson | jq '.accounts[] | select(.bucket_claim_owner | test("'${obc_name}'"))?' | jq '.nsfs_account_config')
     local account_distinguished_name=$(echo $obc_account_nsfs_account | jq -r '.distinguished_name')
     test_noobaa obc delete ${obc_name}


pkg/validations/common_bucket_validations.go
```diff
@@ -7,8 +7,38 @@ import (
...
 
+// ValidateNSFSAccountConfig validates that the provided NSFS config is valid
+func ValidateNSFSAccountConfig(NSFSConfig string) error {
+	log := util.Logger()
+
+	if NSFSConfig == "" {
+		return nil
+	}
+
+	var configObj nbv1.AccountNsfsConfig
```
romayalon, line 21: We should update nsfsAccountConfig to comply the newest api - (specifically distinguished_name) - 
https://github.com/noobaa/noobaa-core/blob/master/src/api/common_api.js#L1292

pkg/validations/common_bucket_validations.go
```diff
@@ -7,8 +7,38 @@ import (
...
+	}
+
+	// GID validation
+	if configObj.GID < 0 {
+		return fmt.Errorf("GID must be a whole positive number")
+	}
+
+	return nil
+}
+
```
romayalon, line 41: Please add unit tests

pkg/validations/common_bucket_validations.go
```diff
@@ -7,8 +7,38 @@ import (
...
+	if configObj.UID < 0 {
+		return fmt.Errorf("UID must be a whole positive number")
+	}
+
+	// GID validation
+	if configObj.GID < 0 {
+		return fmt.Errorf("GID must be a whole positive number")
+	}
+
+	return nil
```
romayalon, line 39: please apply these changes on the relevant docs

pkg/obc/obc.go
```diff
@@ -176,13 +177,21 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
 	if replicationPolicy != "" {
-		replication, err := util.LoadBucketReplicationJSON(replicationPolicy)
+		replication, err := util.LoadConfigurationJSON(replicationPolicy)
 		if err != nil {
 			log.Fatalf(`❌ %q`, err)
 		}
 		obc.Spec.AdditionalConfig["replicationPolicy"] = replication
 	}
 
+	if nsfsConfig != "" {
```
romayalon, line 187: what about supporting it on COSI? OBCs won't be with as for long..

pkg/validations/common_bucket_validations.go
romayalon: and add some more validations for the new addition

romayalon (changes requested):

doc/obc-provisioner.md
```diff
@@ -101,7 +101,39 @@ spec:
...
+It is possible to create an OBC on top of NSFS. In order to do this, the user has to provide relevant data in `nsfsAccountConfig` under the `spec.additionalConfig` key.
+
+The config needs to contain the following fields so that NooBaa can access the filesystem:
+gid - the group ID of the account that should be mimicked within the filesystem
+uid - the user ID of the account that should be mimicked within the filesystem
+OR
+distinguished_name - the distinguished name of the account that should be mimicked within the filesystem
+
+These optional parameters can be provided by the user to further configure the NSFS account:
+new_buckets_path - the filesystem path that should be 'mounted' as the bucket's root (`path` parameter in the CLI)
```
romayalon, line 118: @Neon-White I don't think that we would like to let the user create other bucket's in the OBC's path..
IMO we should not let this account create any buckets since it's created for accessing this OBC. WDYT?

romayalon: Actually the current behavior of this account is [AllowBucketCreation=false](https://github.com/noobaa/noobaa-operator/pull/1342/files#diff-fc66f732e7eb85134eb34fcca0e5f6a5de879ec044b8e0d194de1c31e35b3c34R589) so this means that new_buckets_path has no meaning, thanks for removing it

pkg/apis/noobaa/v1alpha1/noobaaaccount_types.go
```diff
@@ -74,10 +74,11 @@ type NooBaaAccountSpec struct {
 
 // AccountNsfsConfig is the configuration of NSFS of CreateAccountParams
 type AccountNsfsConfig struct {
-	UID            int    `json:"uid"`
-	GID            int    `json:"gid"`
-	NewBucketsPath string `json:"new_buckets_path"`
-	NsfsOnly       bool   `json:"nsfs_only"`
+	UID               int    `json:"uid,omitempty"`
```
romayalon, line 77: Why this is needed? If I understand correctly we won't create a CR account (we will create only the internal db account entity)

pkg/validations/common_bucket_validations.go
romayalon: As we discussed on Slack - 
validation tests to ValidateNSFSAccountConfig()
we have integration tests also (cli flow)

romayalon (approved):
approved :)
please don't forget to squash the commits before merge :)

romayalon:
@Neon-White please next time don't merge without squashing :(

PR #1343, opened 2024-04-15
Fixing AzureClientCertPath in kms for azure

vh05:
Fixing AzureClientCertPath in kms for azure

The azure kms is expecting the secret to be present
at key azure.AzureClientCertPath. We were storing the
secret with wrong key which which was stopping azure kms
to read the secret the from expected key.

This patch handles the correction

Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2275049


liranmauda (approved):
LGTM

PR #1344, opened 2024-04-15
Support to update credentials manually to an account

achouhan09:
### Explain the changes
1. Added a CLI command to manually update account credentials (access key and secret key). Users can provide the access key and secret key as flags, or they will be prompted to enter them if the flags are not specified.
For example: `noobaa account credentials --access-key=<access-key> --secret-key=<secret-key>`

Additionally, validation checks have been implemented for both keys to ensure the command succeeds only with valid credentials.

### Testing Instructions:
1. `make test-cli-flow`

- [X] Tests added

### DIFF
--- a/pkg/nb/types.go
+++ b/pkg/nb/types.go
@@ -427,6 +427,12 @@ type GenerateAccountKeysParams struct {
 	Email string `json:"email"`
 }
 
+// UpdateAccountKeysParams is the params of account_api.update_account_keys()
+type UpdateAccountKeysParams struct {
+	Email      string       `json:"email"`
+	AccessKeys S3AccessKeys `json:"access_keys"`
+}
+
 // ResetPasswordParams is the params of account_api.reset_password()
 type ResetPasswordParams struct {
 	Email                string       `json:"email"`
--- a/pkg/noobaaaccount/noobaaaccount.go
+++ b/pkg/noobaaaccount/noobaaaccount.go
@@ -36,6 +36,7 @@ func Cmd() *cobra.Command {
 		CmdCreate(),
 		CmdUpdate(),
 		CmdRegenerate(),
+		CmdCredentials(),
 		CmdPasswd(),
 		CmdDelete(),
 		CmdStatus(),
@@ -83,6 +84,24 @@ func CmdRegenerate() *cobra.Command {
 		Short: "Regenerate S3 Credentials",
 		Run:   RunRegenerate,
 	}
+	return cmd
+}
+
+// CmdCredentials returns a CLI command
+func CmdCredentials() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "credentials <noobaa-account-name>",
+		Short: "Update S3 Credentials (access-key and secret-key)",
+		Run:   RunCredentials,
+	}
+	cmd.Flags().String(
+		"access-key", "",
+		`Access key for authentication - The best practice is to **omit this flag**. In that case, the CLI will prompt to securely read it from the terminal, avoiding the risk of leaking secrets in the shell history.`,
+	)
+	cmd.Flags().String(
+		"secret-key", "",
+		`Secret key for authentication - The best practice is to **omit this flag**. In that case, the CLI will prompt to securely read it from the terminal, avoiding the risk of leaking secrets in the shell history.`,
+	)
 	return cmd
 }
 
@@ -372,6 +391,45 @@ func RunRegenerate(cmd *cobra.Command, args []string) {
 		RunStatus(cmd, args)
 	}
 
+}
+
+// RunCredentials runs a CLI command
+func RunCredentials(cmd *cobra.Command, args []string) {
+	log := util.Logger()
+
+	if len(args) != 1 || args[0] == "" {
+		log.Fatalf(`❌ Missing expected arguments: <noobaa-account-name> %s`, cmd.UsageString())
+	}
+
+	var accessKeys nb.S3AccessKeys
+	accessKeys.AccessKey = nb.MaskedString(util.GetFlagStringOrPromptPassword(cmd, "access-key"))
+	accessKeys.SecretKey = nb.MaskedString(util.GetFlagStringOrPromptPassword(cmd, "secret-key"))
+
+	if accessKeys.AccessKey == "" || accessKeys.SecretKey == "" {
+		log.Fatalf(`❌ access_key and secret_key flags must be provided`)
+	}
+	// validating access_keys
+	ValidateAccessKeys(accessKeys)
+
+	name := args[0]
+	o := util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_noobaaaccount_cr_yaml)
+	noobaaAccount := o.(*nbv1.NooBaaAccount)
+	noobaaAccount.Name = name
+	noobaaAccount.Namespace = options.Namespace
+
+	if !util.KubeCheck(noobaaAccount) && (name != "admin@noobaa.io") {
+		err := UpdateNonCrdAccountKeys(name, accessKeys)
+		if err != nil {
+			log.Fatalf(`❌ Could not update credentials for %q: %v`, name, err)
+		}
+	} else {
+		err := UpdateAccountKeys(name, accessKeys)
+		if err != nil {
+			log.Fatalf(`❌ Could not update credentials for %q: %v`, name, err)
+		}
+
+		RunStatus(cmd, args)
+	}
 }
 
 // RunPasswd runs a CLI command
@@ -705,6 +763,114 @@ func GenerateNonCrdAccountKeys(name string) error {
 	fmt.Printf("AWS_SECRET_ACCESS_KEY : %s\n\n", accessKeys.SecretKey)
 
 	return nil
+}
+
+// UpdateAccountKeys update noobaa account (CRD based) S3 keys
+func UpdateAccountKeys(name string, accessKeys nb.S3AccessKeys) error {
+	log := util.Logger()
+
+	sysClient, err := system.Connect(true)
+	if err != nil {
+		return err
+	}
+
+	// Checking that we can find the secret before we are calling the RPC to update the credentials.
+	secret := util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret)
+	secret.Namespace = options.Namespace
+	// Handling a special case when the account is "admin@noobaa.io" we don't have CRD but have a secret
+	if name == "admin@noobaa.io" {
+		secret.Name = "noobaa-admin"
+	} else {
+		secret.Name = fmt.Sprintf("noobaa-account-%s", name)
+	}
+	if !util.KubeCheckQuiet(secret) {
+		log.Fatalf(`❌  Could not find secret: %s, will not update keys.`, secret.Name)
+	}
+
+	err = sysClient.NBClient.UpdateAccountKeysAPI(nb.UpdateAccountKeysParams{
+		Email:      name,
+		AccessKeys: accessKeys,
+	})
+	if err != nil {
+		return err
+	}
+
+	// UpdateAccountKeysAPI have no reply so we need to read the account in order to get the new credentials
+	accountInfo, err := sysClient.NBClient.ReadAccountAPI(nb.ReadAccountParams{
+		Email: name,
+	})
+	if err != nil {
+		return err
+	}
+
+	accessKeys = accountInfo.AccessKeys[0]
+
+	secret.StringData = map[string]string{}
+	secret.StringData["AWS_ACCESS_KEY_ID"] = string(accessKeys.AccessKey)
+	secret.StringData["AWS_SECRET_ACCESS_KEY"] = string(accessKeys.SecretKey)
+
+	// If we will not be able to update the secret we will print the credentials as they already been changed by the RPC
+	if !util.KubeUpdate(secret) {
+		log.Printf(`❌  Please verify the updated credentials for account %s:`, name)
+		fmt.Printf("\nAWS_ACCESS_KEY_ID     : %s\n", string(accessKeys.AccessKey))
+		fmt.Printf("AWS_SECRET_ACCESS_KEY : %s\n\n", string(accessKeys.SecretKey))
+		log.Fatalf(`❌  Failed to update the secret %s with the new accessKeys`, secret.Name)
+	}
+
+	log.Printf("✅ Successfully updated s3 credentials for the account %q", name)
+	return nil
+}
+
+// UpdateNonCrdAccountKeys update noobaa account (none CRD based) S3 keys
+func UpdateNonCrdAccountKeys(name string, accessKeys nb.S3AccessKeys) error {
+	log := util.Logger()
+
+	sysClient, err := system.Connect(true)
+	if err != nil {
+		return err
+	}
+
+	err = sysClient.NBClient.UpdateAccountKeysAPI(nb.UpdateAccountKeysParams{
+		Email:      name,
+		AccessKeys: accessKeys,
+	})
+	if err != nil {
+		if nbErr, ok := err.(*nb.RPCError); ok && nbErr.RPCCode == "NO_SUCH_ACCOUNT" {
+			log.Fatalf(`❌  Could not find the account: %s, will not update keys.`, name)
+		}
+		return err
+	}
+
+	// UpdateAccountKeysAPI have no reply so we need to read the account in order to get the updated credentials
+	accountInfo, err := sysClient.NBClient.ReadAccountAPI(nb.ReadAccountParams{
+		Email: name,
+	})
+	if err != nil {
+		log.Fatalf(`❌  Could not read account: %s, keys were allready updated, please read the account to get the keys`, name)
+	}
+
+	accessKeys = accountInfo.AccessKeys[0]
+
+	log.Printf("✅ Successfully update s3 credentials for the account %q", name)
+	fmt.Printf("\nAWS_ACCESS_KEY_ID     : %s\n", string(accessKeys.AccessKey))
+	fmt.Printf("AWS_SECRET_ACCESS_KEY : %s\n\n", string(accessKeys.SecretKey))
+
+	return nil
+}
+
+// ValidateAccessKeys checks validity of credentials (access_key and secret_key)
+func ValidateAccessKeys(accessKeys nb.S3AccessKeys) {
+	log := util.Logger()
+
+	// validations for access_key
+	if !util.AccessKeyRegexp.MatchString(string(accessKeys.AccessKey)) {
+		log.Fatalf(`❌ Account access key length must be 20, and must contain only alpha-numeric chars`)
+	}
+
+	// validations for secret_key
+	if !util.SecretKeyRegexp.MatchString(string(accessKeys.SecretKey)) {
+		log.Fatalf(`❌ Account secret length must be 40, and must contain only alpha-numeric chars, "+", "/"`)
+	}
 }
 
 // ResetPassword reset noobaa account password
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -90,6 +90,12 @@ type ValidationError struct {
 	Msg string
 }
 
+// AccessKeyRegexp validates access keys, which are 20 characters long and may include alphanumeric characters
+var AccessKeyRegexp, _ = regexp.Compile(`^[a-zA-Z0-9]{20}$`)
+
+// SecretKeyRegexp validates secret keys, which are 40 characters long and may include alphanumeric characters '+' and '/'
+var SecretKeyRegexp, _ = regexp.Compile(`^[a-zA-Z0-9+/]{40}$`)
+
 // IsValidationError check if err is of type ValidationError
 func IsValidationError(err error) bool {
 	_, ok := err.(ValidationError)
--- a/pkg/nb/api.go
+++ b/pkg/nb/api.go
@@ -64,6 +64,7 @@ type Client interface {
 	DeleteBucketReplicationAPI(DeleteBucketReplicationParams) error
 
 	GenerateAccountKeysAPI(GenerateAccountKeysParams) error
+	UpdateAccountKeysAPI(UpdateAccountKeysParams) error
 	ResetPasswordAPI(ResetPasswordParams) error
 }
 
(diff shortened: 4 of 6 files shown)


pkg/noobaaaccount/noobaaaccount.go
shirady, line 1: There is the part of `cmd.AddCommand` and I don't see the `CmdCredentials` in it.
Try to run `noobaa account --help` and make sure you see the added command.

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -374,6 +386,45 @@ func RunRegenerate(cmd *cobra.Command, args []string) {
...
+	var accessKeys nb.S3AccessKeys
+
+	accessKeys.AccessKey, _ = cmd.Flags().GetString("access_key")
+	accessKeys.SecretKey, _ = cmd.Flags().GetString("secret_key")
+
+	if accessKeys.AccessKey == "" || accessKeys.SecretKey == "" {
+		log.Fatalf(`❌ access_key and secret_key flags must be provided`)
+	}
+
+	if !util.KubeCheck(noobaaAccount) && (name != "admin@noobaa.io") {
```
shirady, line 413: Maybe set the condition with a non-negative condition?
if we have CRD or if the name is admin@noobaa.io... (the code that you have in the "else" clause).

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -707,6 +758,99 @@ func GenerateNonCrdAccountKeys(name string) error {
...
+
+	err = sysClient.NBClient.UpdateAccountKeysAPI(nb.UpdateAccountKeysParams{
+		Email:      name,
+		AccessKeys: accessKeys,
+	})
+	if err != nil {
+		return err
+	}
+
+	// UpdateAccountKeysAPI have no replay so we need to read the account in order to get the new credentials
```
shirady, line 791: typo - reply

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -86,6 +86,18 @@ func CmdRegenerate() *cobra.Command {
 	return cmd
 }
 
+// CmdCredentials returns a CLI command
+func CmdCredentials() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "credentials <noobaa-account-name>",
+		Short: "Update S3 Credentials",
```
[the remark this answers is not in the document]
achouhan09: Sure, will do. Thanks

pkg/noobaaaccount/noobaaaccount.go
achouhan09: Forgot to add it, will add. Thanks

pkg/noobaaaccount/noobaaaccount.go
achouhan09: Sure, thanks.

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -80,12 +81,24 @@ func CmdUpdate() *cobra.Command {
 func CmdRegenerate() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "regenerate <noobaa-account-name>",
-		Short: "Regenerate S3 Credentials",
+		Short: "Regenerate S3 Credentials(access-key and secret-key)",
```
shirady, line 84: You added in `CmdRegenerate` instead of `CmdCredentials`.
Anyway, I think you can remove it, because now I understand the whole menu is with "credential" as a known term, so no need to explain.

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -707,6 +759,99 @@ func GenerateNonCrdAccountKeys(name string) error {
 	return nil
 }
 
+// UpdateAccountKeys update noobaa account (CRD based) S3 keys
+func UpdateAccountKeys(name string, accessKeys nb.S3AccessKeys) error {
```
shirady, line 763: Why public function?

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -707,6 +759,99 @@ func GenerateNonCrdAccountKeys(name string) error {
...
+		fmt.Printf("AWS_SECRET_ACCESS_KEY : %s\n\n", accessKeys.SecretKey)
+		log.Fatalf(`❌  Failed to update the secret %s with the new accessKeys`, secret.Name)
+	}
+
+	log.Printf("✅ Successfully updated s3 credentials for the account %q", name)
+	return nil
+}
+
+// UpdateNonCrdAccountKeys update noobaa account (none CRD based) S3 keys
+func UpdateNonCrdAccountKeys(name string, accessKeys nb.S3AccessKeys) error {
```
shirady, line 819: Why public function?

shirady, line 819: I think there are mutual parts between the functions `UpdateNonCrdAccountKeys` and `UpdateAccountKeys`.
What about using one function and adding a variable for the case that you handle in `UpdateAccountKeys` and not in `UpdateNonCrdAccountKeys` (the secret)

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -374,6 +387,45 @@ func RunRegenerate(cmd *cobra.Command, args []string) {
...
+	noobaaAccount.Namespace = options.Namespace
+
+	var accessKeys nb.S3AccessKeys
+
+	accessKeys.AccessKey, _ = cmd.Flags().GetString("access_key")
+	accessKeys.SecretKey, _ = cmd.Flags().GetString("secret_key")
+
+	if accessKeys.AccessKey == "" || accessKeys.SecretKey == "" {
+		log.Fatalf(`❌ access_key and secret_key flags must be provided`)
+	}
```
shirady, line 412: What did you think about adding a check here as well? We would have fewer RPC calls.
And also explain the credentials requirements.

shirady:
@achouhan09, 
General question - At which stage will the user set the credentials manually to the account?

pkg/noobaaaccount/noobaaaccount.go
achouhan09: Sure, will update it. Thanks

pkg/noobaaaccount/noobaaaccount.go
achouhan09: I am not sure about this, all the functions in this file are public because of there reusability and easy access.

achouhan09: Yes we can do that. I am following the same approach here(as we did for Regenerate). But its our call if we want to change it then we can change it. Also later we need to do the same for Regenerate I think.

pkg/noobaaaccount/noobaaaccount.go
achouhan09: Yes, I am thinking of the same. I will add here as well. Thanks

pkg/noobaaaccount/noobaaaccount.go
shirady: I understand, so you can keep it.

achouhan09:
> @achouhan09, General question - At which stage will the user set the credentials manually to the account?

pkg/nb/types.go
```diff
@@ -427,6 +427,12 @@ type GenerateAccountKeysParams struct {
 	Email string `json:"email"`
 }
 
+// UpdateAccountKeysParams is the params of account_api.update_account_keys()
+type UpdateAccountKeysParams struct {
+	Email      string       `json:"email"`
+	AccessKeys S3AccessKeys `json:"access_keys"`
+}
```
shirady, line 434: Do you see the access key and secret key printings in the operator logs after the RPC call?
Please notice PR #1353, in case you see this I think you should wait for the PR to be merged and use the same solution.
cc: @liranmauda

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -86,6 +88,18 @@ func CmdRegenerate() *cobra.Command {
...
 
+// CmdCredentials returns a CLI command
+func CmdCredentials() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "credentials <noobaa-account-name>",
+		Short: "Update S3 Credentials(access-key and secret-key)",
+		Run:   RunCredentials,
+	}
+	cmd.Flags().String("access_key", "", "Update the access_key for an account")
+	cmd.Flags().String("secret_key", "", "Update the secret_key for an account")
```
shirady, line 99: I think you should warn the user that using the flag is not secure, for example:
https://github.com/noobaa/noobaa-operator/blob/fcc5240f5af690596823f2ab63e9debd9a640fda/pkg/namespacestore/namespacestore.go#L82

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -374,6 +388,48 @@ func RunRegenerate(cmd *cobra.Command, args []string) {
...
+
+	o := util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_noobaaaccount_cr_yaml)
+	noobaaAccount := o.(*nbv1.NooBaaAccount)
+	noobaaAccount.Name = name
+	noobaaAccount.Namespace = options.Namespace
+
+	var accessKeys nb.S3AccessKeys
+
+	accessKeys.AccessKey, _ = cmd.Flags().GetString("access_key")
+	accessKeys.SecretKey, _ = cmd.Flags().GetString("secret_key")
```
shirady, line 409: What do you think about using `util.GetFlagStringOrPromptPassword` to avoid this printing in the shell? for example:
https://github.com/noobaa/noobaa-operator/blob/fcc5240f5af690596823f2ab63e9debd9a640fda/pkg/namespacestore/namespacestore.go#L447-L448

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -707,6 +763,152 @@ func GenerateNonCrdAccountKeys(name string) error {
...
+	}
+
+	// Checking complexity for secret_key
+	if !ValidateComplexity(accessKeys.SecretKey, 40, true, true, true, true) {
+		log.Fatalf(`❌ Account secret length must be 40, and must contain uppercase, lowercase, numbers and symbols`)
+	}
+}
+
+// ValidateComplexity checks for complexity
+func ValidateComplexity(str string, length int, upper bool, lower bool, numbers bool, symbols bool) bool {
```
shirady, line 875: To avoid the long list of variables in the function signature maybe use a struct.
Since you are passing the list of boolean, which is hard-coded, I think it would be better to use defaults like you have in core:
https://github.com/noobaa/noobaa-core/blob/ff0007ad2512b0f67efa80112ceb2fcea0f3a63a/src/util/string_utils.js#L226-L242

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -707,6 +763,152 @@ func GenerateNonCrdAccountKeys(name string) error {
...
+
+	if len(str) < length {
+		return false
+	}
```
shirady, line 879: Please notice that in the core repo, you are using `validate_complexity` which uses `validate_length` which checks `len(str) != length`, see:
https://github.com/noobaa/noobaa-core/blob/ff0007ad2512b0f67efa80112ceb2fcea0f3a63a/src/util/string_utils.js#L156-L161

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -707,6 +763,152 @@ func GenerateNonCrdAccountKeys(name string) error {
...
+
+	// Check uppercase
+	if upper {
+		if matched, _ := regexp.MatchString("[A-Z]", str); !matched {
```
shirady, line 883: I think that creating a variable for the regex is more readable.
For example:
https://github.com/noobaa/noobaa-core/blob/ff0007ad2512b0f67efa80112ceb2fcea0f3a63a/src/util/string_utils.js#L11C2-L11C2

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -707,6 +763,152 @@ func GenerateNonCrdAccountKeys(name string) error {
...
+		log.Fatalf(`❌ Account access key length must be 20, and must contain uppercase and numbers`)
+	}
+
+	// Checking complexity for secret_key
+	if !ValidateComplexity(accessKeys.SecretKey, 40, true, true, true, true) {
+		log.Fatalf(`❌ Account secret length must be 40, and must contain uppercase, lowercase, numbers and symbols`)
+	}
+}
+
+// ValidateComplexity checks for complexity
```
shirady, line 874: I think you can add more description

pkg/noobaaaccount/noobaaaccount.go
shirady, line 875: I think you can move this function to `util.go`, if you want to can also separate it for smaller functions like in the core repo:
https://github.com/noobaa/noobaa-core/blob/ff0007ad2512b0f67efa80112ceb2fcea0f3a63a/src/util/string_utils.js#L236-L240

pkg/noobaaaccount/noobaaaccount.go
achouhan09: Make sense, will update it. Thanks

pkg/noobaaaccount/noobaaaccount.go
achouhan09: Yes, I think it is a good idea to get keys using prompt and also not printing them on terminal. Thanks, will update it.

pkg/noobaaaccount/noobaaaccount.go
achouhan09: yes I will move it there.

achouhan09: Sure, Thanks.

pkg/util/util.go
```diff
@@ -90,6 +90,56 @@ type ValidationError struct {
...
+	Lower   bool
+	Numbers bool
+	Symbols bool
+}
+
+// ValidateComplexity checks if a string meets specified complexity requirements.
+// It takes a string 'str' and an 'options' struct containing criteria such as length,
+// uppercase, lowercase, numbers, and symbols. Returns true if the string meets the
+// criteria, otherwise false.
+func ValidateComplexity(str string, options Options) bool {
```
shirady, line 106: Please notice the changes in core in PR [noobaa/noobaa-core/#8008](https://github.com/noobaa/noobaa-core/pull/8008), see here:
https://github.com/noobaa/noobaa-core/blob/56e17689c9937ee91f5516b8ed73f1ceb1eee6e5/src/util/string_utils.js#L11-L12

achouhan09: Thanks, I will update core PR accordingly.

pkg/noobaaaccount/noobaaaccount.go
achouhan09: Updated the code to use a single regex expression instead of multiple functions, just like noobaa-core.Thanks

pkg/noobaaaccount/noobaaaccount.go
achouhan09: Updated. Thanks

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -86,6 +87,24 @@ func CmdRegenerate() *cobra.Command {
 	return cmd
 }
 
+// CmdCredentials returns a CLI command
+func CmdCredentials() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "credentials <noobaa-account-name>",
+		Short: "Update S3 Credentials(access-key and secret-key)",
```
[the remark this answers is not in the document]
achouhan09: Done.

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -86,6 +87,24 @@ func CmdRegenerate() *cobra.Command {
...
+		Run:   RunCredentials,
+	}
+	cmd.Flags().String(
+		"access-key", "",
+		`Access key for authentication - the best practice is to **omit this flag**, in that case the CLI will prompt to prompt and read it securely from the terminal to avoid leaking secrets in the shell history`,
```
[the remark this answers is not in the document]
achouhan09: Sure. Thanks

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -86,6 +87,24 @@ func CmdRegenerate() *cobra.Command {
...
+	)
+	cmd.Flags().String(
+		"secret-key", "",
+		`Secret key for authentication - the best practice is to **omit this flag**, in that case the CLI will prompt to prompt and read it securely from the terminal to avoid leaking secrets in the shell history`,
```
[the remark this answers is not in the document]
achouhan09: Sure, Thanks.

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -707,6 +769,116 @@ func GenerateNonCrdAccountKeys(name string) error {
...
+
+	return nil
+}
+
+// ValidateAccessKeys checks for complexity of credentials(access_key and secret_key)
+func ValidateAccessKeys(accessKeys nb.S3AccessKeys) {
+	log := util.Logger()
+
+	// Checking complexity for access_key
+	regex := regexp.MustCompile(util.AccessKeyRegexp)
```
shirady, line 870: Why did you prefer to use `MustCompile` and not `Compile`?
As I understand it `Compile` (in this case `MustCompile`) should be run in the "init" stage and then you can reuse the regexp, in this function you compile and recreate it on every call ([here is a simple example](https://pkg.go.dev/regexp#example-package)).

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -707,6 +769,116 @@ func GenerateNonCrdAccountKeys(name string) error {
...
+	accessKeys = accountInfo.AccessKeys[0]
+
+	log.Printf("✅ Successfully update s3 credentials for the account %q", name)
+	fmt.Printf("\nAWS_ACCESS_KEY_ID     : %s\n", accessKeys.AccessKey)
+	fmt.Printf("AWS_SECRET_ACCESS_KEY : %s\n\n", accessKeys.SecretKey)
+
+	return nil
+}
+
+// ValidateAccessKeys checks for complexity of credentials(access_key and secret_key)
```
shirady, line 865: In the core repo the term "complexity" was removed.

pkg/noobaaaccount/noobaaaccount.go
achouhan09: Yes, It should be the better way. Thanks will update it.

pkg/noobaaaccount/noobaaaccount.go
achouhan09: I will remove it from operator. Thanks

pkg/nb/types.go
achouhan09: I cannot see any of them(access_key and secret_key) printing in operator logs and in core logs printed as some sensitive string. I think we need not to make changes in code for this. Thanks

shirady: As discussed they currently appear as printing of the RPC request details in the CLI.

liranmauda: Great Catch,
It is unrelated to this PR.
@shirady Could you open an Issue for it? 
@achouhan09 We can proceed with this PR

shirady: @liranmauda sure
@achouhan09 resolving according to the comment above.

liranmauda: https://github.com/noobaa/noobaa-operator/pull/1357 should address this issue

achouhan09: Thanks Liran, Should I need to update the type to `MaskedString` for accessKeys?

shirady: After PR #1357 will be merged it will work, you will not need to change something in your code.

shirady (approved):
LGTM

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -374,6 +393,48 @@ func RunRegenerate(cmd *cobra.Command, args []string) {
...
+func RunCredentials(cmd *cobra.Command, args []string) {
+	log := util.Logger()
+
+	if len(args) != 1 || args[0] == "" {
+		log.Fatalf(`❌ Missing expected arguments: <noobaa-account-name> %s`, cmd.UsageString())
+	}
+
+	name := args[0]
+
+	o := util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_noobaaaccount_cr_yaml)
```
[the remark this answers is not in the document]
achouhan09: Make sense, I will update it. Thanks

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -374,6 +393,48 @@ func RunRegenerate(cmd *cobra.Command, args []string) {
...
+	accessKeys.SecretKey = nb.MaskedString(util.GetFlagStringOrPromptPassword(cmd, "secret-key"))
+
+	if accessKeys.AccessKey == "" || accessKeys.SecretKey == "" {
+		log.Fatalf(`❌ access_key and secret_key flags must be provided`)
+	}
+
+	// validating access_keys
+	ValidateAccessKeys(accessKeys)
+
+	if !util.KubeCheck(noobaaAccount) && (name != "admin@noobaa.io") {
```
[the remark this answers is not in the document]
achouhan09: I think it will take lots of code changes as we have used it in various places, will do it in separate PR if required. Thanks

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -701,12 +762,120 @@ func GenerateNonCrdAccountKeys(name string) error {
...
+	}
+
+	accessKeys = accountInfo.AccessKeys[0]
 
+	secret.StringData = map[string]string{}
+	secret.StringData["AWS_ACCESS_KEY_ID"] = string(accessKeys.AccessKey)
+	secret.StringData["AWS_SECRET_ACCESS_KEY"] = string(accessKeys.SecretKey)
+
+	// If we will not be able to update the secret we will print the credentials as they already been changed by the RPC
+	if !util.KubeUpdate(secret) {
```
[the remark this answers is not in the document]
achouhan09: Not relevant to this PR, will do it in another PR. Thanks.

PR #1359, opened 2024-05-09
[5.15] SCC Fix

tangledbytes:
### Explain the changes
In the master branch and 5.16, because we do no longer have an init container, we no longer require a lot of privileges to operate. The same is **not** true for versions before and including 5.15.

I tried multiple combinations of:
```
AUDIT_WRITE
CHOWN
DAC_OVERRIDE
FOWNER
FSETID
KILL
MKNOD
NET_BIND_SERVICE
SETFCAP
SETGID
SETPCAP
SETUID
SYS_CHROOT
```
capabilities but nothing really worked hence adding back the dropped default privileges. Will seek exception for this container (and I think we did do that initially for 5.14).

1. Broke 4.15.3 RC build - fixing that

### Testing Instructions:
1. Created a fresh openshift-cluster from cluster bot.
2. Used CLI built from this patch and ran `nb install ...`.
3. Waited for the system to be stable.
4. Ran `aws s3 ls` against the deployment.
5. Created obc and then S3 ops against that bucket using admin credentials.

The above testing was done to ensure that system is indeed functional at least in 5.15.

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/scc_db.yaml
+++ b/deploy/scc_db.yaml
@@ -10,8 +10,6 @@ allowHostPID: false
 allowHostPorts: false
 allowPrivilegedContainer: false
 readOnlyRootFilesystem: false
-requiredDropCapabilities:
-  - ALL
 fsGroup:
   type: RunAsAny
 runAsUser:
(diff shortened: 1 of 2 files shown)


tangledbytes:
Please note that we will need to request exception for the DB pod just like we did for noobaa-endpoint pod. The same isn't true for the recent versions of NooBaa though.

nimrod-becker (approved):

PR #1362, opened 2024-05-21
Added BucketLogging type to enable different types of bucket logging

achouhan09:
### Explain the changes
1.  Added support to enable different types of bucket logging, i.e, best-effort and guaranteed.


pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -237,6 +244,20 @@ type AutoscalerSpec struct {
...
+	// There are two types available: guaranteed and non-guaranteed logging.
+	// - guaranteed logging ensures that all log entries are recorded without fail.
+	// - non-guaranteed logging may not record all log entries
+	// By default, guaranteed logging is enabled if no type is specified.
+	// +optional
+	LoggingType BucketLoggingTypes `json:"loggingType,omitempty"`
+
+	// TODO
+	// +optional
+	BucketClass string `json:"bucketclass,omitempty"`
```
jackyalbo, line 258: I guess that's should be storageclass right? it's the type of storageclass that we will use to create a PVC on top of in case of using guaranteed logging.

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -237,6 +244,20 @@ type AutoscalerSpec struct {
 	PrometheusNamespace string `json:"prometheusNamespace,omitempty"`
 }
 
+type BucketLoggingSpec struct {
+	// LoggingType specifies the type of logging for the bucket.
+	// There are two types available: guaranteed and non-guaranteed logging.
+	// - guaranteed logging ensures that all log entries are recorded without fail.
+	// - non-guaranteed logging may not record all log entries
+	// By default, guaranteed logging is enabled if no type is specified.
```
jackyalbo, line 252: At least for now, non-guaranteed will be the default.

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -539,3 +560,14 @@ const (
 	// AutoscalerTypeHPAV2 is hpav2
 	AutoscalerTypeHPAV2 AutoscalerTypes = "hpav2"
 )
+
+type BucketLoggingTypes string
```
jackyalbo, line 564: Add a comment explaining what this type is used for

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -223,6 +223,13 @@ type NooBaaSpec struct {
 	// DenyHTTP (optional) if given will deny access to the NooBaa S3 service using HTTP (only HTTPS)
 	// +optional
 	DenyHTTP bool `json:"denyHTTP,omitempty"`
+
+	// BucketLogging sets the configuration for bucket logging.
```
jackyalbo, line 227: let's keep this a one-liner. extra info will come from the properties' comments.

pkg/apis/noobaa/v1alpha1/noobaa_types.go
achouhan09: Got it.

pkg/apis/noobaa/v1alpha1/noobaa_types.go
achouhan09: Added, Thanks.

pkg/apis/noobaa/v1alpha1/noobaa_types.go
achouhan09: Updated, Thanks.

pkg/bundle/deploy.go
```diff
@@ -2463,6 +2463,23 @@ spec:
...
+                      BucketLoggingStorageClass (otional) specifies the type of storage class that will be used to create a
+                      PVC(Persistent Volume Claim) for guaranteed logging, if enabled.
+                    type: string
+                  loggingType:
+                    description: |-
+                      LoggingType specifies the type of logging for the bucket
+                      There are two types available: guaranteed and non-guaranteed logging
+                      - guaranteed logging ensures that all log entries are recorded without fail
+                      - non-guaranteed logging may not record all log entries
+                      By default, non-guaranteed logging is enabled if no type is specified
```
jackyalbo, line 2480: I want to change non-guaranteed to best-effort, as I don't want it to sound like something you shouldn't use.
please change in all the needed places. and about the description, I suggest something like this:
```suggestion
                    description: |-
                      LoggingType specifies the type of logging for the bucket
                      There are two types available: best-effort and guaranteed logging
                      - best-effort(default) - less immune to failures but with better performance
                      - guaranteed - much more reliable but need to provide a storage class that supports RWX PVs
```

achouhan09: sure, thanks.

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -237,6 +241,21 @@ type AutoscalerSpec struct {
 	PrometheusNamespace string `json:"prometheusNamespace,omitempty"`
 }
 
+// BucketLoggingSpec defines the configuration for the bucket
```
jackyalbo, line 244: change here to bucket logging configuration
```suggestion
// BucketLoggingSpec defines the bucket logging configuration
```

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -237,6 +241,21 @@ type AutoscalerSpec struct {
...
+type BucketLoggingSpec struct {
+	// LoggingType specifies the type of logging for the bucket
+	// There are two types available: best-effort and guaranteed logging
+	// - best-effort(default) - less immune to failures but with better performance
+	// - guaranteed - much more reliable but need to provide a storage class that supports RWX PVs
+	// +optional
+	LoggingType BucketLoggingTypes `json:"loggingType,omitempty"`
+
+	// BucketLoggingStorageClass (otional) specifies the type of storage class that will be used to create a
```
jackyalbo, line 253:
```suggestion
	// BucketLoggingStorageClass (optional) specifies the type of storage class that will be used to create a
```

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -539,3 +558,15 @@ const (
 	// AutoscalerTypeHPAV2 is hpav2
 	AutoscalerTypeHPAV2 AutoscalerTypes = "hpav2"
 )
+
+// BucketLoggingTypes is a string enum type for specifying the types of bucketlogging supported.
+type BucketLoggingTypes string
+
+// These are the valid BucketLoggingTypes types:
+const (
+	// BucketLoggingTypeNonGuaranteed is best-effort
```
jackyalbo, line 567:
```suggestion
	// BucketLoggingTypeBestEffort is best-effort
```

pkg/apis/noobaa/v1alpha1/noobaa_types.go
achouhan09: Okay, Thanks

jackyalbo (approved):

PR #1366, opened 2024-05-23
Nsfs : Default resources on hci odf

aspandey:
NSFS: Create default resource objects on Fusion HCI
    
    ODF MCG on Fusion HCI with Spectrum Scale CNSA should
    configure the default BucketClass, Namespacestore and
    default PVC to use NSFS.
    
    To set up NSFS on kubernetes, we need to create pvc
    and namespacestore which can be used to create nsfs
    bucket class. These nsfs buckets are used to create
    a user bucket by s3 client.

### DIFF
--- a/deploy/internal/nsfs-pvc-cr.yaml
+++ b/deploy/internal/nsfs-pvc-cr.yaml
@@ -1,0 +1,10 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: noobaa-default-nsfs-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 30Gi
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -801,7 +801,6 @@ func (r *Reconciler) ReconcileIBMCredentials() error {
 	return nil
 }
 
-
 // SetDesiredAgentProfile updates the value of the AGENT_PROFILE env
 func (r *Reconciler) SetDesiredAgentProfile(profileString string) string {
 	agentProfile := map[string]interface{}{}
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -79,6 +79,12 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 	if err := r.ReconcileDefaultBackingStore(); err != nil {
 		return err
 	}
+	if err := r.ReconcileDefaultNsfsPvc(); err != nil {
+		return err
+	}
+	if err := r.ReconcileDefaultNamespaceStore(); err != nil {
+		return err
+	}
 	if err := r.ReconcileDefaultBucketClass(); err != nil {
 		return err
 	}
@@ -682,6 +688,87 @@ func (r *Reconciler) RegisterToCluster() error {
 	return r.NBClient.RegisterToCluster()
 }
 
+// ReconcileDefaultNamespaceStore checks if the default NSFS pvc exists or not
+// and attempts to create default NSFS namespacestore using NSFS pvc which in turn uses
+// spectrum scale storage class.
+func (r *Reconciler) ReconcileDefaultNamespaceStore() error {
+	// Skip if joining another NooBaa
+	if r.JoinSecret != nil {
+		return nil
+	}
+
+	log := r.Logger.WithField("func", "ReconcileDefaultNamespaceStore")
+
+	if r.DefaultNsfsPvc.UID == "" {
+		log.Infof("PVC %s does not  exist. skipping Reconcile %s", r.DefaultNsfsPvc.Name, r.DefaultNamespaceStore.Name)
+		return nil
+	}
+
+	util.KubeCheck(r.DefaultNamespaceStore)
+
+	if r.DefaultNamespaceStore.UID != "" {
+		log.Infof("NamespaceStore %s already exists. skipping Reconcile", r.DefaultNamespaceStore.Name)
+		return nil
+	}
+
+	r.DefaultNamespaceStore.Spec.Type = nbv1.NSStoreTypeNSFS
+	r.DefaultNamespaceStore.Spec.NSFS = &nbv1.NSFSSpec{}
+	r.DefaultNamespaceStore.Spec.NSFS.PvcName = r.DefaultNsfsPvc.Name
+	r.DefaultNamespaceStore.Spec.NSFS.SubPath = ""
+
+	r.Own(r.DefaultNamespaceStore)
+
+	if err := r.Client.Create(r.Ctx, r.DefaultNamespaceStore); err != nil {
+		log.Errorf("got error on DefaultNamespaceStore creation. error: %v", err)
+		return err
+	}
+	return nil
+}
+
+// ReconcileDefaultNsfsPvc checks if the noobaa is running on Fusion HCI with
+// spectrum scale and attempts to create default PVC for nsfs using spectrum scale
+// storage class.
+func (r *Reconciler) ReconcileDefaultNsfsPvc() error {
+	// Skip if joining another NooBaa
+	if r.JoinSecret != nil {
+		return nil
+	}
+
+	// Check if ODF is installed on Fusion HCI cluster with spectrum scale.
+	if !util.IsFusionHCIWithScale() {
+		return nil
+	}
+	r.Logger.Info("IBM Fusion HCI with Spectrum Scale detected.")
+	log := r.Logger.WithField("func", "ReconcileDefaultNsfsPvc")
+
+	util.KubeCheck(r.DefaultNsfsPvc)
+
+	if r.DefaultNsfsPvc.UID != "" {
+		log.Infof("DefaultNsfsPvc %s already exists. skipping Reconcile", r.DefaultNsfsPvc.Name)
+		return nil
+	}
+
+	if r.NooBaa.Spec.ManualDefaultBackingStore {
+		r.Logger.Info("ManualDefaultBackingStore is true, Skip Reconciling default nsfs pvc")
+		return nil
+	}
+
+	var sc = "ibm-spectrum-scale-csi-storageclass-version2"
+	defaultPVCSize := int64(30) * 1024 * 1024 * 1024 // 30GB
+	r.DefaultNsfsPvc.Spec.AccessModes = []corev1.PersistentVolumeAccessMode{corev1.ReadWriteMany}
+	r.DefaultNsfsPvc.Spec.Resources.Requests.Storage()
+	r.DefaultNsfsPvc.Spec.Resources.Requests[corev1.ResourceStorage] = *resource.NewQuantity(defaultPVCSize, resource.BinarySI)
+	r.DefaultNsfsPvc.Spec.StorageClassName = &sc
+
+	r.Own(r.DefaultNsfsPvc)
+
+	if err := r.Client.Create(r.Ctx, r.DefaultNsfsPvc); err != nil {
+		log.Errorf("got error on DefaultNsfsPvc creation. error: %v", err)
+		return err
+	}
+	return nil
+}
+
 // ReconcileDefaultBackingStore attempts to get credentials to cloud storage using the cloud-credentials operator
 // and use it for the default backing store
 func (r *Reconciler) ReconcileDefaultBackingStore() error {
@@ -691,6 +778,12 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
 	}
 
 	log := r.Logger.WithField("func", "ReconcileDefaultBackingStore")
+
+	// Check if ODF is installed on Fusion HCI cluster with spectrum scale.
+	if util.IsFusionHCIWithScale() {
+		r.Logger.Info("IBM Fusion HCI with Spectrum Scale detected. Not creating Default Backing Store.")
+		return nil
+	}
 
 	util.KubeCheck(r.DefaultBackingStore)
 	// backing store already exists - we can skip
@@ -1281,12 +1374,21 @@ func (r *Reconciler) ReconcileDefaultBucketClass() error {
 		return nil
 	}
 
-	r.DefaultBucketClass.Spec.PlacementPolicy = &nbv1.PlacementPolicy{
-		Tiers: []nbv1.Tier{{
-			BackingStores: []nbv1.BackingStoreName{
-				r.DefaultBackingStore.Name,
+	if util.KubeCheck(r.DefaultNamespaceStore) {
+		r.DefaultBucketClass.Spec.NamespacePolicy = &nbv1.NamespacePolicy{
+			Type: nbv1.NSBucketClassTypeSingle,
+			Single: &nbv1.SingleNamespacePolicy{
+				Resource: r.DefaultNamespaceStore.Name,
 			},
-		}},
+		}
+	} else {
+		r.DefaultBucketClass.Spec.PlacementPolicy = &nbv1.PlacementPolicy{
+			Tiers: []nbv1.Tier{{
+				BackingStores: []nbv1.BackingStoreName{
+					r.DefaultBackingStore.Name,
+				},
+			}},
+		}
 	}
 
 	r.Own(r.DefaultBucketClass)
--- a/pkg/system/reconciler.go
+++ b/pkg/system/reconciler.go
@@ -98,6 +98,8 @@ type Reconciler struct {
 	IBMCosBucketCreds         *corev1.Secret
 	DefaultBackingStore       *nbv1.BackingStore
 	DefaultBucketClass        *nbv1.BucketClass
+	DefaultNamespaceStore     *nbv1.NamespaceStore
+	DefaultNsfsPvc            *corev1.PersistentVolumeClaim
 	OBCStorageClass           *storagev1.StorageClass
 	PrometheusRule            *monitoringv1.PrometheusRule
 	ServiceMonitorMgmt        *monitoringv1.ServiceMonitor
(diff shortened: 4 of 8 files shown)


pkg/system/phase2_creating.go
```diff
@@ -594,6 +594,11 @@ func (r *Reconciler) ReconcileBackingStoreCredentials() error {
 	if util.IsAWSPlatform() {
 		return r.ReconcileAWSCredentials()
 	}
+
+	if util.IsHCIPlatform() {
+		// return r.ReconcileAzureCredentials()
```
jackyalbo, line 599: remove this comment. Why do we need this part? just for this print? debugging?

pkg/system/phase4_configuring.go
```diff
@@ -747,6 +791,22 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
 	return nil
 }
 
+func (r *Reconciler) prepareDefaultPVC() error {
+	// We need to create a PVC with specturm scale storageclass. nsfs-local is just for testing.
```
jackyalbo, line 795: I guess we should fix this before merge?

pkg/util/util.go
```diff
@@ -1031,6 +1031,14 @@ func IsAWSPlatform() bool {
 	return isAWS
 }
 
+func IsHCIPlatform() bool {
```
jackyalbo, line 1034: Add comment

policy.json
```diff
@@ -0,0 +1,12 @@
+{
```
jackyalbo, line 1: should be removed?

pkg/system/phase4_configuring.go
```diff
@@ -682,6 +685,47 @@ func (r *Reconciler) RegisterToCluster() error {
...
+	// Skip if joining another NooBaa
+	if r.JoinSecret != nil {
+		return nil
+	}
+
+	// Check if ODF is installed on Fusion HCI cluster with spectrum scale.
+	if !util.IsHCIPlatform() {
+		r.Logger.Info(" IBM Fusion HCI with spectrum scale NOT detected.")
+		return nil
+	} else {
```
jackyalbo, line 698: no need for the else

deploy/crds/noobaa.io_v1alpha1_pvc_cr.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
jackyalbo, line 1: move this file to internal, it's not a CRD we are creating: deploy/internal/pvc-nsfs.yaml, or something similar.

pkg/system/phase4_configuring.go
```diff
@@ -94,6 +94,9 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 	if err := r.ReconcileReadSystem(); err != nil {
 		return err
 	}
+	if err := r.ReconcileDefaultNamespaceStore(); err != nil {
```
jackyalbo, line 97: move this before the ReconcileDefaultBucketClass. also, you are missing on the code that should put the default namespace store inside of the default bucket class.

jackyalbo (changes requested):

pkg/system/phase4_configuring.go
```diff
@@ -682,6 +685,47 @@ func (r *Reconciler) RegisterToCluster() error {
 	return r.NBClient.RegisterToCluster()
 }
 
+func (r *Reconciler) ReconcileDefaultNamespaceStore() error {
```
achouhan09, line 688: should require description comment to be exported

pkg/system/system.go
```diff
@@ -205,19 +205,19 @@ func LoadSystemDefaults() *nbv1.NooBaa {
 	}
 	if options.MiniEnv {
 		coreResourceList := corev1.ResourceList{
-			corev1.ResourceCPU:    *resource.NewScaledQuantity(int64(100), resource.Milli),
+			corev1.ResourceCPU:    *resource.NewScaledQuantity(int64(500), resource.Milli),
```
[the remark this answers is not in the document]
aspandey: No, It was not needed. I do these changes for my local setup and by mistake checked-in these files also.

pkg/system/phase2_creating.go
aspandey: Done. Removed.

pkg/system/phase4_configuring.go
aspandey: Yes, that is the plan. Will change it.

pkg/system/phase4_configuring.go
aspandey: Yes, will do that

pkg/system/reconciler.go
```diff
@@ -98,6 +98,8 @@ type Reconciler struct {
 	IBMCosBucketCreds         *corev1.Secret
 	DefaultBackingStore       *nbv1.BackingStore
 	DefaultBucketClass        *nbv1.BucketClass
+	DefaultNamespaceStore     *nbv1.NamespaceStore
+	DefaultPVC                *corev1.PersistentVolumeClaim
```
guymguym, line 102: maybe the name should be DefaultNsfsPvc because default pvc is a bit vague in the operator context

pkg/util/util.go
guymguym: hci is a generic term, better rename to something specific like IsFusionHCIWithScale

aspandey: Make sense.

policy.json
aspandey: Agree.

pkg/system/phase4_configuring.go
aspandey: Yes.

deploy/crds/noobaa.io_v1alpha1_pvc_cr.yaml
aspandey: done

pkg/system/phase4_configuring.go
aspandey: done

pkg/system/phase4_configuring.go
```diff
@@ -1281,12 +1341,23 @@ func (r *Reconciler) ReconcileDefaultBucketClass() error {
 		return nil
 	}
 
-	r.DefaultBucketClass.Spec.PlacementPolicy = &nbv1.PlacementPolicy{
-		Tiers: []nbv1.Tier{{
-			BackingStores: []nbv1.BackingStoreName{
-				r.DefaultBackingStore.Name,
+	if !util.IsFusionHCIWithScale() {
```
jackyalbo, line 1344: I prefer to check if we have DefaultBackingStore or if we have DefaultNamespaceStore, as in the future we might support more default namespacestores...

pkg/system/phase4_configuring.go
```diff
@@ -1281,12 +1341,23 @@ func (r *Reconciler) ReconcileDefaultBucketClass() error {
...
+		r.DefaultBucketClass.Spec.PlacementPolicy = &nbv1.PlacementPolicy{
+			Tiers: []nbv1.Tier{{
+				BackingStores: []nbv1.BackingStoreName{
+					r.DefaultBackingStore.Name,
+				},
+			}},
+		}
+	} else {
+		r.Logger.Info("IBM Fusion HCI with spectrum scale detected.")
+		r.DefaultBucketClass.Name = "noobaa-default-nsfs-bucket-class"
```
jackyalbo, line 1354: I prefer keeping the same name, as we will probably have other default namespace stores.

pkg/system/phase4_configuring.go
```diff
@@ -747,6 +790,23 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
 	return nil
 }
 
+func (r *Reconciler) prepareDefaultNsfsPvc() error {
+	// We need to create a PVC with specturm scale storageclass. nsfs-local-test is just for testing.
+	var sc = "ibm-spectrum-scale-csi-storageclass-version2"
+	defaultPVCSize := int64(16) * 1024 * 1024 * 1024 // 16GB
```
jackyalbo, line 796: 16GB is too low, Let's use 30GB like we did for pv-pool.

deploy/internal/nsfs-pvc-cr.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: default
```
jackyalbo, line 4: I agree with @achouhan09 on the name. let's set here the name we intend to use later: `noobaa-default-nsfs-pvc`

deploy/internal/nsfs-pvc-cr.yaml
```diff
@@ -0,0 +1,11 @@
...
+spec:
+  storageClassName: default
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 1Gi
```
jackyalbo, line 11: also here, write the size that we intend to use - 30GB. and add a new line in the end.

deploy/internal/nsfs-pvc-cr.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: default
+spec:
+  storageClassName: default
```
jackyalbo, line 6: we can remove this

aspandey: Just to be clear - You are suggesting this yaml to be look like this - 

apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: noobaa-default-nsfs-pvc
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 30Gi

pkg/system/phase4_configuring.go
jackyalbo: yes. We don't need to handle HCI separately in bucket class.
in bucket class we should see: if we have default backingstore set -> it should point to it.
if not, if we have default namespacestore set -> it should point to it.
the logic on what to create was already handled in the default namepsace/backingstore logic.

deploy/internal/nsfs-pvc-cr.yaml
aspandey: done.

deploy/internal/nsfs-pvc-cr.yaml
aspandey: done.

deploy/internal/nsfs-pvc-cr.yaml
aspandey: done.

pkg/system/phase4_configuring.go
aspandey: done.

pkg/system/phase4_configuring.go
aspandey: done.

pkg/system/phase4_configuring.go
aspandey: done.

pkg/system/phase4_configuring.go
```diff
@@ -747,6 +793,24 @@ func (r *Reconciler) ReconcileDefaultBackingStore() error {
 	return nil
 }
 
+// prepareDefaultNsfsPvc creates a PVC with specturm scale storageclass
+// and configure DefaultNamespaceStore.
+func (r *Reconciler) prepareDefaultNsfsPvc() error {
+	var sc = "ibm-spectrum-scale-csi-storageclass-version2"
+	defaultPVCSize := int64(30) * 1024 * 1024 * 1024 // 16GB
```
jackyalbo, line 800: fix comment
```suggestion
	defaultPVCSize := int64(30) * 1024 * 1024 * 1024 // 30GB
```

pkg/system/phase4_configuring.go
```diff
@@ -1281,12 +1345,22 @@ func (r *Reconciler) ReconcileDefaultBucketClass() error {
...
+	if util.KubeCheck(r.DefaultBackingStore) {
+		r.DefaultBucketClass.Spec.PlacementPolicy = &nbv1.PlacementPolicy{
+			Tiers: []nbv1.Tier{{
+				BackingStores: []nbv1.BackingStoreName{
+					r.DefaultBackingStore.Name,
+				},
+			}},
+		}
+	}
+	if util.KubeCheck(r.DefaultNamespaceStore) {
```
jackyalbo, line 1357: add else - no need to check both (correct me if I'm wrong...)

jackyalbo (commented):
just 2 minors to fix, but looks good

pkg/system/phase4_configuring.go
aspandey: Done.

pkg/system/phase4_configuring.go
aspandey: Done.

jackyalbo (approved):

pkg/system/phase4_configuring.go
```diff
@@ -682,6 +685,49 @@ func (r *Reconciler) RegisterToCluster() error {
...
+// storage class.
+func (r *Reconciler) ReconcileDefaultNamespaceStore() error {
+	// Skip if joining another NooBaa
+	if r.JoinSecret != nil {
+		return nil
+	}
+
+	// Check if ODF is installed on Fusion HCI cluster with spectrum scale.
+	if !util.IsFusionHCIWithScale() {
+		r.Logger.Info(" IBM Fusion HCI with spectrum scale NOT detected.")
```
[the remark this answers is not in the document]
aspandey: This was done on purpose to highlight "NOT"

pkg/system/phase4_configuring.go
```diff
@@ -682,6 +685,49 @@ func (r *Reconciler) RegisterToCluster() error {
...
+		return nil
+	}
+
+	log := r.Logger.WithField("func", "ReconcileDefaultNamespaceStore")
+
+	util.KubeCheck(r.DefaultNamespaceStore)
+	// Namespacestore already exist
+
+	if r.DefaultNamespaceStore.UID != "" {
```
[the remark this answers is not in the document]
aspandey: I think if we have DefaultNamespaceStore then that means we have DefaultNsfsPvc here.
What do you think @jackyalbo ?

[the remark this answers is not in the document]
jackyalbo: Can you have a small test checking what happens if the pvc is already created. if it failes the create then we should check, as it will stop us from creating the backingstore. if not, leave it like this.
another thing I can see missing also is this check: `if r.NooBaa.Spec.ManualDefaultBackingStore`
if user ask us not to create defaultBackingStore so we should also skip the defaultNamespaceStore.

pkg/system/phase4_configuring.go
```diff
@@ -682,6 +685,49 @@ func (r *Reconciler) RegisterToCluster() error {
...
+
+	if err := r.prepareDefaultNsfsPvc(); err != nil {
+		return err
+	}
+	r.Own(r.DefaultNsfsPvc)
+	r.Own(r.DefaultNamespaceStore)
+
+	if err := r.Client.Create(r.Ctx, r.DefaultNsfsPvc); err != nil {
+		log.Errorf("got error on DefaultNsfsPvc creation. error: %v", err)
+		// return err
```
jackyalbo, line 721: why is this in comment? is it related to cases we already have the PVC?

pkg/system/phase4_configuring.go
```diff
@@ -682,6 +688,97 @@ func (r *Reconciler) RegisterToCluster() error {
 	return r.NBClient.RegisterToCluster()
 }
 
+// ReconcileDefaultNamespaceStore checks if the default nsfs pvc exist or not
+// and attempts to create default nsfsnamespacestore using nsfs pvc which inturn uses
+// spectrum scale storage class.
```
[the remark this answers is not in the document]
aspandey: OKay.

pkg/system/phase4_configuring.go
```diff
@@ -682,6 +688,97 @@ func (r *Reconciler) RegisterToCluster() error {
...
+// storage class.
+func (r *Reconciler) ReconcileDefaultNsfsPvc() error {
+	// Skip if joining another NooBaa
+	if r.JoinSecret != nil {
+		return nil
+	}
+
+	// Check if ODF is installed on Fusion HCI cluster with spectrum scale.
+	if !util.IsFusionHCIWithScale() {
+		r.Logger.Info(" IBM Fusion HCI with spectrum scale NOT detected.")
```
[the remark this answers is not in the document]
aspandey: You are right. We should not print this log for every system which is not HCI. 
However, I would like this message to be printed, mainly because we have not checked this code on HCI.
Now, I am logging it only when we detect HCI. That will help us in debugging also and at the same time it will not print it for every other platform.

pkg/system/phase4_configuring.go
```diff
@@ -682,6 +688,97 @@ func (r *Reconciler) RegisterToCluster() error {
...
+	util.KubeCheck(r.DefaultNsfsPvc)
+	// DefaultNsfsPvc already exist
+
+	if r.DefaultNsfsPvc.UID != "" {
+		log.Infof("DefaultNsfsPvc %s already exists. skipping Reconcile", r.DefaultNsfsPvc.Name)
+		return nil
+	}
+
+	// If Manual default backing store is enable, skip creating default nsfs pvc
+	if r.NooBaa.Spec.ManualDefaultBackingStore {
```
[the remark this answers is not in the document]
aspandey: For this and the previous comment on "ManualDefaultBackingStore", @jackyalbo  and I discussed it over call and we also thought that the same flag can be used. It does look like misleading but creating flag could be more challenging as you have also mentioned.

pkg/system/phase4_configuring.go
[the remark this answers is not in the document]
aspandey: I think we can remove following check from Namespacestore reconciler.
if r.NooBaa.Spec.ManualDefaultBackingStore

If it is true default NSFS PVC will not be created and that mean we will not reconcile Namespacestore.

aspandey:
> I might be missing it, are we skipping creating a default backingstore when we create NSFS namespacestore (and vice versa)?

NO, its not like that. Both will be created.
In ReconcileDefaultBucketClass only, we are creating Bucketclass with either one of backingstore or namespacestore.

pkg/system/phase4_configuring.go
aspandey: done

liranmauda:
> > I might be missing it, are we skipping creating a default backingstore when we create NSFS namespacestore (and vice versa)?
> 
> NO, its not like that. Both will be created. In ReconcileDefaultBucketClass only, we are creating Bucketclass with either one of backingstore or namespacestore.

If a default namespacestore is being created we should not create a default backingstore.

aspandey:
> > > I might be missing it, are we skipping creating a default backingstore when we create NSFS namespacestore (and vice versa)?
> > 
> > 
> > NO, its not like that. Both will be created. In ReconcileDefaultBucketClass only, we are creating Bucketclass with either one of backingstore or namespacestore.
> 
> If a default namespacestore is being created we should not create a default backingstore.

Made changes. Now checking if platform is IsFusionHCIWithScale or not and skipping the defaultbackingstore creation.

liranmauda (approved):

deploy/internal/nsfs-pvc-cr.yaml
```diff
@@ -0,0 +1,10 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: noobaa-default-nsfs-pvc
+spec:
+  accessModes:
+    - ReadWriteOnce
```
guymguym, line 7: @aspandey The default access mode for nsfs PVs better be ReadWriteMany

PR #1368, opened 2024-05-28
Validations for BucketLoggingPVC in case of guaranteed bucket logging

achouhan09:
### Explain the changes
1.  Added validation checks for `BucketLoggingPVC` to ensure it is specified for 'guaranteed' logging and has ReadWriteMany (RWX) access. Also, in case of ODF we will create a new PVC if not provided by the user.

### DIFF
--- a/pkg/system/phase1_verifying.go
+++ b/pkg/system/phase1_verifying.go
@@ -7,6 +7,9 @@ import (
 
 	// this is the driver we are using for psql
 	_ "github.com/lib/pq"
+	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"github.com/asaskevich/govalidator"
 	semver "github.com/coreos/go-semver/semver"
@@ -59,6 +62,12 @@ func (r *Reconciler) ReconcilePhaseVerifying() error {
 			os.Setenv("PGSSLCERT", "/tmp/tls.crt")
 		}
 		if err := r.checkExternalPg(r.ExternalPgSecret.StringData["db_url"]); err != nil {
+			return err
+		}
+	}
+
+	if r.NooBaa.Spec.BucketLogging.LoggingType == nbv1.BucketLoggingTypeGuaranteed {
+		if err := r.checkBucketLoggingPVC(); err != nil {
 			return err
 		}
 	}
@@ -251,4 +260,40 @@ func (r *Reconciler) checkExternalPg(postgresDbURL string) error {
 				dbURL, collation))
 	}
 	return nil
+}
+
+// checkBucketLoggingPVC validates the configuration of bucket logging pvc
+func (r *Reconciler) checkBucketLoggingPVC() error {
+	// Rejecting if 'BucketLoggingPVC' is not provided for 'guaranteed' logging and
+	// also the operator is not running in the ODF environment.
+	sc := &storagev1.StorageClass{
+		TypeMeta:   metav1.TypeMeta{Kind: "StorageClass"},
+		ObjectMeta: metav1.ObjectMeta{Name: "ocs-storagecluster-cephfs"},
+	}
+	if r.NooBaa.Spec.BucketLogging.BucketLoggingPVC == nil && !util.KubeCheck(sc) {
+		return util.NewPersistentError("InvalidBucketLoggingConfiguration",
+			"'Guaranteed' BucketLogging requires a Persistent Volume Claim (PVC) with ReadWriteMany (RWX) access mode. Please specify the 'BucketLoggingPVC' to ensure guaranteed logging")
+	}
+
+	// Check if pvc exists in the cluster
+	BucketLoggingPVC := &corev1.PersistentVolumeClaim{
+		TypeMeta: metav1.TypeMeta{Kind: "PersistenVolumeClaim"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      *r.NooBaa.Spec.BucketLogging.BucketLoggingPVC,
+			Namespace: r.Request.Namespace,
+		},
+	}
+	if !util.KubeCheck(BucketLoggingPVC) {
+		return util.NewPersistentError("InvalidBucketLoggingConfiguration",
+			"The specified 'BucketLoggingPVC' was not found")
+	}
+
+	// Check if pvc supports RWX access mode
+	for _, accessMode := range BucketLoggingPVC.Spec.AccessModes {
+		if accessMode == corev1.ReadWriteMany {
+			return nil
+		}
+	}
+	return util.NewPersistentError("InvalidBucketLoggingConfiguration",
+		"The specified 'BucketLoggingPVC' does not support RWX access mode")
 }
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -128,6 +128,12 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 			return err
 		}
 	}
+	// create bucket logging pvc if not provided by user for 'Guaranteed' logging in ODF env
+	if r.NooBaa.Spec.BucketLogging.LoggingType == nbv1.BucketLoggingTypeGuaranteed {
+		if err := r.ReconcileODFBucketLoggingPVC(); err != nil {
+			return err
+		}
+	}
 	if err := r.ReconcileObject(r.ServiceMgmt, r.SetDesiredServiceMgmt); err != nil {
 		return err
 	}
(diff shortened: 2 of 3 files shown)


pkg/system/phase1_verifying.go
```diff
@@ -171,6 +173,18 @@ func (r *Reconciler) CheckSystemCR() error {
 			fmt.Sprintf("Autoscaler %s missing prometheusNamespace property ", r.NooBaa.Spec.Autoscaler.AutoscalerType))
 	}
 
+	// Rejecting NooBaa CR if 'BucketLoggingStorageClass' is not provided for 'guaranteed' logging and
+	// also the operator is not running in an ODF environment.
+	sc := &storagev1.StorageClass{
+		TypeMeta:   metav1.TypeMeta{Kind: "StorageClass"},
+		ObjectMeta: metav1.ObjectMeta{Name: "ocs-storagecluster-cephfs"},
+	}
+	if !util.KubeCheck(sc) && (r.NooBaa.Spec.BucketLogging.LoggingType == nbv1.BucketLoggingTypeGuaranteed) &&
```
jackyalbo, line 182: Any reason for the brackets? 
```suggestion
	if !util.KubeCheck(sc) && r.NooBaa.Spec.BucketLogging.LoggingType == nbv1.BucketLoggingTypeGuaranteed &&
```

pkg/system/phase4_configuring.go
```diff
@@ -1323,6 +1327,49 @@ func (r *Reconciler) ReconcileOBCStorageClass() error {
...
+
+	if r.NooBaa.Spec.BucketLogging.BucketLoggingStorageClass == nil {
+		sc := &storagev1.StorageClass{
+			TypeMeta:   metav1.TypeMeta{Kind: "StorageClass"},
+			ObjectMeta: metav1.ObjectMeta{Name: "ocs-storagecluster-cephfs"},
+		}
+
+		// fallback to cephfs storageclass if running on odf and BucketLoggingStorageClass not provided
+		if util.KubeCheck(sc) {
+			r.BucketLoggingPVC.Spec.StorageClassName = &sc.Name
```
jackyalbo, line 1363: Please add a warning here to the logs that we are using ceph-fs. Something like: "BucketLoggingStorageClass wasn't provided - defaulting to ceph-fs storage class"

pkg/system/phase4_configuring.go
```diff
@@ -1323,6 +1327,49 @@ func (r *Reconciler) ReconcileOBCStorageClass() error {
...
+		} else {
+			return util.NewPersistentError("InvalidBucketLoggingConfiguration",
+				"'guaranteed' BucketLogging requires a storageclass. Please specify BucketLoggingStorageClass to enable PVC creation.")
```
jackyalbo, line 1366:
```suggestion
				"Guaranteed BucketLogging requires a storageclass. Please specify BucketLoggingStorageClass to enable PVC creation.")
```

jackyalbo (commented):
Very small minor requests. Looks good in general.

pkg/system/phase4_configuring.go
```diff
@@ -1323,6 +1327,49 @@ func (r *Reconciler) ReconcileOBCStorageClass() error {
 	return nil
 }
 
+// ReconcileBucketLoggingPVC ensures the bucket logging PVC is properly configured for 'guaranteed' logging
+func (r *Reconciler) ReconcileBucketLoggingPVC() error {
+	// Return if logging is not enabled or if 'guaranteed' logging is not specified
+	if r.NooBaa.Spec.BucketLogging.LoggingType != nbv1.BucketLoggingTypeGuaranteed {
```
aspandey, line 1333: IN the comment you are saying "check if the logging is not enable". But are you checking this condition?
if bucketlogging is not enabled , probably r.NooBaa.Spec.BucketLogging will be NULL? and if it is then the code will crash.

pkg/system/phase4_configuring.go
```diff
@@ -1323,6 +1327,49 @@ func (r *Reconciler) ReconcileOBCStorageClass() error {
...
+
+	return nil
+}
+
+// prepareBucketLoggingPVC configures the bucket logging pvc
+func (r *Reconciler) prepareBucketLoggingPVC() error {
+	r.BucketLoggingPVC.Spec.AccessModes = []corev1.PersistentVolumeAccessMode{corev1.ReadWriteMany}
+	r.BucketLoggingPVC.Spec.StorageClassName = r.NooBaa.Spec.BucketLogging.BucketLoggingStorageClass
+
+	if r.NooBaa.Spec.BucketLogging.BucketLoggingStorageClass == nil {
```
aspandey, line 1355: assignment -
r.BucketLoggingPVC.Spec.StorageClassName = r.NooBaa.Spec.BucketLogging.BucketLoggingStorageClass

should be in the else part of this "if" condition.

pkg/system/phase1_verifying.go
aspandey: To me it make more sense to first check if we have Guaranteedlogging or not and then check for storage classes - 
if (r.NooBaa.Spec.BucketLogging.LoggingType == nbv1.BucketLoggingTypeGuaranteed)  {

        sc := &storagev1.StorageClass{
		TypeMeta:   metav1.TypeMeta{Kind: "StorageClass"},
		ObjectMeta: metav1.ObjectMeta{Name: "ocs-storagecluster-cephfs"},
	}
         if !util.KubeCheck(sc) && 
                  r.NooBaa.Spec.BucketLogging.BucketLoggingStorageClass == nil

-------
-------
}

Point is, why even  bother about storageclass if we don't need at all at first place.

achouhan09: Yes Jacky, we can remove the brackets. Thanks Ashish, I am also thinking of the same, will update it.

pkg/system/phase4_configuring.go
achouhan09: Sure, I will add. Thanks

pkg/system/phase4_configuring.go
achouhan09: @aspandey here `BucketLogging` is not of pointer type we cannot check for nil value. And as it is not of pointer type so every field contains zero value in case of `r.NooBaa.Spec.BucketLogging.LoggingType` if user doesn't provide any value then it contains empty string(zero value of string). So this way it is satisfying both the conditions either it is not defined or not of type `Guaranteed`.

cc @jackyalbo  wdyt?

pkg/system/phase4_configuring.go
```diff
@@ -1338,19 +1338,27 @@ func (r *Reconciler) ReconcileBucketLoggingPVC() error {
...
 	r.Own(r.BucketLoggingPVC)
-	// TODO: return only for the specific error, like RWX mode not supported by provisioner
-	if err := r.Client.Create(r.Ctx, r.BucketLoggingPVC); err != nil {
-		return err
+
+	if r.BucketLoggingPVC.UID != "" && r.BucketLoggingPVC.Status.Phase == corev1.ClaimPending {
+		log.Infof("BucketLoggingPVC %s is in pending state. Investigating the reason and updating pvc", r.BucketLoggingPVC.Name)
+		// TODO: fetch reason and return only for the specific error, like RWX mode not supported by provisioner
+
+		log.Infof("Deleting and recreating the pending BucketLoggingPVC %s", r.BucketLoggingPVC.Name)
```
aspandey, line 1350: Why do we have two log info back to back? One is asking for investigating the reason and other is saying deleting and recreating?

achouhan09: Actually, as you can see the comment we need to fetch events on pvc here and we should return an error if error related to `RWX` mode not supported.

aspandey:
looks good to me.

pkg/system/phase4_configuring.go
```diff
@@ -1323,6 +1327,82 @@ func (r *Reconciler) ReconcileOBCStorageClass() error {
...
+			// 	return err
+			// }
+
+			log.Infof("Deleting and recreating the pending BucketLoggingPVC %s", r.BucketLoggingPVC.Name)
+			if err := r.Client.Delete(r.Ctx, r.BucketLoggingPVC); err != nil {
+				return err
+			}
+			if err := r.Client.Create(r.Ctx, r.BucketLoggingPVC); err != nil {
+				return err
+			}
```
liranmauda, line 1366: Could you explain this? 
Can't a PVC be pending and have data on it?

pkg/system/phase4_configuring.go
```diff
@@ -1323,6 +1327,82 @@ func (r *Reconciler) ReconcileOBCStorageClass() error {
...
+			// TODO: fetch reason and return only for the specific error, like RWX mode not supported by provisioner
+
+			// events := &corev1.EventList{}
+			// listOpts := &client.ListOptions{
+			// 	Namespace:     r.BucketLoggingPVC.Namespace,
+			// 	FieldSelector: fields.OneTermEqualSelector("involvedObject.uid", string(r.BucketLoggingPVC.UID)),
+			// }
+			// if err := r.Client.List(r.Ctx, events, listOpts); err != nil {
+			// 	return err
+			// }
```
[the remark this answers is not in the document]
achouhan09: I am trying to get events on the pvc which is in pending state and will check for specific error related to `RWX` mode and return with the error like `RWX mode not supported` but for different provisioners I found a different errors and description. So, this approach doesn't work here as per my understanding we are not getting a way to identify that the StorageClass(provided by user) would supports `RWX` or not.

Thats why I have commented this part and we need to think of a different approach here. Thanks

pkg/system/phase4_configuring.go
achouhan09: Thanks, I will take care of it.

Actually the idea here is to update the PVC if new StorageClass was provided by user and PVC is in pending state from the start(PV not provisioned). Although we will return before it on the basis of specific error if it is possible.

pkg/system/phase4_configuring.go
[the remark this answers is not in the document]
liranmauda: I believe we should avoid leaving dead code in comments, especially if it isn't functional.
could you remove it?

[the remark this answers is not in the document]
achouhan09: Sure.

pkg/system/phase4_configuring.go
```diff
@@ -1425,6 +1429,76 @@ func (r *Reconciler) ReconcileOBCStorageClass() error {
...
+			"'Guaranteed' BucketLogging requires a Persistent Volume Claim (PVC) with ReadWriteMany (RWX) access mode. The specified 'BucketLoggingPVC' does not found")
+	}
+
+	util.KubeCheck(r.BucketLoggingPVC)
+	if err := r.prepareODFBucketLoggingPVC(); err != nil {
+		return err
+	}
+	r.Own(r.BucketLoggingPVC)
+
+	if r.BucketLoggingPVC.UID != "" {
```
jackyalbo, line 1466: this check needs to come before prepareODFBucketLoggingPVC, if PVC already created we don't need to recreate it

pkg/system/phase4_configuring.go
```diff
@@ -1425,6 +1429,76 @@ func (r *Reconciler) ReconcileOBCStorageClass() error {
...
+	log := r.Logger.WithField("func", "ReconcileBucketLoggingPVC")
+
+	// Return if Bucket logging type is either not specified or not set to 'guaranteed'
+	if r.NooBaa.Spec.BucketLogging.LoggingType != nbv1.BucketLoggingTypeGuaranteed {
+		log.Info("Bucket logging type is either not specified or not set to 'Guaranteed'. skipping ReconcileODFBucketloggingPVC")
+		return nil
+	}
+
+	// Return if bucket logging PVC already exist and support RWX access mode
+	if r.NooBaa.Spec.BucketLogging.BucketLoggingPVC != nil {
```
jackyalbo, line 1443: Is there any reason all of these checks can't move to phase1_verifying?

pkg/system/phase4_configuring.go
achouhan09: Done, Thanks.

pkg/system/phase4_configuring.go
achouhan09: Moved it. Thanks

pkg/system/phase1_verifying.go
```diff
@@ -252,3 +273,26 @@ func (r *Reconciler) checkExternalPg(postgresDbURL string) error {
 	}
 	return nil
 }
+
+func (r *Reconciler) checkBucketLoggingPVC() error {
```
jackyalbo, line 277: Add a short description about what this function does.

pkg/system/phase1_verifying.go
```diff
@@ -252,3 +273,26 @@ func (r *Reconciler) checkExternalPg(postgresDbURL string) error {
...
+	BucketLoggingPVC := &corev1.PersistentVolumeClaim{
+		TypeMeta: metav1.TypeMeta{Kind: "PersistenVolumeClaim"},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      *r.NooBaa.Spec.BucketLogging.BucketLoggingPVC,
+			Namespace: r.Request.Namespace,
+		},
+	}
+	if !util.KubeCheck(BucketLoggingPVC) {
+		return util.NewPersistentError("InvalidBucketLoggingConfiguration",
+			"'Guaranteed' BucketLogging requires a Persistent Volume Claim (PVC) with ReadWriteMany (RWX) access mode. The specified 'BucketLoggingPVC' was not found")
```
jackyalbo, line 287: I think we can take out all of beginning part: 
```suggestion
			"The specified 'BucketLoggingPVC' was not found")
```

pkg/system/phase1_verifying.go
```diff
@@ -252,3 +273,26 @@ func (r *Reconciler) checkExternalPg(postgresDbURL string) error {
...
+	}
+
+	// Check if pvc supports RWX access mode
+	for _, accessMode := range BucketLoggingPVC.Spec.AccessModes {
+		if accessMode == corev1.ReadWriteMany {
+			return nil
+		}
+	}
+	return util.NewPersistentError("InvalidBucketLoggingConfiguration",
+		"'Guaranteed' BucketLogging requires a Persistent Volume Claim (PVC) with ReadWriteMany (RWX) access mode. The specified 'BucketLoggingPVC' does not support RWX access mode")
```
jackyalbo, line 297: Also here we can take this out:
```suggestion
		"The specified 'BucketLoggingPVC' does not support RWX access mode")
```

pkg/system/phase1_verifying.go
```diff
@@ -171,6 +180,18 @@ func (r *Reconciler) CheckSystemCR() error {
 			fmt.Sprintf("Autoscaler %s missing prometheusNamespace property ", r.NooBaa.Spec.Autoscaler.AutoscalerType))
 	}
 
+	// Rejecting NooBaa CR if 'BucketLoggingPVC' is not provided for 'guaranteed' logging and
```
jackyalbo, line 183: I think we can combine both check in 69 with this to one if that check both case. in anyway we want to check the bucketLoggingPVC just in case we use BucketLoggingTypeGuaranteed

pkg/system/phase1_verifying.go
achouhan09: Done.

pkg/system/phase1_verifying.go
achouhan09: Make sense I will update it. Thanks.

jackyalbo (approved):

PR #1380, opened 2024-06-18
[Direct to 5.15] Added error handling and retries in postgres upgrade scripts

dannyzaken:
### Explain the changes
* Postgres upgrade scripts are being run in an init container. since postgres server is required for dump and restore in different Postgres versions, it is being run in the background in the upgrade scripts using the container's script `/usr/bin/run-postgresql`.
* `run-postgresql` runs postgres twice, first time for initialization if needed, and second time to actually run the server with the given params.
* in some cases the upgrade script identifies postgres as ready in the initialization run, and starts the restore from dump. then if `run-postgresql` stops the server, the restore fails.
* To fix - added 30 sec sleep after pg_isready, to alow for the init phase to complete (usually it should take a few seconds)
* Also added error handling and retries in case there are any errors in the dump or restore phases.

1. https://bugzilla.redhat.com/show_bug.cgi?id=2280834

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/internal/configmap-postgres-initdb.yaml
+++ b/deploy/internal/configmap-postgres-initdb.yaml
@@ -54,7 +54,24 @@ data:
           fi
           echo "Info: Free space $USE% is below $THRESHOLD% threshold. Starting upgrade!"
           until pg_isready; do sleep 1; done;
+          # after pg_isready, sleep for 30 seconds, since /usr/bin/run-postgresql script will restart the postgres service
+          sleep 30
+          # make sure again that postgres is running and ready
+          until pg_isready; do sleep 1; done;
+          set +e
+          # dump the db. retry 3 times in case of failure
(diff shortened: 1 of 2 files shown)


deploy/internal/configmap-postgres-initdb.yaml
```diff
@@ -54,7 +54,24 @@ data:
...
+            if [ $? -eq 0 ]; then
+              break
+            elif [ $i -eq 3 ]; then
+              echo "Failed to dump the db"
+              exit 1
+            fi
+            sleep 10
+          done
+          until pg_isready; do sleep 1; done;
+          pg_dumpall -U postgres > /$HOME/data/dump.sql
```
liranmauda, line 74: Is this a left out?

PR #1389, opened 2024-07-09
Added support to join external noobaa system from hosted clusters

ezio-auditore:
### Explain the changes
Part of : [RHSTOR-5187](https://issues.redhat.com//browse/RHSTOR-5187)
1. Edited Join Secret check to differentiate between `remote endpoints` behaviour and `remote noobaa operator` via a annonation `remote-client-noobaa:true`
2. Create a `auth_token` secret for noobaa accounts having an annotation `remote-operator`
3. Delete the auth_token secret when noobaa account is deleted

1.

### Testing Instructions:
1. This change works on a the provider-client model of OCS. 
2. When a new client cluster joins with ocs provider cluster, it creates a Noobaa account CR in the provider cluster
4. An auth token is created for each noobaa remote operator account and is supplied to the client cluster
5. ocs-client operator creates a join secret containing `auth_token` and `mgmt_addr` obtained from the provider noobaa cluster
6. ocs-client opertor creates a Noobaa CR with a secret referring to the join secret created.
7. Noobaa operator then reconciles the client Noobaa CR
8. New OBCs can be created in the client cluster
Ref: https://github.com/red-hat-storage/ocs-operator/pull/2680 
https://github.com/red-hat-storage/ocs-client-operator/pull/176

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/noobaaaccount/reconciler.go
+++ b/pkg/noobaaaccount/reconciler.go
@@ -305,6 +305,17 @@ func (r *Reconciler) ReconcileDeletion() error {
 			return fmt.Errorf("failed to delete account %q. got error: %v", r.NooBaaAccount.Name, err)
 		}
 	} else {
+		if exists := util.IsAnnotationPresent(r.NooBaa.GetAnnotations(), "remote-operator"); exists {
+			nbremoteAuthToken := &corev1.Secret{}
+			nbremoteAuthToken.Name = r.NooBaaAccount.Name
+			nbremoteAuthToken.Namespace = r.NooBaaAccount.Namespace
+			err = r.Client.Delete(r.Ctx, r.Secret)
+			if err != nil {
+				r.Logger.Errorf("got error on deleting remote noobaa secret. error: %v", err)
+			}
+			r.Logger.Infof("✅ Successfully deleted auth_token secret %q", r.NooBaaAccount.Name)
+
+		}
 		r.Logger.Infof("✅ Successfully deleted account %q", r.NooBaaAccount.Name)
 	}
 	return r.FinalizeDeletion()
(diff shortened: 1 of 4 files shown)


pkg/noobaaaccount/reconciler.go
```diff
@@ -373,6 +374,30 @@ func (r *Reconciler) CreateNooBaaAccount() error {
 		r.Logger.Errorf("got error on NooBaaAccount creation. error: %v", err)
 		return err
 	}
+	// create join secret conatining auth token for remote noobaa account
+	if r.NooBaaAccount.Spec.Role != "" && r.NooBaaAccount.Spec.Role == "noobaa-remote" {
+		res, err := r.NBClient.CreateAuthAPI(nb.CreateAuthParams{
+			System: r.Request.Name,
+			Role:   "remote-operator",
```
dannyzaken, line 381: Do we need to distinguish between `operator` and `remote-operator`? in any case, you need support for that in the core side

https://github.com/noobaa/noobaa-core/blob/a6e8837cb96be68fb76f8e5b007b28fe80045001/src/server/system_services/schemas/role_schema.js#L26-L30

dannyzaken: I'll go over noobaa-core, to see if now when we have access control by owner_id, maybe we don't need to distinguish between operator and remote-operator

ezio-auditore: Using existing `operator` role

pkg/noobaaaccount/reconciler.go
```diff
@@ -305,6 +305,16 @@ func (r *Reconciler) ReconcileDeletion() error {
 			return fmt.Errorf("failed to delete account %q. got error: %v", r.NooBaaAccount.Name, err)
 		}
 	} else {
+		if _, exists := r.NooBaaAccount.GetAnnotations()["remote-operator"]; exists {
```
dannyzaken, line 308: check if `Annotations` map exist, since it's optional

dannyzaken: see `GetAnnotationValue` that Amit has added here. can be useful

ezio-auditore: Updated

pkg/util/util.go
```diff
@@ -2190,3 +2190,13 @@ func IsDevEnv() bool {
 	}
 	return false
 }
+
+// IsAnnotationPresent checks if an annotation is
+// is present in the Annotations map
+func IsAnnotationPresent(annotations map[string]string, name string) bool {
```
dannyzaken, line 2196: Please use the `GetAnnotationValue` function implemented in Amit's PR [here](https://github.com/noobaa/noobaa-operator/pull/1393/files#diff-0fa449830b5dc2b274b7c95a58b5071edc23f1e44af84a99c1fa9e600767829fR1427) once it's merged

pkg/noobaaaccount/reconciler.go
```diff
@@ -305,6 +305,17 @@ func (r *Reconciler) ReconcileDeletion() error {
 			return fmt.Errorf("failed to delete account %q. got error: %v", r.NooBaaAccount.Name, err)
 		}
 	} else {
+		if exists := util.IsAnnotationPresent(r.NooBaa.GetAnnotations(), "remote-operator"); exists {
```
dannyzaken, line 308: I think it's better to check for a specific value, and not just the existance of the annotation

PR #1390, opened 2024-07-09
Checking for pvc only if BucketLoggingPVC is not nil

achouhan09:
### Explain the changes
1.  Checking for bucket logging pvc only if BucketLoggingPVC is provided by the user. Currently, checks are failing if cephfs storage class is already present inside the cluster.

1. Improvement: #1368 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/phase1_verifying.go
+++ b/pkg/system/phase1_verifying.go
@@ -266,11 +266,15 @@ func (r *Reconciler) checkExternalPg(postgresDbURL string) error {
 func (r *Reconciler) checkBucketLoggingPVC() error {
 	// Rejecting if 'BucketLoggingPVC' is not provided for 'guaranteed' logging and
 	// also the operator is not running in the ODF environment.
-	sc := &storagev1.StorageClass{
-		TypeMeta:   metav1.TypeMeta{Kind: "StorageClass"},
-		ObjectMeta: metav1.ObjectMeta{Name: "ocs-storagecluster-cephfs"},
-	}
-	if r.NooBaa.Spec.BucketLogging.BucketLoggingPVC == nil && !util.KubeCheck(sc) {
+	if r.NooBaa.Spec.BucketLogging.BucketLoggingPVC == nil {
+		sc := &storagev1.StorageClass{
+			TypeMeta:   metav1.TypeMeta{Kind: "StorageClass"},
+			ObjectMeta: metav1.ObjectMeta{Name: "ocs-storagecluster-cephfs"},
+		}
+		// Return nil as the operator running in ODF environment
+		if util.KubeCheck(sc) {
+			return nil
+		}
 		return util.NewPersistentError("InvalidBucketLoggingConfiguration",
 			"'Guaranteed' BucketLogging requires a Persistent Volume Claim (PVC) with ReadWriteMany (RWX) access mode. Please specify the 'BucketLoggingPVC' to ensure guaranteed logging")
 	}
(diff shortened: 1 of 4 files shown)


pkg/system/phase1_verifying.go
```diff
@@ -275,25 +275,28 @@ func (r *Reconciler) checkBucketLoggingPVC() error {
...
+		BucketLoggingPVC := &corev1.PersistentVolumeClaim{
+			TypeMeta: metav1.TypeMeta{Kind: "PersistenVolumeClaim"},
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      *r.NooBaa.Spec.BucketLogging.BucketLoggingPVC,
+				Namespace: r.Request.Namespace,
+			},
+		}
+		if !util.KubeCheck(BucketLoggingPVC) {
+			return util.NewPersistentError("InvalidBucketLoggingConfiguration",
+				"The specified 'BucketLoggingPVC' was not found")
```
aspandey, line 289: As you are sending this patch, log name of the pvc also, *r.NooBaa.Spec.BucketLogging.BucketLoggingPVC.
On line 288 too.

pkg/system/phase1_verifying.go
```diff
@@ -275,25 +275,28 @@ func (r *Reconciler) checkBucketLoggingPVC() error {
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      *r.NooBaa.Spec.BucketLogging.BucketLoggingPVC,
-			Namespace: r.Request.Namespace,
-		},
-	}
-	if !util.KubeCheck(BucketLoggingPVC) {
-		return util.NewPersistentError("InvalidBucketLoggingConfiguration",
-			"The specified 'BucketLoggingPVC' was not found")
-	}
+	if r.NooBaa.Spec.BucketLogging.BucketLoggingPVC != nil {
```
aspandey, line 278: Observation on line number 273 and on this condition + your comment on changes 
"Checking for bucket logging pvc only if BucketLoggingPVC is provided by the user. Currently, checks are failing if cephfs storage class is already present inside the cluster."

Let's say BucketLoggingPVC != nil so online 273 it will not check if cephfs sc is present or not.
Now, it will enter in this condition and will have he same impact. Right?

Is this what we want to check?

pkg/system/phase1_verifying.go
```diff
@@ -275,25 +275,28 @@ func (r *Reconciler) checkBucketLoggingPVC() error {
 			"'Guaranteed' BucketLogging requires a Persistent Volume Claim (PVC) with ReadWriteMany (RWX) access mode. Please specify the 'BucketLoggingPVC' to ensure guaranteed logging")
```
jackyalbo, line 275: I would just add here the check.
```suggestion
     if r.NooBaa.Spec.BucketLogging.BucketLoggingPVC == nil {
            if ceph-fs exist -> good -> return nil
            else throw the persistent error
     } 
```
rest of the code is the same we know for BucketLoggingPVC is not nil

pkg/system/phase1_verifying.go
jackyalbo: I agree let's add to the prints the name of the provided PVC, also in 298.

achouhan09: Sure, make sense. Thanks

pkg/system/phase1_verifying.go
achouhan09: This should be the best way, I will update. Thanks

pkg/system/phase1_verifying.go
achouhan09: Updated, Thanks.

aspandey (commented):
LGTM

jackyalbo (approved):

PR #1393, opened 2024-07-22
Implement operator support for pool ownership

bernerhat:
### Explain the changes
1.  part of the change required in [Implement OCS and MCG operator support for pool ownership](https://issues.redhat.com/browse/RHSTOR-6123) adds an indication to the core that the cluster is set to provider mode upon deployment.

1. This implementation supports [RHSTOR-5187 - Support Provider & Consumer mode](https://issues.redhat.com/browse/RHSTOR-5187)

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -450,6 +450,11 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 	r.CoreApp.Spec.Selector.MatchLabels["noobaa-core"] = r.Request.Name
 	r.CoreApp.Spec.ServiceName = r.ServiceMgmt.Name
 
+	//check if provider mode is enabled and signal the core
+	if annotationValue, _ := util.GetAnnotationValue(r.NooBaa.Annotations, "MulticloudObjectGatewayProviderMode"); annotationValue == "true" {
+		util.GetEnvVariable(&r.CoreApp.Spec.Template.Spec.Containers[0].Env, "CONFIG_JS_RESTRICT_RESOURCE_DELETION").Value = "true"
+	}
+
 	podSpec := &r.CoreApp.Spec.Template.Spec
 	podSpec.ServiceAccountName = "noobaa"
 	coreImageChanged := false
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -1424,6 +1424,15 @@ func GetEnvVariable(env *[]corev1.EnvVar, name string) *corev1.EnvVar {
 	return nil
 }
 
+// GetAnnotationValue searches for an annotation within a map of strings and returns if it exists and what its value
+func GetAnnotationValue(annotations map[string]string, name string) (string, bool) {
+	if annotations != nil {
+		val, exists := annotations[name]
+		return val, exists
+	}
+	return "", false
+}
+
 // ReflectEnvVariable will add, update or remove an env variable base on the existence and value of an
 // env variable with the same name on the container running this function.
 func ReflectEnvVariable(env *[]corev1.EnvVar, name string) {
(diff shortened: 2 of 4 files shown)


pkg/system/phase2_creating.go
```diff
@@ -450,6 +450,11 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 	r.CoreApp.Spec.Selector.MatchLabels["noobaa-core"] = r.Request.Name
 	r.CoreApp.Spec.ServiceName = r.ServiceMgmt.Name
 
+	//check if provider mode is enabled and signal the core
+	if util.GetAnnotationValue(r.NooBaa.Annotations, "MulticloudObjectGatewayProviderMode") == "true" {
+		util.GetEnvVariable(&r.CoreApp.Spec.Template.Spec.Containers[0].Env, "RESTRICT_RESOURCE_DELETION").Value = "true"
```
dannyzaken, line 455: to modify the `RESTRICT_RESOURCE_DELETION` config you need to add the `CONFIG_JS_` prefix in the env.
 ```suggestion
		util.GetEnvVariable(&r.CoreApp.Spec.Template.Spec.Containers[0].Env, "CONFIG_JS_RESTRICT_RESOURCE_DELETION").Value = "true"
```

File: pkg/util/util.go [hunk 2]
```diff
@@ -1424,6 +1424,14 @@ func GetEnvVariable(env *[]corev1.EnvVar, name string) *corev1.EnvVar {
 	return nil
 }
 
+func GetAnnotationValue(annotations map[string]string, name string) string {
+	val, err := annotations[name]
```
dannyzaken, line 1428: verify `annotations != nil`

pkg/system/phase2_creating.go
bernerhat: ack, changed.

pkg/util/util.go
bernerhat: ack, changed.

pkg/util/util.go
```diff
@@ -1424,6 +1424,17 @@ func GetEnvVariable(env *[]corev1.EnvVar, name string) *corev1.EnvVar {
 	return nil
 }
 
+func GetAnnotationValue(annotations map[string]string, name string) string {
+	if annotations != nil {
+		val, err := annotations[name]
+		if err {
+			return ""
+		}
+		return val
```
dannyzaken, line 1433: This code does the opposite of what you intended. Golang map returns `ok` as the second return value, so it is true if the value exist.
Also, if it does not exist, it should return a zero value, which in a string is `""`, so there is no need to really check for `ok`.

```suggestion
		return annotations[name]
```

pkg/util/util.go
```diff
@@ -1424,6 +1424,13 @@ func GetEnvVariable(env *[]corev1.EnvVar, name string) *corev1.EnvVar {
 	return nil
 }
 
+func GetAnnotationValue(annotations map[string]string, name string) string {
```
dannyzaken, line 1427: It is worth returning a boolean `exists` as a second return value, the same as the map return value.

dannyzaken: see this change in Kaustav's PR
https://github.com/noobaa/noobaa-operator/pull/1389/files#diff-0fa449830b5dc2b274b7c95a58b5071edc23f1e44af84a99c1fa9e600767829fR2196

bernerhat: ack, changed.

dannyzaken (approved):
one small comment (it will probably fail the tests anyway) . otherwise approved

dannyzaken:
@bernerhat The unit tests failed due to diff with the expected generated files.
please run the following and commit
```
go run pkg/bundler/bundler.go deploy/ pkg/bundle/deploy.go
```

bernerhat:
@dannyzaken done

PR #1396, opened 2024-07-25
Keep using psql 12 if failed upgrading to psql15

dannyzaken:
### Explain the changes
1. If Postgres has not been upgraded to Postgres-15 and is still using the Postgres-12 image, continue to use the Postgres-12 image, but set the noobaa CR as rejected. 
2. This prevents noobaa-db from going into CLBO due to postgres-15 running with postgres-12 data

1. https://bugzilla.redhat.com/show_bug.cgi?id=2299835

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/system.go
+++ b/pkg/system/system.go
@@ -1116,7 +1116,19 @@ func Connect(isExternal bool) (*Client, error) {
 }
 
 // GetDesiredDBImage returns the desired DB image according to spec or env or default (in options)
-func GetDesiredDBImage(sys *nbv1.NooBaa) string {
+func GetDesiredDBImage(sys *nbv1.NooBaa, currentImage string) string {
+	// Postgres upgrade failure workaround
+	// if the current Postgres image is a postgresql-12 image, use NOOBAA_PSQL12_IMAGE. otherwise use GetdesiredDBImage
+	if IsPostgresql12Image(currentImage) {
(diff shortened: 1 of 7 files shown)


pkg/system/system.go
```diff
@@ -1116,7 +1116,19 @@ func Connect(isExternal bool) (*Client, error) {
...
 // GetDesiredDBImage returns the desired DB image according to spec or env or default (in options)
-func GetDesiredDBImage(sys *nbv1.NooBaa) string {
+func GetDesiredDBImage(sys *nbv1.NooBaa, currentImage string) string {
+	// Postgres upgrade failure workaround
+	// if the current Postgres image is a postgresql-12 image, use NOOBAA_PSQL12_IMAGE. otherwise use GetdesiredDBImage
+	if IsPostgresql12Image(currentImage) {
+		psql12Image, ok := os.LookupEnv("NOOBAA_PSQL_12_IMAGE")
+		util.Logger().Warnf("The current Postgres image is a postgresql-12 image. using (%s)", psql12Image)
+		if !ok {
+			psql12Image = currentImage
```
tangledbytes, line 1126: I am not sure if I understand the entire context. We can't return `options.Psql12Image` here?

tangledbytes (approved):

PR #1399, opened 2024-07-29
disable automatic mount of SA token for pv-pool pod

dannyzaken:
### Explain the changes
1. Kubernetes automatically mounts a service account token for every pod by default. 
2. The best practice is to mount it only for pods that require access to the Kubernetes API. 
3. For now, disabling automount for pv-pool pods. We should consider it for all pods other than the operator.
4. see more info [here](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#opt-out-of-api-credential-automounting) 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/internal/pod-agent.yaml
+++ b/deploy/internal/pod-agent.yaml
@@ -35,6 +35,7 @@ spec:
       securityContext:
         runAsNonRoot: true
         allowPrivilegeEscalation: false
+  automountServiceAccountToken: false
   securityContext:
     runAsUser: 10001
     runAsGroup: 0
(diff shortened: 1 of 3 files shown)


tangledbytes:
This reminds me that we have `create_k8s_auth` in noobaa-core, I can't find its usage right now but that seems to read SA token.

tangledbytes (approved):

nimrod-becker (approved):

dannyzaken:
> This reminds me that we have `create_k8s_auth` in noobaa-core, I can't find its usage right now but that seems to read SA token.

I think it was used for SSO into noobaa UI, so it's not relevant anymore

PR #1401, opened 2024-07-31
Added TopologySpreadConstraints to the backingstore pod for better performance

achouhan09:
### Explain the changes
1. Currently, when we create a pv backingstore then the backingstore pod got scheduled on the random basis on any node. This is not good as if any node goes down all the backingstore pod on that node will be down. To increase the performance we have added topology spread constraints to the backingstore pod so they get scheduled on spread basis.

1. fix: https://bugzilla.redhat.com/show_bug.cgi?id=2294846

### Testing Instructions:
1. After installing noobaa, try creating new pv-backingstore and check if the pv backingstore pod got scheduled and spread across the multiple nodes. Each pod should be spread in different nodes with a maximum skew difference of 1.

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -1277,14 +1277,37 @@ func (r *Reconciler) updatePodTemplate() error {
 			[]corev1.LocalObjectReference{*r.NooBaa.Spec.ImagePullSecret}
 	}
 	r.PodAgentTemplate.Labels = map[string]string{
-		"app":  "noobaa",
-		"pool": r.BackingStore.Name,
+		"app":                 "noobaa",
+		"pool":                r.BackingStore.Name,
+		"noobaa-backingstore": r.Request.Name,
 	}
 	if r.NooBaa.Spec.Tolerations != nil {
 		r.PodAgentTemplate.Spec.Tolerations = r.NooBaa.Spec.Tolerations
 	}
 	if r.NooBaa.Spec.Affinity != nil {
 		r.PodAgentTemplate.Spec.Affinity = r.NooBaa.Spec.Affinity
+	}
+
+	// Add topology spread constraints
+	honor := corev1.NodeInclusionPolicyHonor
+	if r.PodAgentTemplate.Spec.TopologySpreadConstraints != nil {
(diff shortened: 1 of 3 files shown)


pkg/backingstore/reconciler.go
```diff
@@ -1277,8 +1277,9 @@ func (r *Reconciler) updatePodTemplate() error {
 			[]corev1.LocalObjectReference{*r.NooBaa.Spec.ImagePullSecret}
 	}
 	r.PodAgentTemplate.Labels = map[string]string{
-		"app":  "noobaa",
-		"pool": r.BackingStore.Name,
+		"app":                 "noobaa",
+		"pool":                r.BackingStore.Name,
+		"noobaa-backingstore": r.Request.Name,
```
naveenpaul1, line 1282: why `noobaa-backingstore` added?

pkg/backingstore/reconciler.go
```diff
@@ -1287,6 +1288,28 @@ func (r *Reconciler) updatePodTemplate() error {
 		r.PodAgentTemplate.Spec.Affinity = r.NooBaa.Spec.Affinity
 	}
 
+	// Add topology spread constraints
+	honor := corev1.NodeInclusionPolicyHonor
+	if r.PodAgentTemplate.Spec.TopologySpreadConstraints != nil {
+		r.Logger.Info("TopologySpreadConstraints already exists...")
```
naveenpaul1, line 1294: mention it is backing store in log, to make debugging easier.

achouhan09: Sure, I will try to do something better. Thanks

pkg/backingstore/reconciler.go
achouhan09: We need to add a label to all the backingstore pods so that topology constraints should work. The current two labels would not work in here so I need to add one. Please correct me if I am wrong. Thanks

pkg/backingstore/reconciler.go
achouhan09: We cannot print the backingstore name here, as in this step we are updating the pod template itself for backingstore pod

liranmauda (approved):

PR #1402, opened 2024-08-01
Fix provider env var panic

bernerhat:
### Explain the changes
1. moved the check for the env var to the appropriate location
2. changed env var location in yaml file and made it optional

1. https://bugzilla.redhat.com/show_bug.cgi?id=2301889
2. change introduced as part of [epic](https://issues.redhat.com/browse/RHSTOR-5187) were causing clusters deployed in provider mode to fail deployment, operator crashed while trying to change env var value which was not found.

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/internal/statefulset-core.yaml
+++ b/deploy/internal/statefulset-core.yaml
@@ -93,6 +93,8 @@ spec:
                 configMapKeyRef:
                   name: noobaa-config
                   key: NOOBAA_LOG_LEVEL
+            - name: RESTRICT_RESOURCE_DELETION
+              value: "false"
             - name: POSTGRES_HOST
               value: "noobaa-db-pg-0.noobaa-db-pg"
             - name: POSTGRES_PORT
@@ -175,5 +177,3 @@ spec:
               valueFrom:
                 resourceFieldRef:
                   resource: limits.memory
-            - name: CONFIG_JS_RESTRICT_RESOURCE_DELETION
-              value: "false"
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -38,6 +38,8 @@ import (
 const (
 	webIdentityTokenPath string = "/var/run/secrets/openshift/serviceaccount/token"
 	roleARNEnvVar        string = "ROLEARN"
+	trueStr              string = "true"
+	falseStr             string = "false"
 )
 
 // ReconcilePhaseCreating runs the reconcile phase
@@ -434,6 +436,18 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 			} else {
 				c.Env[j].Value = ""
 			}
+		case "RESTRICT_RESOURCE_DELETION":
+			// check if provider mode is enabled and signal the core
+			annotationValue, annotationExists := util.GetAnnotationValue(r.NooBaa.Annotations, "MulticloudObjectGatewayProviderMode")
+			annotationBoolVal := false
+			if annotationExists {
+				annotationBoolVal = strings.ToLower(annotationValue) == trueStr
+			}
+			if annotationBoolVal {
+				c.Env[j].Value = trueStr
+			} else {
+				c.Env[j].Value = falseStr
+			}
 		}
 	}
 }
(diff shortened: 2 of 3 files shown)


liranmauda (commented):
LGTM
@dannyzaken WDYT?

pkg/system/phase2_creating.go
```diff
@@ -1239,6 +1235,15 @@ func (r *Reconciler) SetDesiredCoreAppConfig() error {
 		"DISABLE_DEV_RANDOM_SEED":    "true",
 		"NOOBAA_LOG_LEVEL":           "default_level",
 	}
+	// check if provider mode is enabled and signal the core
+	annotationValue, annotationExists := util.GetAnnotationValue(r.NooBaa.Annotations, "MulticloudObjectGatewayProviderMode")
+	annotationBoolVal := false
+	if annotationExists {
+		annotationBoolVal = strings.ToLower(annotationValue) == strconv.FormatBool(true)
```
dannyzaken, line 1242: Why is `strconv.FormatBool` necessary?

dannyzaken (changes requested):
There is no need to pass the env through the noobaa-config config map. the purpose of this configmap is to provide configurable envs that can be modified by the user. This env is controlled by the annotation and is not meant to be configured by the user.

I suggest adding this env to the `setDesiredCoreEnv`function. add a case for this env that will set the value according to the annotation

pkg/system/phase2_creating.go
bernerhat: to prevent typos at the very least, i try to avoid hardcoding text strings if not necessary

pkg/system/phase2_creating.go
```diff
@@ -433,6 +434,16 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 			} else {
 				c.Env[j].Value = ""
 			}
+		case "CONFIG_JS_RESTRICT_RESOURCE_DELETION":
+			// check if provider mode is enabled and signal the core
+			annotationValue, annotationExists := util.GetAnnotationValue(r.NooBaa.Annotations, "MulticloudObjectGatewayProviderMode")
+			annotationBoolVal := false
+			if annotationExists {
+				annotationBoolVal = strings.ToLower(annotationValue) == strconv.FormatBool(true)
```
dannyzaken, line 442: The `strconv` is not so readable. Please use a const instead. You can define `trueStr` or a similar const in the file scope. Use it also for setting the env value below

dannyzaken (commented):
one small comment. other than that looks good

pkg/system/phase2_creating.go
```diff
@@ -433,6 +434,16 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
...
+		case "CONFIG_JS_RESTRICT_RESOURCE_DELETION":
+			// check if provider mode is enabled and signal the core
+			annotationValue, annotationExists := util.GetAnnotationValue(r.NooBaa.Annotations, "MulticloudObjectGatewayProviderMode")
+			annotationBoolVal := false
+			if annotationExists {
+				annotationBoolVal = strings.ToLower(annotationValue) == trueStr
+			}
+			if annotationBoolVal {
+				c.Env[j].Value = "true"
+			}
```
dannyzaken, line 446:
```suggestion
			if annotationBoolVal {
				c.Env[j].Value = trueStr
			} else {
			        c.Env[j].Value = falseStr
			}
```

deploy/internal/statefulset-core.yaml
```diff
@@ -95,6 +95,8 @@ spec:
                 configMapKeyRef:
                   name: noobaa-config
                   key: NOOBAA_LOG_LEVEL
+            - name: CONFIG_JS_RESTRICT_RESOURCE_DELETION
```
liranmauda, line 98: An env variable should not start with `CONFIG_JS_`
```suggestion
            - name: RESTRICT_RESOURCE_DELETION
```

pkg/system/phase2_creating.go
```diff
@@ -433,6 +434,16 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 			} else {
 				c.Env[j].Value = ""
 			}
+		case "CONFIG_JS_RESTRICT_RESOURCE_DELETION":
```
liranmauda, line 437:
```suggestion
		case "RESTRICT_RESOURCE_DELETION":
```

liranmauda (changes requested):

deploy/internal/statefulset-core.yaml
```diff
@@ -95,6 +95,8 @@ spec:
                 configMapKeyRef:
                   name: noobaa-config
                   key: NOOBAA_LOG_LEVEL
+            - name: RESTRICT_RESOURCE_DELETION
+              value: "false"
```
liranmauda, line 99: We do not need a default.
anything other than `true` will result in false.
With that said, we can keep it to make it more explicit. 
@dannyzaken WDYT?

dannyzaken (approved):

deploy/internal/statefulset-core.yaml
dannyzaken: we can leave the default

nb-ohad: @liranmauda Can you please approve the PR so this could merge

liranmauda (approved):

PR #1405, opened 2024-08-05
Added support to join external noobaa system from hosted clusters

bernerhat:
### Explain the changes
1.  continuation of Kaustav's [PR](https://github.com/noobaa/noobaa-operator/pull/1389) 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/noobaaaccount/reconciler.go
+++ b/pkg/noobaaaccount/reconciler.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"strings"
 	"time"
 
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
@@ -22,6 +23,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+const (
+	strTrue string = "true"
 )
 
 // Reconciler is the context for loading or reconciling a noobaa system
@@ -351,22 +356,38 @@ func (r *Reconciler) CreateNooBaaAccount() error {
 		return err
 	}
 
-	var accessKeys nb.S3AccessKeys
-	// if we didn't get the access keys in the create_account reply we might be talking to an older noobaa version (prior to 5.1)
-	// in that case try to get it using read account
-	if len(accountInfo.AccessKeys) == 0 {
-		log.Info("CreateAccountAPI did not return access keys. calling ReadAccountAPI to get keys..")
-		readAccountReply, err := r.NBClient.ReadAccountAPI(nb.ReadAccountParams{Email: r.NooBaaAccount.Name})
-		if err != nil {
-			return err
+	annotationValue, exists := util.GetAnnotationValue(r.NooBaaAccount.Annotations, "remote-operator")
+	if exists {
+		if strings.ToLower(annotationValue) == strTrue {
+			// create join secret conatining auth token for remote noobaa account
+			res, err := r.NBClient.CreateAuthAPI(nb.CreateAuthParams{
+				System: r.NooBaa.Name,
+				Role:   "operator",
+				Email:  options.OperatorAccountEmail,
+			})
+			if err != nil {
+				return fmt.Errorf("cannot create an auth token for remote operator, error: %v", err)
+			}
+			r.Secret.StringData["auth_token"] = res.Token
 		}
-		accessKeys = readAccountReply.AccessKeys[0]
 	} else {
-		accessKeys = accountInfo.AccessKeys[0]
+		var accessKeys nb.S3AccessKeys
+		// if we didn't get the access keys in the create_account reply we might be talking to an older noobaa version (prior to 5.1)
+		// in that case try to get it using read account
+		if len(accountInfo.AccessKeys) == 0 {
+			log.Info("CreateAccountAPI did not return access keys. calling ReadAccountAPI to get keys..")
+			readAccountReply, err := r.NBClient.ReadAccountAPI(nb.ReadAccountParams{Email: r.NooBaaAccount.Name})
+			if err != nil {
+				return err
+			}
+			accessKeys = readAccountReply.AccessKeys[0]
+		} else {
+			accessKeys = accountInfo.AccessKeys[0]
+		}
+		r.Secret.StringData = map[string]string{}
+		r.Secret.StringData["AWS_ACCESS_KEY_ID"] = string(accessKeys.AccessKey)
+		r.Secret.StringData["AWS_SECRET_ACCESS_KEY"] = string(accessKeys.SecretKey)
 	}
-	r.Secret.StringData = map[string]string{}
-	r.Secret.StringData["AWS_ACCESS_KEY_ID"] = string(accessKeys.AccessKey)
-	r.Secret.StringData["AWS_SECRET_ACCESS_KEY"] = string(accessKeys.SecretKey)
 	r.Own(r.Secret)
 	err = r.Client.Create(r.Ctx, r.Secret)
 	if err != nil {
--- a/pkg/system/phase1_verifying.go
+++ b/pkg/system/phase1_verifying.go
@@ -190,21 +190,24 @@ func (r *Reconciler) CheckJoinSecret() error {
 		return util.NewPersistentError("InvalidJoinSecert",
 			"JoinSecret is missing mgmt_addr")
 	}
-	if r.JoinSecret.StringData["bg_addr"] == "" {
-		return util.NewPersistentError("InvalidJoinSecert",
-			"JoinSecret is missing bg_addr")
-	}
-	if r.JoinSecret.StringData["md_addr"] == "" {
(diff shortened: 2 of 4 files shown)


pkg/noobaaaccount/reconciler.go
```diff
@@ -305,7 +307,20 @@ func (r *Reconciler) ReconcileDeletion() error {
...
 		}
 	} else {
-		r.Logger.Infof("✅ Successfully deleted account %q", r.NooBaaAccount.Name)
+		annotationValue, exists := util.GetAnnotationValue(r.NooBaa.GetAnnotations(), "remote-operator")
+		if exists {
+			if strings.ToLower(annotationValue) != strconv.FormatBool(false) {
+				nbremoteAuthToken := &corev1.Secret{}
+				nbremoteAuthToken.Name = r.NooBaaAccount.Name
+				nbremoteAuthToken.Namespace = r.NooBaaAccount.Namespace
+				err = r.Client.Delete(r.Ctx, r.Secret)
```
dannyzaken, line 316: If I'm not mistaken, R.Secret is deleted automtically, since it's owned by the noobaaAccount CR. I don't think that anything is changed in this PR with respect to deletion. Can you please verify it?

pkg/noobaaaccount/reconciler.go
```diff
@@ -373,6 +388,33 @@ func (r *Reconciler) CreateNooBaaAccount() error {
 		r.Logger.Errorf("got error on NooBaaAccount creation. error: %v", err)
 		return err
 	}
+	// create join secret conatining auth token for remote noobaa account
+	annotationValue, exists := util.GetAnnotationValue(r.NooBaa.GetAnnotations(), "remote-operator")
+	if exists {
+		if strings.ToLower(annotationValue) != strconv.FormatBool(false) {
```
dannyzaken, line 394:
```suggestion
		if strings.ToLower(annotationValue) == "true" {
```

pkg/noobaaaccount/reconciler.go
```diff
@@ -373,6 +388,33 @@ func (r *Reconciler) CreateNooBaaAccount() error {
...
+			res, err := r.NBClient.CreateAuthAPI(nb.CreateAuthParams{
+				System: r.Request.Name,
+				Role:   "operator",
+				Email:  options.OperatorAccountEmail,
+			})
+			if err != nil {
+				return fmt.Errorf("cannot create an auth token for remote operator, error: %v", err)
+			}
+			if res.Token == "" {
```
dannyzaken, line 403: Is an empty token somthing that can be returned by noobaa-core in case there is no error? did you encounter this?

pkg/system/phase1_verifying.go
```diff
@@ -190,22 +192,30 @@ func (r *Reconciler) CheckJoinSecret() error {
...
-		return util.NewPersistentError("InvalidJoinSecert",
-			"JoinSecret is missing hosted_agents_addr")
-	}
 	if r.JoinSecret.StringData["auth_token"] == "" {
 		return util.NewPersistentError("InvalidJoinSecert",
 			"JoinSecret is missing auth_token")
 	}
+
+	annotationValue, exists := util.GetAnnotationValue(r.NooBaa.GetAnnotations(), "remote-client-noobaa")
+	annotationBoolVal := true
```
dannyzaken, line 201: I suggest giving clearer names to the variables. `annotationVal` and `annotationBoolVal` are too generic

pkg/system/phase1_verifying.go
```diff
@@ -190,22 +192,30 @@ func (r *Reconciler) CheckJoinSecret() error {
...
+	if exists {
+		annotationBoolVal = strings.ToLower(annotationValue) == strconv.FormatBool(true)
```
dannyzaken, line 203: Again, unless I'm missing somthing, I don't see a reason to use `strconv.FormatBool(true)` and not `"true"`

pkg/system/phase4_configuring.go
```diff
@@ -66,12 +66,20 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
...
-	}
-	if err := r.ReconcileHPAEndpoint(); err != nil {
-		return err
+	// No endpoint creation is required for remote noobaa client
+	annotationValue, exists := util.GetAnnotationValue(r.NooBaa.GetAnnotations(), "remote-client-noobaa")
+	annotationBoolVal := true
+	if exists {
+		annotationBoolVal = strings.ToLower(annotationValue) == strconv.FormatBool(true)
+	}
+	if !exists || !annotationBoolVal {
```
dannyzaken, line 75: since you repeat this code more than once, mayb extract this to a function `isRemoteClientNoobaa`

dannyzaken (changes requested):

pkg/noobaaaccount/reconciler.go
```diff
@@ -373,6 +388,33 @@ func (r *Reconciler) CreateNooBaaAccount() error {
 		r.Logger.Errorf("got error on NooBaaAccount creation. error: %v", err)
 		return err
 	}
+	// create join secret conatining auth token for remote noobaa account
+	annotationValue, exists := util.GetAnnotationValue(r.NooBaa.GetAnnotations(), "remote-operator")
```
dannyzaken, line 392: If the `remote-operator` annotation exist, do you need to also create access keys (`r.Secret`)?

Maybe just reuse the same `r.Secret` and populate it with the relevant values according to the annotatioin?

pkg/noobaaaccount/reconciler.go
```diff
@@ -373,6 +388,33 @@ func (r *Reconciler) CreateNooBaaAccount() error {
...
+			if res.Token == "" {
+				return fmt.Errorf("received empty auth token")
+			}
+			nbremoteAuthToken := &corev1.Secret{}
+			nbremoteAuthToken.Name = r.NooBaaAccount.Name
+			nbremoteAuthToken.Namespace = r.NooBaaAccount.Namespace
+			nbremoteAuthToken.StringData["auth_token"] = res.Token
+			r.Own(nbremoteAuthToken)
+
+			err = r.Client.Create(r.Ctx, r.Secret)
```
dannyzaken, line 412: did you mean to create `nbremoteAuthToken`?


```suggestion
			err = r.Client.Create(r.Ctx, nbremoteAuthToken)
```

pkg/noobaaaccount/reconciler.go
bernerhat: it's a valid point but since it was Kaustav's decision to use a new secret, i'm not sure what led to him creating a new secret.
Lets leave the new secret in place and when Kaustav returns from his leave we can change this. i'll open an issue for it as well (since changing it will most likely cause changes to code in other repos such as ocs-operator which uses this secret)

dannyzaken: I prefer to make the code simpler and not to create 2 secrets for a single noobaa account. I suggest that for now add an `"auth_token"` key on `r.Secret`. If there is a reason to create a separate secret than we can make this change later

pkg/noobaaaccount/reconciler.go
```diff
@@ -373,6 +375,30 @@ func (r *Reconciler) CreateNooBaaAccount() error {
 		r.Logger.Errorf("got error on NooBaaAccount creation. error: %v", err)
 		return err
 	}
+	// create join secret conatining auth token for remote noobaa account
+	annotationValue, exists := util.GetAnnotationValue(r.NooBaa.GetAnnotations(), "remote-operator")
```
dannyzaken, line 379: I see now that the annotation is on the noobaa CR and not the noobaaAccount CR. In that case, what's the difference between `remote-operator` and `MulticloudObjectGatewayProviderMode`?

bernerhat: The `remote-operator` is an annotation which is being added to the account CR by the ocs-operator while we are on the provider side in order to indicate when we onboard a new consumer and therefore need a new auth token to be added to the account. The `MulticloudObjectGatewayProviderMode` annotation is being added to the Noobaa CR while we are running remotely and need to indicate that we only wish to deploy the operator and connect to a remote cluster.

dannyzaken: But here you check the annotaion on the noobaa CR and not the account CR (`r.NooBaa.GetAnnotations()`)

bernerhat: you are correct it's a mistake, will fix.

pkg/util/util.go
```diff
@@ -1435,6 +1435,17 @@ func GetAnnotationValue(annotations map[string]string, name string) (string, boo
 	return "", false
 }
 
+// IsRemoteClientNoobaa checks for the existance and value of the remote-client-noobaa annotation
+// within an annotation map, if the annotation doesnt exist it's the same as if its value is false.
+func IsRemoteClientNoobaa(annotations map[string]string) bool {
+	annotationValue, exists := GetAnnotationValue(annotations, "remote-client-noobaa")
+	annotationBoolVal := false
+	if exists {
+		annotationBoolVal = strings.ToLower(annotationValue) == strconv.FormatBool(true)
```
dannyzaken, line 1444: `trueStr` instead of strconv

bernerhat: done

PR #1407, opened 2024-08-11
Add NOOBAA_LOG_LEVEL to the agent pod

liranmauda:
### Explain the changes

- Add NOOBAA_LOG_LEVEL to the agent pod
- Watching for changes in the noobaa-config on the backingstore reconciler
- Updating the NOOBAA_LOG_LEVEL according to the config map

### Fixes:
- Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2297448

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -81,6 +81,7 @@ type Reconciler struct {
 	PodAgentTemplate *corev1.Pod
 	PvcAgentTemplate *corev1.PersistentVolumeClaim
 	ServiceAccount   *corev1.ServiceAccount
+	CoreAppConfig    *corev1.ConfigMap
 
 	SystemInfo             *nb.SystemInfo
 	ExternalConnectionInfo *nb.ExternalConnectionInfo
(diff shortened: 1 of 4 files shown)


pkg/backingstore/reconciler.go
```diff
@@ -1160,6 +1167,16 @@ func (r *Reconciler) needUpdate(pod *corev1.Pod) bool {
 			return true
 		}
 	}
+
+	var noobaaLogEnv = "NOOBAA_LOG_LEVEL"
+	var configMapLogLevel = r.CoreAppConfig.Data[noobaaLogEnv]
+	noobaaLogEnvVar := util.GetEnvVariable(&c.Env, noobaaLogEnv)
+
+	if (configMapLogLevel != noobaaLogEnvVar.Value) {
+		r.Logger.Warnf("NOOBAA_LOG_LEVEL Env variable change detected: on the config map")
```
jackyalbo, line 1176: you missed the values in the print:
```suggestion
		r.Logger.Warnf("NOOBAA_LOG_LEVEL Env variable change detected:(%v) on the config map(%v)", noobaaLogEnvVar.Value, configMapLogLevel)
```

pkg/backingstore/reconciler.go
```diff
@@ -154,6 +158,9 @@ func (r *Reconciler) Reconcile() (reconcile.Result, error) {
 		return reconcile.Result{}, nil // final state
 	}
 
+	//Should we fail/ignore in case the config map is not present?
+	util.KubeCheck(r.CoreAppConfig)
```
jackyalbo, line 162: shouldn't this happen only for pvpool? and then maybe you can throw error if it doesn't exist

pkg/controller/backingstore/backingstore_controller.go
```diff
@@ -79,6 +85,10 @@ func Add(mgr manager.Manager) error {
 	if err != nil {
 		return err
 	}
+	err = c.Watch(source.Kind(mgr.GetCache(), &corev1.ConfigMap{}), ownerHandler, &filterForNoobaaOwnerPredicate, &logEventsPredicate)
```
jackyalbo, line 88: wouldn't this will run all backingstores in case of change in the config map? can we run somehow only the pvpool type?

liranmauda: This will only happen when the configmap noobaa-config will change, which is not a lot.
in that case the reconcile will trigger, but the actual change is only on the reconcile of the PV pool.

jackyalbo (approved):

PR #1410, opened 2024-08-13
noobaa/operator: Provide option to disable log colors

aspandey:
### Explain the changes

This patch is in sync with the changes on noobaa-core side.
https://github.com/noobaa/noobaa-core/pull/8281

We need to make sure a user can enable/disable the log color generated by core, endpoint and backing store.

### Fixes:
- Fixes: https://bugzilla.redhat.com/show_bug.cgi?id=2297435

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -1169,13 +1169,14 @@ func (r *Reconciler) needUpdate(pod *corev1.Pod) bool {
 		}
 	}
 
-	var noobaaLogEnv = "NOOBAA_LOG_LEVEL"
-	var configMapLogLevel = r.CoreAppConfig.Data[noobaaLogEnv]
-	noobaaLogEnvVar := util.GetEnvVariable(&c.Env, noobaaLogEnv)
+	for _, name := range []string{"NOOBAA_LOG_LEVEL", "NOOBAA_LOG_COLOR"} {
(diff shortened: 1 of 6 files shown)


pkg/backingstore/reconciler.go
```diff
@@ -1173,9 +1173,18 @@ func (r *Reconciler) needUpdate(pod *corev1.Pod) bool {
...
+	}
+
+	var noobaaLogColEnv = "NOOBAA_LOG_COLOR"
+	var configMapLogCol = r.CoreAppConfig.Data[noobaaLogColEnv]
+	noobaaLogColEnvVar := util.GetEnvVariable(&c.Env, noobaaLogColEnv)
+
+	if configMapLogCol != noobaaLogColEnvVar.Value {
+		r.Logger.Warnf("NOOBAA_LOG_COLOR Env variable change detected: (%v) on the config map (%v)", noobaaLogColEnvVar.Value, configMapLogCol)
+		return true
 	}
```
liranmauda, line 1188: seems like we can make it the same code, and itirate on the variables names. 
something like:
```for _, name := range []string{"NOOBAA_LOG_LEVEL", "NOOBAA_LOG_COLOR"} {```

liranmauda (changes requested):

pkg/backingstore/reconciler.go
aspandey: Done.

PR #1415, opened 2024-08-19
db upgrage: upgrade postgres 15 to postgres 16

vh05:
The Postgres upgrade from 15 to 16 or any later versions can be done by adding the env variable named `POSTGRES_UPGRADE = copy` in the pg db pod. 

More about upgrade process is here: https://www.postgresql.org/docs/current/pgupgrade.html 

In our case, we achieve the upgrade by setting this env var in `noobaa-db-pg` STS.

Once after upgrade is done, we have to remove the `POSTGRES_UPGRADE` env from the STS. Otherwise pg db pod fails to come up because the db pod check PG directory version against upgrade image version and exits if there is no upgrade required and `POSTGRES_UPGRADE` is still provided.

To remove this env, we store the upgrade status in `noobaa-core` CR after the PG upgrade and refer the same status during db reconcile. If the upgrade is already done,  we remove this env from desired sts status.

This will help to upgrade the db smoothly

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -32,6 +32,7 @@ import (
 	nbapis "github.com/noobaa/noobaa-operator/v5/pkg/apis"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
+	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	secv1 "github.com/openshift/api/security/v1"
 	cloudcredsv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
@@ -185,6 +186,7 @@ func init() {
 	Panic(autoscalingv1.AddToScheme(scheme.Scheme))
 	Panic(kedav1alpha1.AddToScheme(scheme.Scheme))
 	Panic(apiregistration.AddToScheme(scheme.Scheme))
+	Panic(configv1.AddToScheme(scheme.Scheme))
 }
 
 // KubeConfig loads kubernetes client config from default locations (flags, user dir, etc)
@@ -236,6 +238,7 @@ func MapperProvider(config *rest.Config, httpClient *http.Client) (meta.RESTMapp
 				g.Name == "autoscaling" ||
 				g.Name == "batch" ||
 				g.Name == "keda.sh" ||
+				g.Name == "config.openshift.io" ||
 				strings.HasSuffix(g.Name, ".k8s.io") {
 				return true
 			}
@@ -1122,9 +1125,11 @@ func GetIBMRegion() (string, error) {
 	return region, nil
 }
 
-// GetAWSRegion parses the region from a node's name
+// GetAWSRegion determines the AWS region from cluster infrastructure or node name
 func GetAWSRegion() (string, error) {
-	// parse the node name to get AWS region according to this:
+	// Determine the AWS region based on cluster infrastructure or node name
+	// If infrastructure details are unavailable, the node's name is parsed to extract the region
+	// Refer to the following for more details:
 	// https://docs.aws.amazon.com/en_pv/vpc/latest/userguide/vpc-dns.html#vpc-dns-hostnames
 	// The list of regions can be found here:
 	// https://docs.aws.amazon.com/general/latest/gr/rande.html
@@ -1164,17 +1169,36 @@ func GetAWSRegion() (string, error) {
 		"af-south-1":     "af-south-1",
 		"il-central-1":   "il-central-1",
 	}
-	nodesList := &corev1.NodeList{}
-	if ok := KubeList(nodesList); !ok || len(nodesList.Items) == 0 {
-		return "", fmt.Errorf("Failed to list kubernetes nodes")
+	var awsRegion string
+	infrastructure := &configv1.Infrastructure{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster",
+		},
 	}
-	nameSplit := strings.Split(nodesList.Items[0].Name, ".")
-	if len(nameSplit) < 2 {
-		return "", fmt.Errorf("Unexpected node name format: %q", nodesList.Items[0].Name)
+	if _, _, err := KubeGet(infrastructure); err != nil {
+		log.Infof("Failed to fetch cluster infrastructure details: %v", err)
 	}
-	awsRegion := mapValidAWSRegions[nameSplit[1]]
+	if infrastructure.Status.PlatformStatus != nil && infrastructure.Status.PlatformStatus.AWS != nil {
+		awsRegion = mapValidAWSRegions[infrastructure.Status.PlatformStatus.AWS.Region]
+	}
+
+	// Parsing the aws region from node name if not fetched from cluster
+	log.Warn("Falling back to parsing the node name as infrastructure details are unavailable or incomplete")
 	if awsRegion == "" {
-		return "", fmt.Errorf("The parsed AWS region is invalid: %q", awsRegion)
+		nodesList := &corev1.NodeList{}
+		if ok := KubeList(nodesList); !ok || len(nodesList.Items) == 0 {
+			log.Infof("Failed to list kubernetes nodes")
+		}
+		nameSplit := strings.Split(nodesList.Items[0].Name, ".")
+		if len(nameSplit) < 2 {
+			log.Infof("Unexpected node name format: %q", nodesList.Items[0].Name)
+		}
+		awsRegion = mapValidAWSRegions[nameSplit[1]]
+	}
+
+	// returning error if not fetched from either cluster or node name
+	if awsRegion == "" {
+		return "", fmt.Errorf("Failed to determine the AWS Region.")
 	}
 	return awsRegion, nil
 }
--- a/deploy/cluster_role.yaml
+++ b/deploy/cluster_role.yaml
@@ -183,3 +183,11 @@ rules:
       - delete
       - update
       - create
+  - apiGroups:
+      - config.openshift.io
+    resources:
+      - infrastructures
+    verbs:
+      - get
+      - list
+      - watch
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -2,7 +2,7 @@ package bundle
 
 const Version = "5.18.0"
 
-const Sha256_deploy_cluster_role_yaml = "3f8118853db73926c4f9d14be84ac8f81833c3a7a94a52ecf1e9ebcf712eee93"
+const Sha256_deploy_cluster_role_yaml = "31fc622ff7fa66617be3895bddcb6cfdb97883b75b20bdb2bf04052bd14221a8"
 
 const File_deploy_cluster_role_yaml = `apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
@@ -189,6 +189,14 @@ rules:
       - delete
       - update
       - create
+  - apiGroups:
+      - config.openshift.io
+    resources:
+      - infrastructures
+    verbs:
+      - get
+      - list
+      - watch
 `
 
 const Sha256_deploy_cluster_role_binding_yaml = "15c78355aefdceaf577bd96b4ae949ae424a3febdc8853be0917cf89a63941fc"


dannyzaken:
@vh05, can you please improve the PR description? What are the changes, technical details on the env used, etc.

vh05:
> @vh05, can you please improve the PR description? What are the changes, technical details on the env used, etc.

sure

deploy/internal/statefulset-postgres-db.yaml
```diff
@@ -26,6 +26,8 @@ spec:
         - name: db
           image: NOOBAA_DB_IMAGE
           env:
+            - name: POSTGRESQL_UPGRADE
```
dannyzaken, line 29: This is only relevant for upgrades. when hardcoded here it will set the env for every new deployment as well.
The `POSTGRES_UPGRADE` env should be set only when necessary

dannyzaken (commented):
I think that instead of deciding on upgrade and applying changes inside `SetDesiredNooBaaDB` it will be more manageable, easy to follow, and less error-prone if you implement an `UpgradeDB` function that will be performed before the DB reconciliation. This method will ensure that everything is upgraded correctly, and will only return nil if the reconciliation is ready to proceed.

deploy/internal/statefulset-postgres-db.yaml
vh05: Yes. This is not needed anymore. The reconcile will detect upgrade status and flag will be set on the fly in memory

vh05:
> I think that instead of deciding on upgrade and applying changes inside `SetDesiredNooBaaDB` it will be more manageable, easy to follow, and less error-prone if you implement an `UpgradeDB` function that will be performed before the DB reconciliation. This method will ensure that everything is upgraded correctly, and will only return nil if the reconciliation is ready to proceed.

Sure Danny. Let me check and do the changes

pkg/system/phase2_creating.go
```diff
@@ -223,6 +223,80 @@ func (r *Reconciler) SetDesiredServiceDBForPostgres() error {
...
+			psql12Image = currentImage
+			util.Logger().Warnf("NOOBAA_PSQL_12_IMAGE is not set. using the current image %s", currentImage)
+		}
+		return psql12Image
+	}
+
+	if r.isPostgresDBUpgradeReqd() {
+		r.NooBaa.Status.BeforeUpgradeDbImage = r.NooBaa.Spec.DBImage
+		r.NooBaa.Spec.DBImage = &options.DBImage
+	}
```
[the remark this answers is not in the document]
vh05: sure

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -365,6 +365,10 @@ type NooBaaStatus struct {
 	// BeforeUpgradeDbImage is the db image used before last db upgrade
 	// +optional
 	BeforeUpgradeDbImage *string `json:"beforeUpgradeDbImage,omitempty"`
+
+	// PostgresMajorVersion is the major version of postgress db image
+	// +optional
+	PostgresMajorVersion int `json:"postgresMajorVersion,omitempty"`
```
dannyzaken, line 371: Use a string with const values like we do with `UpgradePhase`. it will give us more flexibility and is more readable

pkg/system/phase2_creating.go
```diff
@@ -130,6 +131,12 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 			return err
 		}
 	}
+
+	// Once the db is reconciled, upgrade it if it is required
+	if err := r.UpgradeDb(); err != nil {
```
dannyzaken, line 136: Upgrade DB should be performed before `ReconcileDB`

pkg/system/phase2_creating.go
```diff
@@ -152,6 +159,46 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 	return nil
 }
 
+func (r *Reconciler) UpgradeDb() error {
```
dannyzaken, line 162: maybe rename to `UpgradeDbPostgres16`, since it is very specific for that version

pkg/system/phase2_creating.go
```diff
@@ -152,6 +159,46 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
+	var result controllerutil.OperationResult
+	var err error
+
+	// Check the need for upgrade
```
dannyzaken, line 166: I suggest using a `switch` with the different upgrade phases. 
for each phase, check if it's completed and it's safe to change it to the next phase.
`nil` should only be returned from this function after the upgrade is completed successfully, and the rest of the system can be reconciled. Otherwise, return a temp error that indicates the current step of the upgrade.

pkg/system/phase2_creating.go
```diff
@@ -152,6 +159,46 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
+	if r.NooBaa.Status.PostgresUpdatePhase == nbv1.UpgradePhasePrepare {
+		r.NooBaa.Status.PostgresUpdatePhase = nbv1.UpgradePhaseUpgrade
```
dannyzaken, line 168: in the prepare phase you should scale down the core and endpoints, so they will not run during the upgrade. after the DB is upgraded successfully you should restore it back to normal

pkg/system/phase2_creating.go
vh05: This in general should handle any future version of pg upgrade according to `options.DBImage `. The db image upgrade is decided based the current image and the image provided in the options.go.

pkg/apis/noobaa/v1alpha1/noobaa_types.go
vh05: sure

pkg/system/phase2_creating.go
vh05: Sure. Added it in the updated PR

pkg/system/phase2_creating.go
vh05: Yes. Using swtich

pkg/system/phase2_creating.go
vh05: renaming it to `UpgradeDbPostgres16`

pkg/system/phase2_creating.go
vh05: Made changes as suggested

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -365,6 +365,10 @@ type NooBaaStatus struct {
 	// BeforeUpgradeDbImage is the db image used before last db upgrade
 	// +optional
 	BeforeUpgradeDbImage *string `json:"beforeUpgradeDbImage,omitempty"`
+
+	// PostgresMajorVersion is the major version of postgress db image
+	// +optional
+	PostgresMajorVersion string `json:"postgresMajorVersion,omitempty"`
```
dannyzaken, line 371: Please define an alias with const values instead of using string directly

vh05: sure

pkg/system/system.go
```diff
@@ -1129,14 +1129,6 @@ func GetDesiredDBImage(sys *nbv1.NooBaa, currentImage string) string {
 		return psql12Image
 	}
 
-	if sys.Spec.DBImage != nil {
```
dannyzaken, line 1132: This should not be removed. after removal you will always get the hardcoded image

vh05: Restoring it and adding check for desired image in `UpgradeDB`

pkg/system/phase2_creating.go
```diff
@@ -152,6 +157,101 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
 
+func (r *Reconciler) UpgradeDbPostgres16() error {
+	var err error = nil
+
+	// Check the desired image, if it is not as same as db image to upgrade,
+	// don't consider for upgrade
+	// Note: Intention of the below call is to fetch the desired image from Noobaa CR or env var
+	// Passing empty Arg "" to GetDesiredDBImage() in the below case
+	desiredImage := GetDesiredDBImage(r.NooBaa, "")
+	if desiredImage != options.DBImage {
```
dannyzaken, line 168: I don't understand this condition. `options.DBImage` is hardcoded to the upstream DB image.

vh05: Yes, on ds, it will be [registry.redhat.io/rhel9/postgresql-16](http://registry.redhat.io/rhel9/postgresql-16)

vh05: Apologies. I thought we will go for whatever is there in options.DBImage for upgrade. Got clarity that the we have to match the noobaa state and use the image indiacated in the noobaa CR. Updating the PR according to that

pkg/system/phase2_creating.go
```diff
@@ -171,6 +176,101 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
 
+func (r *Reconciler) UpgradeDbPostgres16() error {
+	var err error = nil
+
+	// Check the desired image, if it is not as same as db image to upgrade,
+	// don't consider for upgrade
+	// Note: Intention of the below call is to fetch the desired image from Noobaa CR or env var
+	// Passing empty Arg "" to GetDesiredDBImage() in the below case
+	desiredImage := GetDesiredDBImage(r.NooBaa, "")
+	if !strings.Contains(desiredImage, "postgresql-16") {
```
dannyzaken, line 187: what is the name of the DS image? can we rely on this comparison?

As an alternative for the image name comparison, you can first compare the desired image to the one set in the DB STS:
- if the images are the same, nothing to do
- if they differ AND the current major version is not 16 yet, start the upgrade and modify the upgrade phase

pkg/system/phase2_creating.go
```diff
@@ -171,6 +176,101 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 	return nil
 }
 
+func (r *Reconciler) UpgradeDbPostgres16() error {
+	var err error = nil
+
+	// Check the desired image, if it is not as same as db image to upgrade,
+	// don't consider for upgrade
+	// Note: Intention of the below call is to fetch the desired image from Noobaa CR or env var
+	// Passing empty Arg "" to GetDesiredDBImage() in the below case
```
dannyzaken, line 185: As a first condition, check the postgres major version

pkg/system/phase2_creating.go
vh05: upstream: quay.io/sclorg/postgresql-16-c9s 
downstream: registry.redhat.io/rhel9/postgresql-16

pkg/system/phase2_creating.go
vh05: Sure

pkg/system/phase2_creating.go
vh05: `isPostgresUpgradeReqd()` does 
```

    if the images are the same, nothing to do
    if they differ AND the current major version is not 16 yet, start the upgrade and modify the upgrade phase

```

This check is done in 2 cases
1. case "": This is for the 1st time when we upgrade to the new build, the `PostgresUpdatePhase` is empty
2. case "UpgradePhaseNone": This check is done everytime until the Postgres is upgrade to 16. As soon as desired image is `postgres-16`, we upgrade to `postgres-16` and set the status to upgrade finished.

pkg/system/phase2_creating.go
vh05: The check is added in `isPostgresUpgradeReqd()`

jackyalbo (commented):
Other than my comments, I think the main issue is to check for issues in the middle of the upgrade. I would recommend testing the flows with failing in the middle of every phase. try to delete the pod in the middle of upgrade/cleanning and so on, and make sure we recover from it in a good manner.

pkg/system/phase2_creating.go
```diff
@@ -242,6 +369,94 @@ func (r *Reconciler) SetDesiredServiceDBForPostgres() error {
...
+// SetEndpointsDeploymentReplicas updates the number of replicas on the endpoints deployment
+func (r *Reconciler) setEndpointsDeploymentReplicas(replicas int32) error {
+	r.Logger.Infof("UpgradeMigrateDB:: setting endpoints replica count to %d", replicas)
+	return r.ReconcileObject(r.DeploymentEndpoint, func() error {
+		r.DeploymentEndpoint.Spec.Replicas = &replicas
+		return nil
+	})
+}
+
+// Set env "POSTGRESQL_UPGRADE=copy" in PG sts to iniate the upgrade
```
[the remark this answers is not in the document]
vh05: sure

pkg/system/phase2_creating.go
```diff
@@ -171,6 +176,128 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
+			r.NooBaa.Status.PostgresMajorVersion = nbv1.PostgresMajorVersionV16
+		} else {
+			desiredImage := GetDesiredDBImage(r.NooBaa, "")
+			// Check whether upgrade is reqd, set phase depending the requirement
+			if r.isPostgresUpgradeReqd(desiredImage, stsObj.(*appsv1.StatefulSet)) {
+				r.Logger.Infof("UpgradePostgresDB: upgrading postgres16")
+				r.NooBaa.Status.PostgresUpdatePhase = nbv1.UpgradePhasePrepare
+			} else {
+				if r.NooBaa.Status.PostgresMajorVersion == nbv1.PostgresMajorVersionV16 {
+					r.NooBaa.Status.PostgresMajorVersion = nbv1.PostgresMajorVersionV16
```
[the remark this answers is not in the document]
vh05: noobaa db sts object has all nil values. So fetching db sts object explicitly. This reqd to check what is the current db image inside `isPostgresUpgradeReqd`

pkg/system/phase2_creating.go
```diff
@@ -171,6 +176,128 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 	return nil
 }
 
+func (r *Reconciler) UpgradeDbPostgres16() error {
+	var err error = nil
+
+	if r.NooBaa.Status.PostgresUpdatePhase == nbv1.UpgradePhaseFinished {
```
dannyzaken, line 182: I think that it is possible to have  `PostgresUpdatePhase == nbv1.UpgradePhaseFinished` from the previous upgrade (12->15). you should check for the major version here

pkg/system/phase2_creating.go
```diff
@@ -171,6 +176,128 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
+
+	switch r.NooBaa.Status.PostgresUpdatePhase {
+	case "":
+		// Check whether it is fresh install case
+		sts := util.KubeObject(bundle.File_deploy_internal_statefulset_postgres_db_yaml).(*appsv1.StatefulSet)
+		sts.Name = "noobaa-db-pg"
+		sts.Namespace = r.NooBaa.Namespace
+		stsObj, err := util.KubeGetObject(sts)
+		// Postgres database doesn't exists, no need to upgrade.
+		if err != nil {
```
dannyzaken, line 195: I think that this check should be outside of the `switch`

dannyzaken:
@vh05 @vh05, it will be helpful to add a markdown document in the PR (under `doc` dir) similar to [what we had for 15 upgrade](https://github.com/noobaa/noobaa-operator/blob/master/doc/psql15-upgrade.md). 
In this doc, describe the primary upgrade function in pseudo-code so we can provide feedback on the design before you implement it into the code. It will simplify the implementation once all the steps are defined. 

I believe that such a document will help us both with the implementation and as a reference for debugging later.

pkg/system/phase2_creating.go
```diff
@@ -171,6 +176,128 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
+			} else {
+				r.NooBaa.Status.PostgresUpdatePhase = nbv1.UpgradePhaseNone
+			}
+		}
+	case nbv1.UpgradePhasePrepare:
+		if err := r.setEndpointsDeploymentReplicas(0); err != nil {
+			r.Logger.Errorf("UpgradePostgresDB::got error on endpoints deployment reconcile %v", err)
+			return err
+		}
+		desiredImage := GetDesiredDBImage(r.NooBaa, "")
```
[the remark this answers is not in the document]
vh05: Yes. sure. It was kept outside earlier then I thouht to minimize the scope just to avoid calling it for unnecessary cases. Now we are checking "finished" start of the function itself, its okay to move it next to it

pkg/system/phase2_creating.go
vh05: sure

pkg/system/phase2_creating.go
```diff
@@ -171,6 +176,128 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
+		return nil
+	}
+
+	switch r.NooBaa.Status.PostgresUpdatePhase {
+	case "":
+		// Check whether it is fresh install case
+		sts := util.KubeObject(bundle.File_deploy_internal_statefulset_postgres_db_yaml).(*appsv1.StatefulSet)
+		sts.Name = "noobaa-db-pg"
+		sts.Namespace = r.NooBaa.Namespace
+		stsObj, err := util.KubeGetObject(sts)
```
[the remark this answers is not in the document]
vh05: The db sts object's content was coming nil. Thats the reason fetch kube object from the cluster

pkg/system/phase2_creating.go
vh05: Empty phrase is used for the cases when we move from older version. On older version `upgradephase` flag was not used/updated

pkg/system/phase2_creating.go
```diff
@@ -171,6 +176,128 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
+		r.NooBaa.Status.PostgresUpdatePhase = nbv1.UpgradePhaseUpgrade
+	case nbv1.UpgradePhaseUpgrade:
+		dbPod := &corev1.Pod{}
+		dbPod.Name = "noobaa-db-pg-0"
+		dbPod.Namespace = r.NooBaaPostgresDB.Namespace
+		if !util.KubeCheckQuiet(dbPod) {
+			return nil
+		}
+		// make sure previous step has finished
+		if dbPod.ObjectMeta.DeletionTimestamp != nil {
```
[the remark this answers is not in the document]
vh05: sure.

pkg/system/phase2_creating.go
```diff
@@ -171,6 +176,128 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
...
+		}); err != nil {
+			r.Logger.Infof("UpgradePostgresDB: Cleaning of POSTGRES_UPGRADE env failed: err %v, retry", err)
+			return nil
+		}
+
+		if err := r.setEndpointsDeploymentReplicas(1); err != nil {
+			r.Logger.Errorf("UpgradePostgresDB::got error on endpoints deployment reconcile %v", err)
+			return err
+		}
+		r.NooBaa.Status.PostgresMajorVersion = nbv1.PostgresMajorVersionV16
```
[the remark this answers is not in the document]
vh05: sure

pkg/util/util.go
```diff
@@ -623,6 +623,15 @@ func KubeGet(obj client.Object) (name string, kind string, err error) {
 	return name, kind, err
 }
 
+// KubeGetObject gets a client.Object, fills the given object and returns
+func KubeGetObject(obj client.Object) (client.Object, error) {
```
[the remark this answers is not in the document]
vh05: Yes. Removing

pkg/system/phase2_creating.go
```diff
@@ -242,6 +369,94 @@ func (r *Reconciler) SetDesiredServiceDBForPostgres() error {
...
+				c.Resources = *r.NooBaa.Spec.DBResources
+			}
+
+			c.Lifecycle = &corev1.Lifecycle{
+				PreStop: &corev1.LifecycleHandler{
+					Exec: &corev1.ExecAction{
+						Command: []string{"/bin/sh", "-c", "pg_ctl -D /var/lib/pgsql/data/userdata/ -w -t 60 -m fast stop"},
+					},
+				},
+			}
```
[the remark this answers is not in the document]
vh05: Not really sure of this. PG upgrade takes care of this. Just doing it explicitly that db upgrade can start safely. Your opinion ?

pkg/system/phase2_creating.go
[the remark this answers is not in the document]
jackyalbo: I just think we should check as much as we can before moving to next phase. you set the env POSTGRES_COPY before the upgrade and the pod doen't come up, you will try this phase again and set it again, only when the pod is running with no errors we will assume the phase was successful. moving to the next phase will make us finish the upgrade and mark it as done while the upgrade completely failed.

pkg/system/phase2_creating.go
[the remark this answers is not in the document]
jackyalbo: we should at least fail the upgrade no? add an event? Finished should be set when we tried starting with v16 and to pod starts with no issues.

pkg/system/phase2_creating.go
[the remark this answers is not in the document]
vh05: Removed this part

vh05:
> @vh05 @vh05, it will be helpful to add a markdown document in the PR (under `doc` dir) similar to [what we had for 15 upgrade](https://github.com/noobaa/noobaa-operator/blob/master/doc/psql15-upgrade.md). In this doc, describe the primary upgrade function in pseudo-code so we can provide feedback on the design before you implement it into the code. It will simplify the implementation once all the steps are defined.
> 
> I believe that such a document will help us both with the implementation and as a reference for debugging later.

Here is the PR for doc:
https://github.com/noobaa/noobaa-operator/pull/1496

pkg/system/phase2_creating.go
vh05: Moved this part out of switch

pkg/system/phase2_creating.go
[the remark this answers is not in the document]
vh05: We cannot fail the upgrade. Once the env to do the upgrade is set, the postgres will take the responsibility of upgrading. For any further errors we need to track the logs of the db pod

PR #1418, opened 2024-08-22
Added a validation check for quota config in bucket

achouhan09:
### Explain the changes
1. Currently, bucket quota is not working as expected it is failing to update quota when we pass max-size in other then bytes format. We have added a validation check for the quota and applied some fixes to work with any format of max-size.

1. Improvement: #1252 

### Testing Instructions:
1. Update the bucket quota using various formats (Gi, Ti etc.) and verify the quota is correctly reflected in binary format using cli commands.
2. Test invalid formats to ensure the system returns an error and does not update the quota.

### DIFF
--- a/pkg/bucket/bucket.go
+++ b/pkg/bucket/bucket.go
@@ -9,6 +9,7 @@ import (
 	"github.com/noobaa/noobaa-operator/v5/pkg/options"
 	"github.com/noobaa/noobaa-operator/v5/pkg/system"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+	"k8s.io/apimachinery/pkg/api/resource"
 
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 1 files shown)


pkg/bucket/bucket.go
```diff
@@ -113,17 +114,17 @@ func RunCreate(cmd *cobra.Command, args []string) {
...
 	if err != nil {
-		log.Fatal(err)
+		log.Fatalf(`❌ Could not create bucket "%q" validation failed %q`, bucketName, err)
 	}
 
-	quota, updateQuota := prepareQuotaConfig(maxSize, maxObjects)
-	if updateQuota {
+	err = nbClient.CreateBucketAPI(nb.CreateBucketParams{Name: bucketName, Tiering: tierName, ForceMd5Etag: forceMd5Etag})
+	if err == nil && updateQuota {
 		err = nbClient.UpdateBucketAPI(nb.CreateBucketParams{Name: bucketName, Quota: &quota})
```
achouhan09, line 124: Not sure why we havn't added quota in create_bucket api, so we need to call update_bucket  api with quota again here.

@vh05 @liranmauda Anything you want to add?

vh05: I think we have not introduce there because we fully not supporting quota. @liranmauda @dannyzaken Please correct me if wrong

pkg/bucket/bucket.go
```diff
@@ -239,25 +243,30 @@ func RunList(cmd *cobra.Command, args []string) {
...
+
 	if bucketMaxSize > 0 {
 		f, u := nb.GetBytesAndUnits(bucketMaxSize, 2)
 		quota.Size = &nb.SizeQuotaConfig{Value: f, Unit: u}
 	}
 	if bucketMaxObjects > 0 {
 		quota.Quantity = &nb.QuantityQuotaConfig{Value: int(bucketMaxObjects)}
 	}
-	return quota, updateQuota
+	return quota, true, err
```
vh05, line 271: 2 return values quota, update_quota are enough. Is n't it.

achouhan09: yes we can return `quota, update_quota=true, err=nil` (as there is no error) here. Thanks

pkg/bucket/bucket.go
```diff
@@ -248,25 +248,24 @@ func prepareQuotaConfig(bucketName string, maxSize string, maxObjects string) (n
...
 
-	if bucketMaxSize > 0 {
-		f, u := nb.GetBytesAndUnits(bucketMaxSize, 2)
-		quota.Size = &nb.SizeQuotaConfig{Value: f, Unit: u}
-	}
-	if bucketMaxObjects > 0 {
-		quota.Quantity = &nb.QuantityQuotaConfig{Value: int(bucketMaxObjects)}
-	}
-	return quota, true, err
+	return quota, true, nil
```
vh05, line 270: Am I checking the right commit ? Still can see 3 values

achouhan09: Updated the code. Thanks

vh05 (approved):
/lgtm

vec-tr (changes requested):
Please squal and push the changes.

vh05 (changes requested):
Please squash and push the changes

vh05 (approved):
\lgtm

PR #1215, opened 2023-09-07
Added support for force_md5_etag for bucket and account

achouhan09:
### Explain the changes
1. Added support for force_md5_etag in operator.

PR #1430, opened 2024-09-09
Updated forceMD5Etag to boolean pointer type

achouhan09:
### Explain the changes
1. Updated forceMd5Etag to a Boolean Pointer Type (*bool): The forceMd5Etag flag is now handled as a pointer to a boolean (*bool) to explicitly track whether the flag was provided by the user:

- If the flag was set, forceMd5Etag points to a true or false value. 
- If the flag was not set, forceMd5Etag remains nil, preventing unintended changes to the ForceMd5Etag field.

1. Improvement: #1215 
2. BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2312533

### Testing Instructions:
1. Try creating and updating the bucket using cli.
2. Try creating and updating the NoobaaAccount using cli and yaml both.

### DIFF
--- a/pkg/bucket/bucket.go
+++ b/pkg/bucket/bucket.go
@@ -109,11 +109,14 @@ func RunCreate(cmd *cobra.Command, args []string) {
 		log.Fatal(fmt.Errorf("CreateTieringStructure for PlacementPolicy failed to create policy %q with error: %v", tierName, err))
 	}
 
-	forceMd5Etag, _ := cmd.Flags().GetBool("force_md5_etag")
+	forceMd5EtagPtr, err := util.GetBoolFlagPtr(cmd, "force_md5_etag")
+	if err != nil {
+		log.Fatal(err)
+	}
 	maxSize, _ := cmd.Flags().GetString("max-size")
 	maxObjects, _ := cmd.Flags().GetString("max-objects")
 
-	err = nbClient.CreateBucketAPI(nb.CreateBucketParams{Name: bucketName, Tiering: tierName, ForceMd5Etag: forceMd5Etag})
+	err = nbClient.CreateBucketAPI(nb.CreateBucketParams{Name: bucketName, Tiering: tierName, ForceMd5Etag: forceMd5EtagPtr})
 	if err != nil {
 		log.Fatal(err)
 	}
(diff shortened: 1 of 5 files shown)


jackyalbo:
@achouhan09 I remember @dannyzaken had issues when changing API schemas between versions. Did you test upgrade where there was a noobaaAccount with true/false set and if it worked after moving to the new schema in the later version? 
Try doing a whole upgrade procedure of running with the old CRD and then changing to the new one and see if everything still works(can maybe check with the cli upgrade command - if it supports upgrading the CRDs)

pkg/apis/noobaa/v1alpha1/noobaaaccount_types.go
```diff
@@ -69,7 +69,7 @@ type NooBaaAccountSpec struct {
 
 	// ForceMd5Etag specifies whether MD5 Etags should be calculated for the account or not
 	// +optional
-	ForceMd5Etag *bool `json:"force_md5_etag,omitempty"`
+	ForceMd5Etag bool `json:"force_md5_etag,omitempty"`
```
dannyzaken, line 72: Why is this change necessary? 
With this code change, the user must set a value for `ForceMd5Etag`. If the user leaves this field empty, it will implicitly be set to `false` and override any other setting\default in noobaa-core.

achouhan09: Done the required changes as per discussed in slack, everything working as expected. Thanks

dannyzaken (changes requested):

pkg/noobaaaccount/noobaaaccount.go
```diff
@@ -272,11 +272,7 @@ func RunUpdate(cmd *cobra.Command, args []string) {
 	name := args[0]
 
 	newDefaultResource := util.GetFlagStringOrPrompt(cmd, "new_default_resource")
-	var forceMd5EtagPtr *bool = nil
-	if cmd.Flags().Changed("force_md5_etag") {
-		forceMd5Etag, _ := cmd.Flags().GetBool("force_md5_etag")
-		forceMd5EtagPtr = &forceMd5Etag
-	}
+	forceMd5Etag, _ := cmd.Flags().GetBool("force_md5_etag")
```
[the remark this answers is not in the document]
achouhan09: Make sense, Thanks.

pkg/bucket/bucket.go
```diff
@@ -187,7 +194,7 @@ func RunStatus(cmd *cobra.Command, args []string) {
 	}
 	fmt.Printf("  %-22s : %s\n", "Type", b.BucketType)
 	fmt.Printf("  %-22s : %s\n", "Mode", b.Mode)
-	fmt.Printf("  %-22s : %t\n", "Force Md5 Etag", b.ForceMd5Etag)
+	fmt.Printf("  %-22s : %t\n", "Force Md5 Etag", *b.ForceMd5Etag)
```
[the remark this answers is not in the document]
achouhan09: Done, Thanks

pkg/util/util.go
```diff
@@ -1238,6 +1238,20 @@ func GetFlagStringOrPromptPassword(cmd *cobra.Command, flag string) string {
 	return str
 }
 
+// GetBoolFlagPtr returns a pointer to the boolean flag value if set, or nil if not set
+func GetBoolFlagPtr(cmd *cobra.Command, flag string) (*bool, error) {
+	flagVal, err := cmd.Flags().GetBool(flag)
+	if err != nil {
+		return nil, err
+	}
```
achouhan09, line 1246: If I am adding this condition, the admission-tests are failing with the below error:
`time="2024-09-20T12:55:53Z" level=fatal msg="flag accessed but not defined: force_md5_etag"`

I think we need not to handle this condition😅 just like other functions in util.go, also without it the tests are passing. As we are returning nil value from the function if the flag is not found.
@dannyzaken  wdyt?

dannyzaken: What is the root cause of this error? I am not sure ignoring the error is the answer

pkg/util/util.go
```diff
@@ -1238,6 +1238,17 @@ func GetFlagStringOrPromptPassword(cmd *cobra.Command, flag string) string {
 	return str
 }
 
+// GetBoolFlagPtr returns a pointer to the boolean flag value if set, or nil if not set
+func GetBoolFlagPtr(cmd *cobra.Command, flag string) *bool {
+	flagVal, _ := cmd.Flags().GetBool(flag)
```
dannyzaken, line 1243: This should be inside the if. I think this is the cause for the tests failing

pkg/util/util.go
```diff
@@ -1238,6 +1238,17 @@ func GetFlagStringOrPromptPassword(cmd *cobra.Command, flag string) string {
...
+
+	if cmd.Flags().Changed(flag) {
+		return &flagVal
+	}
+
+	return nil
+}
```
dannyzaken, line 1250: The `GetBool` should be inside the `if` statement. I think this is why the tests failed. 
Also, we should not ignore the error. Either return it or handle it in another way (log, etc.).

```suggestion
	if cmd.Flags().Changed(flag) {
	        flagVal, err := cmd.Flags().GetBool(flag)
        	if err != nil {
        	// handle error
 	        }
        	return &flagVal
	}
	return nil
}
```

dannyzaken (approved):

PR #1440, opened 2024-09-12
Added a CLI sub-command in noobaa diagnostics to print proxy details

achouhan09:
### Explain the changes
1. The CLI command checks and logs the values of essential environment variables related to proxy configuration (HTTP_PROXY, HTTPS_PROXY, NO_PROXY). It performs the following actions:

    Checks: Verifies if these proxy environment variables are set and not empty.

This implementation currently focuses on proxy environment variables but is designed with future expansion in mind. Additional checks and diagnostics will be introduced in the future.

Screenshot from 2024-09-17 17-48-53

1. BZ: https://bugzilla.redhat.com/show_bug.cgi?id=2302136

### Testing Instructions:
1. Try running noobaa cli  command `noobaa diagnostics report` to check the proxy status. 

**Note:** Use the command `kubectl set env deployment/noobaa-operator -c noobaa-operator HTTP_PROXY=localhost:8080` to update the proxy vars.

### DIFF
--- a/pkg/diagnostics/diagnostics.go
+++ b/pkg/diagnostics/diagnostics.go
@@ -24,6 +24,7 @@ func Cmd() *cobra.Command {
 		CmdCollect(),
 		CmdDbDump(),
 		CmdAnalyze(),
+		CmdReport(),
 	)
 	return cmd
 }
@@ -64,6 +65,17 @@ func CmdAnalyze() *cobra.Command {
 		CmdAnalyzeNamespaceStore(),
 		CmdAnalyzeResources(),
 	)
+	return cmd
+}
+
+// CmdReport returns a CLI command
+func CmdReport() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "report",
+		Short: "Run reports of the status and setup",
+		Run:   RunReport,
+		Args:  cobra.NoArgs,
+	}
 	return cmd
 }
 
--- a/pkg/diagnostics/report.go
+++ b/pkg/diagnostics/report.go
@@ -1,0 +1,43 @@
+package diagnostics
+
+import (
+	"fmt"
+
+	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
+	"github.com/noobaa/noobaa-operator/v5/pkg/options"
+	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+	"github.com/spf13/cobra"
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+// RunReport runs a CLI command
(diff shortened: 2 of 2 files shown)


pkg/diagnostics/diagnostics.go
```diff
@@ -67,6 +68,17 @@ func CmdAnalyze() *cobra.Command {
 	return cmd
 }
 
+// CmdReport returns a CLI command
+func CmdReport() *cobra.Command {
```
shirady, line 72: I think it can be easier to add commands if the proxy status is a subcommand, I mean:
`noobaa diagnostics report <sub-command>`.
As I understand now `noobaa diagnostics report` is a single command to get the proxy details only.

liranmauda: The report will host all the flow we want to report.
We should not cut it with subcommands.

achouhan09: Updated the description. Thanks

shirady (approved):
LGTM

pkg/diagnostics/diagnostics.go
```diff
@@ -67,6 +68,17 @@ func CmdAnalyze() *cobra.Command {
...
+	cmd := &cobra.Command{
+		Use:   "report",
+		Short: "Reports the proxy status",
```
liranmauda, line 75: The command will not only run a report of the proxy, but it will be extended to other issues as well.
```suggestion
		Short: "Run reports of the status and setup",
```

pkg/diagnostics/report.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
+	"github.com/noobaa/noobaa-operator/v5/pkg/options"
+	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+	"github.com/spf13/cobra"
+	appsv1 "k8s.io/api/apps/v1"
+)
+
+// RunReport runs a CLI command
+func RunReport(cmd *cobra.Command, args []string) {
```
liranmauda, line 12: The function should run a bunch of functions that will diagnose each topic
We should get the proxy check part into a different function and call it here.

liranmauda (changes requested):

pkg/diagnostics/diagnostics.go
achouhan09: updated, Thanks

pkg/diagnostics/report.go
achouhan09: Done, Thanks

liranmauda (approved):

PR #1445, opened 2024-09-25
Printing overrided  vars in  cli command `noobaa diagnostics report`

achouhan09:
### Explain the changes
1. Printing overriden  vars in  cli command `noobaa diagnostics report`.

Output:
Screenshot from 2024-09-25 16-41-49

### Testing Instructions:
1. After installing noobaa try running `noobaa diagnostics report` cli command to check the status.

### DIFF
--- a/pkg/diagnostics/report.go
+++ b/pkg/diagnostics/report.go
@@ -2,6 +2,7 @@ package diagnostics
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
 	"github.com/noobaa/noobaa-operator/v5/pkg/options"
@@ -12,26 +13,32 @@ import (
 
 // RunReport runs a CLI command
 func RunReport(cmd *cobra.Command, args []string) {
-
-	// retrieving the status of proxy environment variables
-	proxyStatus()
-
-	// TODO: Add support for additional features
-}
-
-// proxyStatus returns the status of the environment variables: HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
-func proxyStatus() {
 	log := util.Logger()
 
-	log.Print("⏳ Retrieving proxy environment variable details...\n")
+	// Fetching coreApp configurations
 	coreApp := util.KubeObject(bundle.File_deploy_internal_statefulset_core_yaml).(*appsv1.StatefulSet)
 	coreApp.Namespace = options.Namespace
 	if !util.KubeCheck(coreApp) {
 		log.Fatalf(`❌ Could not get core StatefulSet %q in Namespace %q`,
 			coreApp.Name, coreApp.Namespace)
 	}
+	fmt.Println("")
 
-	fmt.Print("\nProxy Environment Variables Check:\n----------------------------------\n")
+	// retrieving the status of proxy environment variables
+	proxyStatus(coreApp)
+
+	// retrieving the overriden env variables using `CONFIG_JS_` prefix
+	overridenEnvVar(coreApp)
+
+	// TODO: Add support for additional features
+}
+
+// proxyStatus returns the status of the environment variables: HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
+func proxyStatus(coreApp *appsv1.StatefulSet) {
+	log := util.Logger()
+
+	log.Print("⏳ Retrieving proxy environment variable details...\n")
+	fmt.Print("Proxy Environment Variables Check:\n----------------------------------\n")
 	for _, proxyName := range []string{"HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY"} {
 		envVar := util.GetEnvVariable(&coreApp.Spec.Template.Spec.Containers[0].Env, proxyName)
 		if envVar != nil && envVar.Value != "" {
@@ -40,4 +47,19 @@ func proxyStatus() {
 			fmt.Printf("	❌ %-12s : not set or empty.\n", proxyName)
 		}
 	}
+	fmt.Println("")
+}
+
+// overridedEnvVar retrieves and displays overridden environment variables with the prefix `CONFIG_JS_` from the noobaa-core-0 pod
+func overridenEnvVar(coreApp *appsv1.StatefulSet) {
+	log := util.Logger()
+
+	log.Print("⏳ Retrieving overridden environment variables from the pod `noobaa-core-0`...\n")
+	fmt.Print("Overriden Environment Variables Check:\n----------------------------------\n")
+	for _, envVar := range coreApp.Spec.Template.Spec.Containers[0].Env {
+		if strings.HasPrefix(envVar.Name, "CONFIG_JS_") {
+			fmt.Printf("    	✔ %-12s : %s\n", envVar.Name, envVar.Value)
+		}
+	}
+	fmt.Println("")
 }
--- a/pkg/diagnostics/report.go
+++ b/pkg/diagnostics/report.go
@@ -22,13 +22,21 @@ func RunReport(cmd *cobra.Command, args []string) {
 		log.Fatalf(`❌ Could not get core StatefulSet %q in Namespace %q`,
 			coreApp.Name, coreApp.Namespace)
 	}
+
+	// Fetching endpoint configurations
+	endpointApp := util.KubeObject(bundle.File_deploy_internal_deployment_endpoint_yaml).(*appsv1.Deployment)
+	endpointApp.Namespace = options.Namespace
+	if !util.KubeCheck(endpointApp) {
+		log.Fatalf(`❌ Could not get endpoint Deployment %q in Namespace %q`,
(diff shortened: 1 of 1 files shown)


pkg/diagnostics/report.go
```diff
@@ -40,4 +47,19 @@ func proxyStatus() {
...
 		}
 	}
+	fmt.Println("")
+}
+
+// overridedEnvVar retrieves and displays overridden environment variables with the prefix `CONFIG_JS_` from the noobaa-core-0 pod
+func overridenEnvVar(coreApp *appsv1.StatefulSet) {
+	log := util.Logger()
+
+	log.Print("⏳ Retrieving overridden environment variables from the pod `noobaa-core-0`...\n")
```
shirady, line 57: In the first printing, you wrote "Retrieving proxy environment variable details" and here it is "Retrieving overridden environment variables from the pod `noobaa-core-0`", I think it should look the same (either with the part of "from... " or without it - anyway, both of them are retrieved from the StatefulSet.

shirady (approved):
LGTM

shirady:
@achouhan09 Could you please update the testing instructions in the PR?

pkg/diagnostics/report.go
achouhan09: I will update it. Thanks

achouhan09:
> I was just wondering - is there a reason we're only checking the core pod for overrides? I think it'd be very useful and important to check one of the endpoints as well

@Neon-White Yes, make sense. I have added a check for endpoint as well. Thanks

pkg/diagnostics/report.go
```diff
@@ -40,4 +55,28 @@ func proxyStatus() {
...
+		}
+	}
+	fmt.Println("")
+
+	fmt.Print("Overriden Environment Variables Check (ENDPOINT):\n----------------------------------\n")
+	for _, envVar := range endpointApp.Spec.Template.Spec.Containers[0].Env {
+		if strings.HasPrefix(envVar.Name, "CONFIG_JS_") {
+			fmt.Printf("    	✔ %-20s : %s\n", envVar.Name, envVar.Value)
+		}
+	}
```
liranmauda, line 80: When there are no envs with `CONFIG_JS_` prefix we should print it out.

achouhan09: sure, Thanks.

pkg/diagnostics/report.go
```diff
@@ -40,4 +55,38 @@ func proxyStatus() {
...
+	fmt.Println("")
+}
+
+// overridedEnvVar retrieves and displays overridden environment variables with the prefix `CONFIG_JS_` from the noobaa-core-0 pod
+func overridenEnvVar(coreApp *appsv1.StatefulSet, endpointApp *appsv1.Deployment) {
+	log := util.Logger()
+
+	log.Print("⏳ Retrieving overridden environment variable details...\n")
+
+	fmt.Print("Overriden Environment Variables Check (NOOBAA-CORE):\n----------------------------------\n")
```
liranmauda, line 67:
```suggestion
	fmt.Print("Overridden Environment Variables Check (NOOBAA-CORE):\n----------------------------------\n")
```

pkg/diagnostics/report.go
```diff
@@ -40,4 +55,38 @@ func proxyStatus() {
...
+			fmt.Printf("    	✔ %s : %s\n", envVar.Name, envVar.Value)
+			foundCoreEnv = true
+		}
+	}
+	if !foundCoreEnv {
+		fmt.Printf("	❌ No overriden environment variables found.")
+	}
+	fmt.Println("")
+
+	fmt.Print("Overriden Environment Variables Check (ENDPOINT):\n----------------------------------\n")
```
liranmauda, line 80:
```suggestion
	fmt.Print("Overridden Environment Variables Check (ENDPOINT):\n----------------------------------\n")
```

pkg/diagnostics/report.go
```diff
@@ -40,4 +55,38 @@ func proxyStatus() {
...
+	fmt.Print("Overriden Environment Variables Check (NOOBAA-CORE):\n----------------------------------\n")
+	foundCoreEnv := false
+	for _, envVar := range coreApp.Spec.Template.Spec.Containers[0].Env {
+		if strings.HasPrefix(envVar.Name, "CONFIG_JS_") {
+			fmt.Printf("    	✔ %s : %s\n", envVar.Name, envVar.Value)
+			foundCoreEnv = true
+		}
+	}
+	if !foundCoreEnv {
+		fmt.Printf("	❌ No overriden environment variables found.")
```
liranmauda, line 76:
```suggestion
		fmt.Printf("	❌ No overridden environment variables found.")
```

pkg/diagnostics/report.go
```diff
@@ -40,4 +55,38 @@ func proxyStatus() {
...
+	fmt.Print("Overriden Environment Variables Check (ENDPOINT):\n----------------------------------\n")
+	foundEndpointEnv := false
+	for _, envVar := range endpointApp.Spec.Template.Spec.Containers[0].Env {
+		if strings.HasPrefix(envVar.Name, "CONFIG_JS_") {
+			fmt.Printf("    	✔ %s : %s\n", envVar.Name, envVar.Value)
+			foundEndpointEnv = true
+		}
+	}
+	if !foundEndpointEnv {
+		fmt.Printf("	❌ No overriden environment variables found.")
```
liranmauda, line 89:
```suggestion
		fmt.Printf("	❌ No overridden environment variables found.")
```

liranmauda:
> I was just wondering - is there a reason we're only checking the core pod for overrides? I think it'd be very useful and important to check one of the endpoints as well


> > I was just wondering - is there a reason we're only checking the core pod for overrides? I think it'd be very useful and important to check one of the endpoints as well
> 
> @Neon-White Yes, make sense. I have added a check for endpoint as well. Thanks

@Neon-White @achouhan09 Do we also need to add the proxy check on the endpoint?

achouhan09:
> > I was just wondering - is there a reason we're only checking the core pod for overrides? I think it'd be very useful and important to check one of the endpoints as well
> 
> > > I was just wondering - is there a reason we're only checking the core pod for overrides? I think it'd be very useful and important to check one of the endpoints as well
> > 
> > 
> > @Neon-White Yes, make sense. I have added a check for endpoint as well. Thanks
> 
> @Neon-White @achouhan09 Do we also need to add the proxy check on the endpoint?

Not sure, but for proxy changes I need to update the operator deployment and then it will get reflected to other pods, so for proxy status we can check for any one of the core/endpoint is updated or not. @Neon-White pls correct me if I am wrong.

liranmauda:
> @achouhan09 @liranmauda "[After configuring the cluster-wide proxy, the Operator Lifecycle Manager (OLM) triggers automatic updates to all of the deployed Operators with the new contents of the HTTP_PROXY, HTTPS_PROXY, and NO_PROXY environment variables.](https://docs.openshift.com/container-platform/4.14/networking/external_dns_operator/nw-configuring-cluster-wide-egress-proxy.html)"
> 
> _Ideally_, this is how users would configure their proxies, which would be a reliable way to detect it. Depending on how proactive we want to be, it can still be valuable to check the core and endpoint individually, on top of the operator, just in case the customer applied a proxy on them in particular.

Great!
@achouhan09 lets merge this PR, and then create a new PR for the proxy to check on both core, and operator

pkg/diagnostics/report.go
```diff
@@ -40,4 +55,38 @@ func proxyStatus() {
...
+	foundEndpointEnv := false
+	for _, envVar := range endpointApp.Spec.Template.Spec.Containers[0].Env {
+		if strings.HasPrefix(envVar.Name, "CONFIG_JS_") {
+			fmt.Printf("    	✔ %s : %s\n", envVar.Name, envVar.Value)
+			foundEndpointEnv = true
+		}
+	}
+	if !foundEndpointEnv {
+		fmt.Printf("	❌ No overridden environment variables found.")
+	}
```
liranmauda, line 90: This duplication can be a function that gets a kubernetes object and is called twice, once for core sts and a second time for endpoint deployment.
@achouhan09 Create a new PR that consolidates those into a single function.

pkg/diagnostics/report.go
```diff
@@ -12,26 +13,40 @@ import (
...
+			coreApp.Name, coreApp.Namespace)
+	}
+
+	// Fetching endpoint configurations
+	endpointApp := util.KubeObject(bundle.File_deploy_internal_deployment_endpoint_yaml).(*appsv1.Deployment)
+	endpointApp.Namespace = options.Namespace
+	if !util.KubeCheck(endpointApp) {
+		log.Fatalf(`❌ Could not get endpoint Deployment %q in Namespace %q`,
+			endpointApp.Name, endpointApp.Namespace)
+	}
```
liranmauda, line 32: This duplication can be a function that gets a kubernetes object and is called twice, once for core sts and a second time for endpoint deployment.
@achouhan09 Create a new PR that consolidates those into a single function.

PR #1462, opened 2024-10-14
Added ARN checks in the diagnostics report

achouhan09:
### Explain the changes
1. Added ARN checks in the diagnostics report for both backingstores and namespacestores having AWS STS ARN string.
Screenshot from 2024-10-30 19-00-55

 
1. 

### Testing Instructions:
1. Try running command `noobaa diagnostics report` to see the output.

### DIFF
--- a/pkg/diagnostics/report.go
+++ b/pkg/diagnostics/report.go
@@ -4,12 +4,16 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/aws/aws-sdk-go/aws/arn"
+	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
 	"github.com/noobaa/noobaa-operator/v5/pkg/options"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
 	"github.com/spf13/cobra"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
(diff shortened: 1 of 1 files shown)


pkg/diagnostics/report.go
```diff
@@ -102,3 +179,8 @@ func printOverriddenEnvVar(appName string, envVars []corev1.EnvVar) {
 	}
 	fmt.Println("")
 }
+
+// isValidArn is a function to validate the ARN format for an s3 buckets
+func isValidArn(arn *string) bool {
+	return strings.HasPrefix(*arn, "arn:aws:s3:::") && len(*arn) > len("arn:aws:s3:::")
```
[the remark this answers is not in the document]
achouhan09: Make sense, Done. Thanks

pkg/diagnostics/report.go
```diff
@@ -102,3 +151,28 @@ func printOverriddenEnvVar(appName string, envVars []corev1.EnvVar) {
 	}
 	fmt.Println("")
 }
+
+// isValidArn is a function to validate the ARN format for an s3 buckets
+func isValidArn(arn *string) bool {
+	return strings.HasPrefix(*arn, "arn:aws:s3::") && len(*arn) > len("arn:aws:s3::")
+}
```
shirady, line 158: I think you could use the built-in AWS SDK check instead of writing it.
When I worked on AWS STS I checked the received ARN of a role: `!arn.IsARN(roleARN)` (you will need to add the import: `"github.com/aws/aws-sdk-go/aws/arn"`), here:

https://github.com/noobaa/noobaa-operator/blob/c6727ce97d1647e47f8cc1deeeff3d2e48e71752/pkg/system/phase2_creating.go#L704-L716

shirady:
@achouhan09, Did you want to validate only AWS STS ARN?
If that is so, then the example in the PR description is confusing because it is not exactly the case.

achouhan09:
> @achouhan09, Did you want to validate only AWS STS ARN? If that is so, then the example in the PR description is confusing because it is not exactly the case.

@liranmauda I am bit confused about AWS ARN check here. What are the requirements for this check?
Do we need to validate AWS STS ARN only for the stores here and print the status? Also, I think checking for bucket or object ARN doesn't make sense.

pkg/diagnostics/report.go
achouhan09: Done, Thanks

shirady (approved):
LGTM

PR #1467, opened 2024-10-31
Bucket Notifcations

alphaprinz:
-New field notificationsPVC in noobaa that is mounted to endpoints and core
-Secrets containing info on how to connect to remote server are mounted to core

### Explain the changes
1. Implement a way to specify a pvc to use for notification files.
2. Implement a way to specify secrets with information of how to connect to external notification target.

### Testing Instructions:
1. Get a pvc.
2. Specify pvc name in notificationsPVC in noobaa crd under spec.
3. Create a secret from the connect file, label it with app=noobaa and noobaa=notifications

- [ ] Doc added/updated
- [ ] Tests added


pkg/system/phase2_creating.go
```diff
@@ -519,6 +529,51 @@ func (r *Reconciler) SetDesiredCoreApp() error {
...
+						},
+					},
+				}}
+				util.MergeVolumeList(&podSpec.Volumes, &notificationVolumes)
+
+				//find secrets that tell us how to connect to remote notifications servers,
+				//mount them so core can read them
+				notificatoinSecrets := &corev1.SecretList{}
+				noobaaNotifSelector, _ := labels.Parse("app=noobaa,noobaa=notifications")
+				util.KubeList(notificatoinSecrets, &client.ListOptions{Namespace: options.Namespace, LabelSelector: noobaaNotifSelector})
```
jackyalbo, line 554: What happens if we don't find any secret? Do we have a default? we should reject? warn?

pkg/system/phase2_creating.go
```diff
@@ -450,6 +450,16 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 			}
 		}
 	}
+
+	if r.NooBaa.Spec.NotificationsPVC != nil {
```
jackyalbo, line 454: I think we should support something similar to what we have for bucket logging. We should create a pvc in case we are in ODF. Also, don't think we should have two if any one of them is turned on. I think we should have a segment like bucketNotifications and inside we should have enabled, yes or no and an optional pvc in case bucket logging is not enabled and we are not in ODF.

dannyzaken: I agree with Jacky. we should probably consolidate the PVCs, and handle it in a similar way

alphaprinz: -Regarding PVC in ODF - OK, I will copy bucket logging notification behavior in ODF case.
-Regarding single PVC - are you sure using a single PVC is better, performance-wise or user-experience-wise? If so, should I rename it (in our code) BucketLoggingPVC? Sth like "PersistentLogsPVC"?

pkg/system/phase2_creating.go
alphaprinz: Using default doesn't make sense, because user at least needs to specify url for remote server and its protocol (http/kafka).
I can look into going into reject, but that can be done in operator only if there are no secrets at all. It would still be possible to be missing some of the secrets and operator won't have a way to know that.

pkg/system/phase2_creating.go
alphaprinz: For using a single PVC for both bucket logging and bucket notifications, the change that makes most sense to me is that we remove pvc name from both of them and add a new field in noobaa crd for a "persistent files pvc", which both will use. Otherwise the features parameters are mixed.
Let me know what you think or if you have a different structure in mind.

pkg/system/phase2_creating.go
jackyalbo: I think we should give the user as much info as we can. If we don't have any secrets and we know the feature won't work this way - we should at least write a warning in the logs (I think we should also Reject)
It can be nice to also check that the secret has the needed fields we expect and maybe add some documentation about how the secret is supposed to look.

jackyalbo: maybe the user should give us the specific secret that he wants us to use and we won't just list all the secrets? WDYT?

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -227,6 +227,11 @@ type NooBaaSpec struct {
 	// BucketLogging sets the configuration for bucket logging
 	// +optional
 	BucketLogging BucketLoggingSpec `json:"bucketLogging,omitempty"`
+
+	// NotificationsPVC (optional) specifies the name of the Persistent Volume Claim (PVC) to be used
+	// for notifications persistent files.
+	// +optional
+	NotificationsPVC *string `json:"notificationsPVC,omitempty"`
```
dannyzaken, line 234: Following a discussion with Amit and Jacky:
* Add `BucketNotificationsSpec` that contains `enable`, optionl `pvc`, array of secrets.
* we will use separate PVCs for bucket logging and bucket notifications.
* Reuse common code between notifications and logging.

alphaprinz: Changes are the two new commits from this week, LMKWYT

dannyzaken (commented):
I added one small comment, other than that LGTM. 
I will let Jacky give the final approval as he is more familiar with this code. 
Don't forget to squash your commits.

pkg/system/phase2_creating.go
```diff
@@ -1163,56 +1229,60 @@ func (r *Reconciler) ReconcileDBConfigMap(cm *corev1.ConfigMap, desiredFunc func
 	return r.isObjectUpdated(result), nil
 }
 
-// ReconcileODFBucketLoggingPVC ensures the bucket logging PVC is properly configured for 'guaranteed' logging
```
jackyalbo, line 1166: Add a new comment instead of the one deleted

pkg/system/phase2_creating.go
```diff
@@ -1163,56 +1229,60 @@ func (r *Reconciler) ReconcileDBConfigMap(cm *corev1.ConfigMap, desiredFunc func
...
-		return err
+	if !r.preparePersistentLoggingPVC(pvc, fieldName) {
+		return util.NewPersistentError(errorName, errorText)
 	}
-	r.Own(r.BucketLoggingPVC)
+	r.Own(pvc);
 
-	log.Infof("BucketLoggingPVC %s does not exist. Creating...", r.BucketLoggingPVC.Name)
-	err := r.Client.Create(r.Ctx, r.BucketLoggingPVC)
+	log.Infof("Persistent loggin PVC %s does not exist. Creating...", *pvcName)
```
jackyalbo, line 1259:
```suggestion
	log.Infof("Persistent logging PVC %s does not exist. Creating...", *pvcName)
```

pkg/system/phase2_creating.go
```diff
@@ -450,6 +468,15 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 			}
 		}
 	}
+
+	if r.NooBaa.Spec.BucketNotifications.Enabled {//TODO - has pvc
```
jackyalbo, line 472: remove the comment, or better explain what is missing.

pkg/system/phase2_creating.go
```diff
@@ -519,6 +546,44 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 				}}
 				util.MergeVolumeMountList(&c.VolumeMounts, &bucketLogVolumeMounts)
 			}
+
+			if r.NooBaa.Spec.BucketNotifications.Enabled {
+				notificationVolumeMounts := []corev1.VolumeMount{{
+					Name:      "notif-vol",
```
jackyalbo, line 552: take out to a constant to guarantee the same name here and in the volume. also, notification-volume sounds better.

jackyalbo (approved):

PR #1473, opened 2024-11-11
Adapting operator to new bucket list API change

jackyalbo:
### Explain the changes
1. After changing the api for list buckets, we need to update operator to support it as well otherwise nb api will fail listing the buckets
2. Adding the new params to the API
3. updating the list bucket nb cli command to support the new params

1. Fixed https://issues.redhat.com/browse/DFBUGS-822

### Testing Instructions:
1. run the scenario in the bug: noobaa-cli bucket list
2. make sure you see the buckets and no schema errors appear

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/bucket/bucket.go
+++ b/pkg/bucket/bucket.go
@@ -239,7 +239,7 @@ func RunStatus(cmd *cobra.Command, args []string) {
 func RunList(cmd *cobra.Command, args []string) {
 	log := util.Logger()
 	nbClient := system.GetNBClient()
-	list, err := nbClient.ListBucketsAPI()
+	list, err := nbClient.ListBucketsAPI(nb.ListBucketsParams{})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


dannyzaken (approved):

liranmauda (approved):
LGTM, 
@jackyalbo I wonder why the CI tests did not find it. can we have a look, and open a gap if needed?

jackyalbo:
@liranmauda I'm not sure if it would have caught the issue, but I don't think that we are using latest noobaa-core in our CI tests for operator. I guess we are using options.ContainerImageTag which is at least 6 months ago (master-20240520).
in anyway we are always running bucket list with silence option which I'm not sure if avoids failing the test: `test_noobaa silence bucket list`
We don't have a dedicated test for buckets in our cli, so a lot is missing.

PR #1482, opened 2024-12-03
Noobaa/Operator: Display error message

aspandey:
While deleting OBC or Bucketclass which does not exist, we should display an error message and should not silently give success.
https://issues.redhat.com/browse/DFBUGS-201
As part of this I have also modified test_cli_functions.sh
Test under "delete_namespacestore_path" function was failing because we do not have a bucket which we are trying to delete. It was passing silently. In this function we are fetching the list of bucketclass and obc freshly and deleting those objects. I think that should cover the cases.

### DIFF
--- a/pkg/bucketclass/bucketclass.go
+++ b/pkg/bucketclass/bucketclass.go
@@ -425,6 +425,11 @@ func RunDelete(cmd *cobra.Command, args []string) {
 	bucketClass.Name = args[0]
 	bucketClass.Namespace = options.Namespace
 
+	if !util.KubeCheck(bucketClass) {
+		log.Fatalf(`❌ Could not delete, BucketClass %q in namespace %q does not exist`,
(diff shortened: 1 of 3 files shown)


shirady (approved):
LGTM

aspandey:
@shirady commented my changes just to see if the test will fail or not.
I will again ask for your approval.

aspandey:
@shirady Patch is ready to be reviewed again.

shirady:
@aspandey I don't have additional comments.
Please update the PR description with the changes that were added in `test/cli/test_cli_functions.sh` so it will be clear why you deleted it and where those are covered.

aspandey:
> @aspandey I don't have additional comments. Please update the PR description with the changes that were added in `test/cli/test_cli_functions.sh` so it will be clear why you deleted it and where those are covered.

Done

PR #1483, opened 2024-12-03
Bug-fix for fetching region from AWS cluster obj in GetAWSRegion()

achouhan09:
### Explain the changes
1. Root Cause: The problem arises when the AWS cluster region cannot be determined from node names, especially in cases where the node naming convention does not include the region or when the DHCP option modifies DNS settings, causing standard parsing methods to fail.

2. Solution: To resolve this, the solution first attempts to fetch the region directly from the cluster configuration, which is the most reliable source. If this fails, it falls back to parsing the region from the node name.

1. Bug: https://issues.redhat.com/browse/DFBUGS-747

### Testing Instructions:
1. Install noobaa operator on AWS cluster and check for default backingstore creation with DHCP option set and not set.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -32,6 +32,7 @@ import (
 	nbapis "github.com/noobaa/noobaa-operator/v5/pkg/apis"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	"github.com/noobaa/noobaa-operator/v5/pkg/bundle"
+	configv1 "github.com/openshift/api/config/v1"
 	routev1 "github.com/openshift/api/route/v1"
 	secv1 "github.com/openshift/api/security/v1"
 	cloudcredsv1 "github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1"
@@ -185,6 +186,7 @@ func init() {
 	Panic(autoscalingv1.AddToScheme(scheme.Scheme))
 	Panic(kedav1alpha1.AddToScheme(scheme.Scheme))
 	Panic(apiregistration.AddToScheme(scheme.Scheme))
+	Panic(configv1.AddToScheme(scheme.Scheme))
 }
 
 // KubeConfig loads kubernetes client config from default locations (flags, user dir, etc)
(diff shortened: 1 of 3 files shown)


pkg/util/util.go
```diff
@@ -1173,6 +1176,21 @@ func GetAWSRegion() (string, error) {
 		return "", fmt.Errorf("Unexpected node name format: %q", nodesList.Items[0].Name)
 	}
 	awsRegion := mapValidAWSRegions[nameSplit[1]]
+
+	// Fetching aws region from cluster if not fetched from node name
+	if awsRegion == "" {
+		infrastructure := &configv1.Infrastructure{}
+		infrastructure.SetName("cluster")
```
liranmauda, line 1183: I think this should work:

```suggestion
		infrastructure := &configv1.Infrastructure{Name: "cluster"}
```

pkg/util/util.go
```diff
@@ -1173,6 +1176,21 @@ func GetAWSRegion() (string, error) {
 		return "", fmt.Errorf("Unexpected node name format: %q", nodesList.Items[0].Name)
 	}
 	awsRegion := mapValidAWSRegions[nameSplit[1]]
+
+	// Fetching aws region from cluster if not fetched from node name
+	if awsRegion == "" {
```
dannyzaken, line 1181: @achouhan09, using the infrastructure CR looks like a better way to get the region than parsing the node name. Maybe we should first attempt to get it this way, and only fallback to the node name if we fail. WDYT?

pkg/util/util.go
achouhan09: Updated, Thanks

pkg/util/util.go
achouhan09: Yes, make sense. I have updated the code. Thanks

pkg/util/util.go
```diff
@@ -1164,17 +1167,35 @@ func GetAWSRegion() (string, error) {
...
+	var awsRegion string
+	infrastructure := &configv1.Infrastructure{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "cluster",
+		},
+	}
+	if _, _, err := KubeGet(infrastructure); err != nil {
+		log.Infof("Failed to fetch cluster infrastructure details: %v", err)
+	}
+	if infrastructure.Status.PlatformStatus != nil && infrastructure.Status.PlatformStatus.AWS != nil {
```
dannyzaken, line 1179: log a warn message if we don't find the expected details in the infrastructure object, before falling back to node name.

dannyzaken (approved):

pkg/util/util.go
achouhan09: Done, Thanks.

PR #1485, opened 2024-12-05
bucket notification - check pvcName, not pvc DFBUGS 988, 991

alphaprinz:
### Explain the changes
1. For bucket logging and bucket notification, we allocate a pvc from cephfs (if no explicit pvc was specified).
The check for whether there's already a pvc was wrong.

1. https://issues.redhat.com/browse/DFBUGS-988
2. https://issues.redhat.com/browse/DFBUGS-991

### Testing Instructions:
1. See bugs for reproduction scenario

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -1242,7 +1242,7 @@ func (r *Reconciler) ReconcileODFPersistentLoggingPVC(
 	log := r.Logger.WithField("func", "ReconcileODFPersistentLoggingPVC")
 
 	// Return if persistent logging PVC already exists
-	if pvc != nil {
+	if pvcName != nil {
 		pvc.Name = *pvcName;
 		log.Infof("PersistentLoggingPVC %s already exists and supports RWX access mode. Skipping ReconcileODFPersistentLoggingPVC.", *pvcName)
 		return nil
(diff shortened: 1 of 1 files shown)


pkg/system/phase2_creating.go
```diff
@@ -1242,7 +1242,7 @@ func (r *Reconciler) ReconcileODFPersistentLoggingPVC(
 	log := r.Logger.WithField("func", "ReconcileODFPersistentLoggingPVC")
 
 	// Return if persistent logging PVC already exists
-	if pvc != nil {
+	if pvcName != nil {
```
aspandey, line 1245: Just a code reading comment.

Somehow it doesn't add up.
if pvcName == nil the we will not enter this if and will not return.
Then  on line number 1253, we will print the name *pvcName.

Next just by checking *pvcName we are saying PersistentLoggingPVC exist and assigning ovcName to pvc.Name. What if pvcName is already have some name because it exist.

I think check on 151 should be the first and then we should check pvcName and act accordingly.

alphaprinz: Good catch, thanks.
I've fix to the correct field in 1253.

aspandey (commented):
LGTM

jackyalbo (approved):

PR #1489, opened 2024-12-10
Noobaa/Backingstore : Need facility to change labels

aspandey:
We need to allow to modify the matchLabel of
topologySpreadConstraints for noobaa's backingstore pods.

https://issues.redhat.com/browse/DFBUGS-277

Testing -
1 - Create a backing store.
2 - Check the label for backingstore and also the pod

kubectl get backingstore my-backingstore -o yaml
kubectl describe pod my-backingstore-noobaa-pod-c2348d65

3 - 
Change the label of a backingstore object like this -
kubectl label backingstore my-backingstore  backingstore=noobaa-new --overwrite
This will trigger the reconciliation and the pod will be reinitialize and now have a "backingstore=noobaa-new" as label

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -1169,6 +1169,13 @@ func (r *Reconciler) needUpdate(pod *corev1.Pod) bool {
 		}
 	}
 
+	var pod_label = pod.Labels["backingstore"]
+	var bs_label = r.BackingStore.Labels["backingstore"]
+	if (bs_label != "") && (pod_label != bs_label) {
+		r.Logger.Warnf("Backingstore lable has been modified to (%v) and it is different from pod lable: (%v)", bs_label, pod_label)
+		return true
+	}
+
 	for _, name := range []string{"NOOBAA_LOG_LEVEL", "NOOBAA_LOG_COLOR"} {
 		configMapValue := r.CoreAppConfig.Data[name]
 		noobaaLogEnvVar := util.GetEnvVariable(&c.Env, name)
@@ -1305,10 +1312,16 @@ func (r *Reconciler) updatePodTemplate() error {
 		r.PodAgentTemplate.Spec.ImagePullSecrets =
 			[]corev1.LocalObjectReference{*r.NooBaa.Spec.ImagePullSecret}
 	}
+
+	bs_label, ok := r.BackingStore.Labels["backingstore"]
+	if !ok || bs_label == "" {
+		bs_label = "noobaa"
+	}
+
 	r.PodAgentTemplate.Labels = map[string]string{
 		"app":          "noobaa",
 		"pool":         r.BackingStore.Name,
-		"backingstore": "noobaa",
+		"backingstore": bs_label,
 	}
 	if r.NooBaa.Spec.Tolerations != nil {
 		r.PodAgentTemplate.Spec.Tolerations = r.NooBaa.Spec.Tolerations
@@ -1329,7 +1342,7 @@ func (r *Reconciler) updatePodTemplate() error {
 			NodeTaintsPolicy:  &honor,
 			LabelSelector: &metav1.LabelSelector{
 				MatchLabels: map[string]string{
-					"backingstore": "noobaa",
+					"backingstore": bs_label,
 				},
 			},
 		}


deploy/crds/noobaa.io_v1alpha1_backingstore_cr.yaml
```diff
@@ -2,4 +2,6 @@ apiVersion: noobaa.io/v1alpha1
 kind: BackingStore
 metadata:
   name: default
+  labels:
+    backingstore: noobaa
```
achouhan09, line 6: I think we should avoid setting a default label for the BackingStore. Instead, users should have the flexibility to add the label as needed.

pkg/backingstore/reconciler.go
```diff
@@ -1169,6 +1169,12 @@ func (r *Reconciler) needUpdate(pod *corev1.Pod) bool {
 		}
 	}
 
+	var pod_label = pod.Labels["backingstore"]
+	var bs_label = r.BackingStore.Labels["backingstore"]
```
achouhan09, line 1173: and we can check here if label is provided in backingstore then we will update the label in pv-pods.

pkg/backingstore/reconciler.go
```diff
@@ -1169,6 +1169,12 @@ func (r *Reconciler) needUpdate(pod *corev1.Pod) bool {
...
+	if pod_label != bs_label {
+		return true
```
achouhan09, line 1175: We should log a warn msg here if the label is updated

pkg/backingstore/reconciler.go
```diff
@@ -1329,7 +1337,7 @@ func (r *Reconciler) updatePodTemplate() error {
 			NodeTaintsPolicy:  &honor,
 			LabelSelector: &metav1.LabelSelector{
 				MatchLabels: map[string]string{
-					"backingstore": "noobaa",
+					"backingstore": bs_label,
```
achouhan09, line 1340:
```suggestion
					"pvbackingstore": bs_label,
```
wdyt? Can we change the label name to `pvbackingstore` ?

deploy/crds/noobaa.io_v1alpha1_backingstore_cr.yaml
aspandey: done

pkg/backingstore/reconciler.go
aspandey: done

pkg/backingstore/reconciler.go
aspandey: Done

pkg/backingstore/reconciler.go
aspandey: I don't think it is needed..and might unnecessary introduce the complications...even  for customer to understand.

pkg/system/phase4_configuring.go
```diff
@@ -417,11 +417,11 @@ func (r *Reconciler) SetDesiredDeploymentEndpoint() error {
 
 			if r.NooBaa.Spec.BucketNotifications.Enabled {
 				envVar := corev1.EnvVar{
-					Name: "NOTIFICATION_LOG_DIR",
+					Name:  "NOTIFICATION_LOG_DIR",
```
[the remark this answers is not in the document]
aspandey: This was made automatically while saving file. Changed it

pkg/backingstore/reconciler.go
```diff
@@ -1305,10 +1312,15 @@ func (r *Reconciler) updatePodTemplate() error {
 		r.PodAgentTemplate.Spec.ImagePullSecrets =
 			[]corev1.LocalObjectReference{*r.NooBaa.Spec.ImagePullSecret}
 	}
+
+	bs_label, ok := r.BackingStore.Labels["backingstore"]
+	if !ok || bs_label == "" {
+		bs_label = "noobaa"
+	}
```
achouhan09, line 1319:
```suggestion
	bs_label := r.BackingStore.Labels["backingstore"]
	if bs_label == "" {
		bs_label = "noobaa"
	}
```
Just a small suggestion, I think it will return zero "" value if the label is not defined so we can update it.

achouhan09 (approved):
LGTM

PR #1493, opened 2024-12-11
Hackathon | Noobaa PGDB integrity and stat check

naveenpaul1:
### Explain the changes
1. Integrity check for Noobaa PGDB​
   Verify Integrity of Noobaa md parts, chunks, fragments​
   Check only PGDB integrity, not the actual storage​
2. Noobaa PGDB stat check​
   Check CPU\time consuming queries and number of query count

1. DB integrity related issues
2. Identify bottleneck  Queries in PGDB

### Testing Instructions:
1. nb diagnostics int-check --dump-data-map --namespace {namespace}​
    -- Return integrity issues count if any​
    -- dump-data-map flag dump data hierarchy map in JSON format
2. nb diagnostics db-stat-prepare --namespace {namespace}​
      -- Create extension pg_stat_statements and rest metrics​

doc : https://ibm.ent.box.com/file/1722250034383
- [X] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/diagnostics/db_stat.go
+++ b/pkg/diagnostics/db_stat.go
@@ -6,6 +6,7 @@ import (
 	"os/exec"
 	"time"
 
+	"github.com/noobaa/noobaa-operator/v5/pkg/options"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


nimrod-becker:
Feedback from the session:
1. add an ability to get a key name (or list of keys) to analyze
2. show only the problematic keys, if everything is ok, don't print

PR #1509, opened 2025-01-16
Setting noobaa account state "rejected"

vh05:
Missing default_resource parameter which is mandatory parameter for account creation, should mark the nooobaa account status as rejected

Fixes: #https://issues.redhat.com/browse/DFBUGS-1297


liranmauda:
@vh05 Why do we only return util.NewPersistentError on the update?

liranmauda (approved):

PR #1510, opened 2025-01-20
Draft: Add field to NooBaa's CR to override environment variables of pods

### LINKED ISSUE #1486: How can I add custom environment variables to noobaa-core deployment through noobaa operator ?
Hi,

I would like to modify NooBaa's default behaviour without having to build a new image with a modified config.js. (Would like to enable config.WORM_ENABLED to allow object lock).

Reading the code I found out that this can be done using environment variables prefixed with `CONFIG_JS_` (https://github.com/noobaa/noobaa-core/blob/4847584ca6b10aeea6f596b30edaabf91a3ce7ae/config.js#L1015). 

Unfortunately the NooBaa CR doesn't seem flexible enough to allow a user to specify custom environment variables (https://github.com/noobaa/noobaa-operator/blob/master/deploy/internal/statefulset-core.yaml) 

Am I correct to assume that for now, building a new image is the only way for me to modify this ? Or am I missing something ?

Thanks

NoOverflow:
### Explain the changes
This is a proposal to add a envVariablesOverride field to NooBaa's CR in order to customize some environment variables for all NooBaa's pods (right now only "core" sts and "endpoint" deploy pods have been tested and work)

This would allow for example[ overriding CONFIG_JS_ variables](https://github.com/noobaa/noobaa-core/blob/4847584ca6b10aeea6f596b30edaabf91a3ce7ae/config.js#L1015) without having to rebuild a custom image on each release.

Currently, the new CR is defined like so, but happy to use another format if you have a better idea:
```YAML
spec:
  autoscaler: {}
  bucketLogging: {}
  bucketNotifications:
    enabled: false
  cleanupPolicy: {}
  dbImage: quay.io/sclorg/postgresql-15-c9s
  dbType: postgres
  envVariablesOverride:
    core:
    - name: CONFIG_JS_WORM_ENABLED
      value: "true"
    endpoint:
    - name: [...]
      value: [...]
  image: noobaa/noobaa-core:master-20240520
  loadBalancerSourceSubnets: {}
  security:
    kms: {}
```

Right now, all env variables containers of a single pod are overriden (it is not on a per-container basis). 
Variables defined in envVariablesOverride have priority over all variables set by the operator in case of a conflict.

1. Fixes #1486 

### Testing Instructions:
1. Create a NooBaa object with the envVariablesOverride field set
2. Confirm each pod mentioned has its environment variables modified

- [ ] Added override for all pods 
- [ ] Test for special cases (variable conflict)
- [ ] Doc added/updated
- [ ] Tests added


tangledbytes:
Hi, thanks for the PR. Right now we simply recommend to use `kubectl set env ...` for the same. Refer: https://kubernetes.io/docs/reference/kubectl/generated/kubectl_set/kubectl_set_env.

Is the above lacking something?

PR #1512, opened 2025-01-22
Build multiarch (amd64 and arm64) images

zregvart:
### Explain the changes
1. Makes changes to the scripts in build/tools to build and push multiarch images. Now runs two `docker build` commands to build images for individual architectures, and an additional `docker manifest create` to create the multiarch manifest. When pushing `docker manifest push` is used instead of `docker push`.

1. n/a

### Testing Instructions:
1. Try manually running a release using the scripts in build/tools

- [ ] Doc added/updated
- [ ] Tests added


tangledbytes:
Hi, thanks for the PR. Can you please rebase it, I think after that all the tests would start passing.

liranmauda:
@tangledbytes Could you take a look?

PR #1533, opened 2025-02-18
Remove the root secret from pod env variables

Neon-White:
### Explain the changes
1. As reported in [DFBUGS-1608](https://issues.redhat.com/browse/DFBUGS-1608), the NooBaa root secret would be unintentionally revealed in some cases. This PR aims to remediate the issue by removing legacy code that was used to pass the secret via env params, before it was changed to be passed via a file mount. ([core](https://github.com/noobaa/noobaa-core/pull/7218), [operator](https://github.com/noobaa/noobaa-operator/pull/1071))

### Issues: 
Fixed:
1. [DFBUGS-1608](https://issues.redhat.com/browse/DFBUGS-1608)

Gap:
1. Some dead code is still present on the core side 

### Testing Instructions:
1. Deploy NooBaa on an AWS IPI Thales configuration
4. Make sure `NOOBAA_ROOT_SECRET` does not show in the env variables of endpoint and core pods

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/internal/deployment-endpoint.yaml
+++ b/deploy/internal/deployment-endpoint.yaml
@@ -117,7 +117,6 @@ spec:
             - name: ENDPOINT_GROUP_ID
             - name: LOCAL_MD_SERVER
             - name: LOCAL_N2N_AGENT
-            - name: NOOBAA_ROOT_SECRET
             - name: NODE_EXTRA_CA_CERTS
             - name: GUARANTEED_LOGS_PATH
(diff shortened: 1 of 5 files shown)


liranmauda:
LGTM,
We need to figure out if the upgrade path would be affected.
If it works, we can manage it, and create a PR on core that removes the env variable.
@romayalon for the core side, do we use this env on NC?

shirady (approved):
LGTM

PR #1534, opened 2025-02-19
Fixing mount volume root keys - limit the secret size

jackyalbo:
### Explain the changes
1. Will mount keys from the last 6 months - minimum keys set to 50.

1. Fixed https://issues.redhat.com/browse/DFBUGS-1344

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/util/kms/kms_version.go
+++ b/pkg/util/kms/kms_version.go
@@ -2,6 +2,8 @@ package kms
 
 import (
 	"fmt"
+	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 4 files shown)


pkg/util/kms/kms_version.go
```diff
@@ -203,6 +205,53 @@ func keyName() string {
...
+	}
+	var entries []struct {
+		key   string
+		value int64
+	}
+	for k := range s {
+		if k == ActiveRootKey {
+			continue
+		}
+		splitKey := strings.Split(k, "-");
```
liranmauda, line 222: Could we add a comment on why we are splitting it? and what each part contains.

liranmauda (approved):

PR #1554, opened 2025-03-10
Added a fix to update bucket tagging when OBC labels are updated

achouhan09:
### Explain the changes
1.  Added a mechanism to detect label changes in OBCs and trigger a custom function whenever an update occurs. This ensures that label modifications are automatically recognized and handled without requiring a full reconciliation.

1. Fixed: https://issues.redhat.com/browse/DFBUGS-1615

### Testing Instructions:
1.  `make test-cli-flow`
- [X] Tests added

### DIFF
--- a/pkg/obc/provisioner.go
+++ b/pkg/obc/provisioner.go
@@ -130,9 +130,9 @@ func (p *Provisioner) Provision(bucketOptions *obAPI.BucketOptions) (*nbv1.Objec
 		return nil, err
 	}
 
-	err = r.putBucketTagging()
+	err = UpdateBucketTagging(r.SysClient, r.OBC)
 	if err != nil {
-		logrus.Warnf("failed executing putBucketTagging on bucket: %v, %v", r.BucketName, err)
+		logrus.Warnf("failed executing UpdateBucketTagging on bucket: %v, %v", r.BucketName, err)
 	}
 	return r.OB, nil
 }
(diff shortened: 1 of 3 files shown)


pkg/obc/provisioner.go
```diff
@@ -243,6 +243,66 @@ func (p *Provisioner) Update(ob *nbv1.ObjectBucket) error {
...
 
+// UpdateBucketTagging is not handled by lib-bucket-provisioner and is a customized function to update bucket tagging with OBC labels
+func UpdateBucketTagging(sysClient *system.Client, obc *nbv1.ObjectBucketClaim) error {
+	if obc == nil || obc.Labels == nil {
+		return fmt.Errorf("OBC is not provided or doesn't contain any label")
+	}
+
+	var err error
+	if sysClient == nil {
+		sysClient, err = system.Connect(false)
```
jackyalbo, line 254: I would take it this out from the function to the EventHandler and not support sysClient to be nil

pkg/obc/provisioner.go
```diff
@@ -677,50 +737,7 @@ func (r *BucketRequest) DeleteBucket() error {
...
-		Bucket: &r.BucketName,
-		Tagging: &s3.Tagging{
-			TagSet: taggingArray,
-		},
-	})
-	if err != nil {
-		return err
-	}
-	return nil
+	return UpdateBucketTagging(r.SysClient, r.OBC)
```
jackyalbo, line 740: just remove this function and call UpdateBucketTagging in the Provision

jackyalbo (commented):
Nice, but I would love to see a test for this. some test that creates an obc w/o tagging. adds/update the tags and verifies it was changed.

pkg/obc/provisioner.go
achouhan09: Done, Thanks

PR #1570, opened 2025-03-24
CA Paths Fixes

shirady:
### Explain the changes
1. In `pkg/utilutil.go` - Change the `/etc/ocp-injected-ca-bundle.crt` to `/etc/ocp-injected-ca-bundle/ca-bundle.crt`.
2. In `pkg/system/reconciler.go` - Rename `r.CaBundleConf.Name` from `r.Request.Name + "-ca-inject"` to `"ocp-injected-ca-bundle"` to align with the rename of `configmap-ca-inject.yaml` from `noobaa-ca-inject` to `ocp-injected-ca-bundle` in PR #1328.
3. In `pkg/system/phase2_creating.go` and `pkg/system/phase4_configuring.go` - the core and the endpoint change the mount path from `/etc/ocp-injected-ca-bundle.crt` (with file extension) to a directory (remove the extension).
4. Rename `AddToRootCAs` to `CombineCaBundle`, add the const `InjectedBundleCertCAFile` which is `/etc/ocp-injected-ca-bundle/ca-bundle.crt`, and assign `r.ApplyCAsToPods` to be the new constant (will be `NODE_EXTRA_CA_CERTS` eventually).
5. Change the condition of checking only the existence of the config map of the CA (`configmap-ca-inject`) to check this and also that we have data in it.
The reason is that when we test the OLM (in the CI), we have the OLM and then we have the operator so the config map as it is created by the OLM, but in the first step it is empty.
When manually looking at the config map in a real cluster there is data, that probably comes from additional needed configuration in the cluster (which is not in the scope of the test).
6. Add a comment about `ApplyCAsToPods` so it would be clear that it is relevant to the endpoint and core pods.
7. in `CombineCaBundle` use const instead of hard-coded string (to avoid circular dependency had to move the consts from `options` to `util` file).
8. (this change might be temporary - need to decide) Change the `ApplyCAsToPods` to `ServiceServingCertCAFile` (previous `InjectedBundleCertCAFile` in comment).

### Issues:
1.  Currently, when I tried to deploy with the cluster-bot I saw that `NODE_EXTRA_CA_CERTS` is empty, while it used to be 
`NODE_EXTRA_CA_CERTS=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt`.

GAPs:
1. Tests end-to-end with certificates.

#### Basic Manual Tests (more details in the [comment below](https://github.com/noobaa/noobaa-operator/pull/1570#issuecomment-2748516358))
Check the value of `NODE_EXTRA_CA_CERTS`:
1. I used the cluster-bot with `launch 4.17 aws`
2. Installed ODF operator and MCG standalone according to the docs [here](https://docs.redhat.com/en/documentation/red_hat_openshift_data_foundation/4.17/html/deploying_and_managing_openshift_data_foundation_using_red_hat_openstack_platform/deploy-standalone-multicloud-object-gateway#deploy-standalone-multicloud-object-gateway).
3. Check the env `NODE_EXTRA_CA_CERTS`:
  - `oc exec statefulset/noobaa-core -c core -n openshift-storage -- printenv | grep NODE_EXTRA_CA_CERTS` (was empty)
  - `oc exec noobaa-endpoint-<characters> -n openshift-storage -- env | grep NODE_EXTRA_CA_CERTS` (was empty)
4. same steps with `launch 4.14 aws` in the cluster-bot (the version was chosen as a version before the changes of PR #1328) we could see `NODE_EXTRA_CA_CERTS=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` (both on core and endpoint).
5. After the change we would see (both on core and endpoint pods): 
  - `NODE_EXTRA_CA_CERTS=/etc/ocp-injected-ca-bundle/ca-bundle.crt`
  - `NODE_EXTRA_CA_CERTS=/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt` (with the commit mentioned in change number 8).

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -48,12 +48,6 @@ const (
 
 	// SystemName is a constant as we want just a single system per namespace
 	SystemName = "noobaa"
-
-	// ServiceServingCertCAFile points to OCP default root CA list
-	ServiceServingCertCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"
-
-	// InjectedBundleCertCAFile points to OCP root CA to be added to the default root CA list
-	InjectedBundleCertCAFile = "/etc/ocp-injected-ca-bundle/ca-bundle.crt"
 )
 
 // Namespace is the target namespace for locating the noobaa system
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -49,8 +49,11 @@ const (
 	// SystemName is a constant as we want just a single system per namespace
 	SystemName = "noobaa"
 
-	// ServiceServingCertCAFile points to OCP root CA to be added to the default root CA list
+	// ServiceServingCertCAFile points to OCP default root CA list
 	ServiceServingCertCAFile = "/var/run/secrets/kubernetes.io/serviceaccount/service-ca.crt"
+
+	// InjectedBundleCertCAFile points to OCP root CA to be added to the default root CA list
+	InjectedBundleCertCAFile = "/etc/ocp-injected-ca-bundle/ca-bundle.crt"
 )
 
 // Namespace is the target namespace for locating the noobaa system
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -547,7 +547,8 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 			if r.NooBaa.Spec.CoreResources != nil {
 				c.Resources = *r.NooBaa.Spec.CoreResources
 			}
-			if util.KubeCheckQuiet(r.CaBundleConf) {
+			// we want to check that the cm exists and also that it has data in it
+			if util.KubeCheckQuiet(r.CaBundleConf) && len(r.CaBundleConf.Data) > 0 {
 				configMapVolumeMounts := []corev1.VolumeMount{{
 					Name:      r.CaBundleConf.Name,
 					MountPath: "/etc/ocp-injected-ca-bundle",
@@ -630,7 +631,8 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 					Limits:   logResourceList,
 				}
 			}
-			if util.KubeCheckQuiet(r.CaBundleConf) {
+			// we want to check that the cm exists and also that it has data in it
+			if util.KubeCheckQuiet(r.CaBundleConf) && len(r.CaBundleConf.Data) > 0 {
 				configMapVolumeMounts := []corev1.VolumeMount{{
 					Name:      r.CaBundleConf.Name,
 					MountPath: "/etc/ocp-injected-ca-bundle",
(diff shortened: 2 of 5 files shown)


shirady:
# Additional Testing Instructions:
## 1) code change - `util.go`
### Explanation:
`NODE_EXTRA_CA_CERTS` is empty has been happening since the value was assigned from: 
(in the core pod)
https://github.com/noobaa/noobaa-operator/blob/3f43b190f8c8d777902f3489a68be71b51699f7a/pkg/system/phase2_creating.go#L467-L468
(in the endpoint pod)
https://github.com/noobaa/noobaa-operator/blob/3f43b190f8c8d777902f3489a68be71b51699f7a/pkg/system/phase4_configuring.go#L407-L408
and `r.ApplyCAsToPods` is assigned after:
https://github.com/noobaa/noobaa-operator/blob/3f43b190f8c8d777902f3489a68be71b51699f7a/pkg/system/reconciler.go#L407-L414
but in the function `AddToRootCAs` the hard-coded path does not exist in the operator pod:
https://github.com/noobaa/noobaa-operator/blob/3f43b190f8c8d777902f3489a68be71b51699f7a/pkg/util/util.go#L151-L154
In version 4.17 I run:
- connect to the operator pod: `oc rsh noobaa-operator-8498d54858-rx8h9`
- check the previous path: `ls -al /etc/pki/ca-trust/extracted/pem` (it's a directory and it is generated from OCP - not NooBaa):
```
total 904
drwxr-xr-x. 3 root root    123 Sep 18  2024 .
drwxr-xr-x. 6 root root     70 Sep 18  2024 ..
-rw-r--r--. 1 root root    898 Aug 19  2024 README
dr-xr-xr-x. 2 root root  16384 Sep 18  2024 directory-hash
-r--r--r--. 1 root root 165521 Sep 18  2024 email-ca-bundle.pem
-r--r--r--. 1 root root 502506 Sep 18  2024 objsign-ca-bundle.pem
-r--r--r--. 1 root root 226489 Sep 18  2024 tls-ca-bundle.pem
```
- check the code path in file level: `ls -al /etc/ocp-injected-ca-bundle.crt`
```
ls: cannot access '/etc/ocp-injected-ca-bundle.crt': No such file or directory
```
- check the path in the directory level: `ls -al /etc/ocp-injected-ca-bundle`
```
total 0
drwxrwsrwx. 3 root 1000710000 81 Mar 24 10:06 .
drwxr-xr-x. 1 root root       36 Mar 24 10:06 ..
drwxr-sr-x. 2 root 1000710000 27 Mar 24 10:06 ..2025_03_24_10_06_38.1871780206
lrwxrwxrwx. 1 root 1000710000 32 Mar 24 10:06 ..data -> ..2025_03_24_10_06_38.1871780206
lrwxrwxrwx. 1 root 1000710000 20 Mar 24 10:06 ca-bundle.crt -> ..data/ca-bundle.crt
```
- from this I took the file path: `ls -al /etc/ocp-injected-ca-bundle/ca-bundle.crt`
```
lrwxrwxrwx. 1 root 1000710000 20 Mar 24 10:06 /etc/ocp-injected-ca-bundle/ca-bundle.crt -> ..data/ca-bundle.crt
```

shirady:
I'm closing this PR as it continues in PR #1587

PR #1571, opened 2025-03-24
Implemented data import from the old DB to the cluster on upgrade

dannyzaken:
### Explain the changes
- When upgrading from a deployment with an existing DB, provide the DB details in the cluster CR, to perform data import.
- Upgrade flow is as follows:
  - If DB sts exist:
    - Stop core\endpoints pods
    - Create a cnpg cluster with import details in the CR
    - The cnpg operator will start the first instance and deploy an import job to import the data. After the import is completed, the second instance will started.
    - Wait for cluster to complete import and become Ready
    - Once ready scale down the noobaa-db-pg sts to 0 replicas.
    - The reconciliation will wait for the cluster to be ready, and a temp error is returned until it is.
- The old sts is not deleted, to have it around for backup. In future versions we should cleanup all the leftovers.
- Updated `noobaa upgrade` command in noobaa cli, to install the necessary cnpg resources.

### Gaps
- CNPG operator performs the import by performing a pg_dump onto the data volume of the instance and then restoring the dump to the new DB. If the existing DB is too large, we can get into an out-of-space situation. To handle this, we should check the percentage of used space in the existing DB and, above a certain threshold, provision a larger volume for the new cluster. 

### Testing Instructions:
1. Install noobaa from an older version.
2. after installed, run `noobaa upgrade` with a latest build of the CLI. this should install missing cnpg resources and update the noobaa CR to perform import

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/db_reconciler.go
+++ b/pkg/system/db_reconciler.go
@@ -6,12 +6,14 @@ import (
 	"slices"
 
 	cnpgv1 "github.com/cloudnative-pg/cloudnative-pg/api/v1"
+	"github.com/go-test/deep"
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	"github.com/noobaa/noobaa-operator/v5/pkg/cnpg"
 	"github.com/noobaa/noobaa-operator/v5/pkg/options"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


alphaprinz (approved):

pkg/system/db_reconciler.go
```diff
@@ -135,52 +233,115 @@ func (r *Reconciler) reconcileCluster() error {
...
-	// check if the cluster is changed
-	if reflect.DeepEqual(*existingClusterSpec, r.CNPGCluster.Spec) {
-		// The cluster spec is unchanged, no need to update
-		return nil
+}
+
+func (r *Reconciler) reconcileClusterImport() error {
+
+	// The bootstrap configuration should only be set for the CR creation.
+
```
dannyzaken, line 253: regardless of this code only running in cluster creation flow. guard it to look at the noobaaDbCluster

pkg/system/db_reconciler.go
```diff
@@ -135,52 +233,115 @@ func (r *Reconciler) reconcileCluster() error {
...
+		externalClusterName := r.NooBaaPostgresDB.Name
+		//setup once the pods are terminated, set import in the bootstrap configuration and continue to create the cluster
+		r.CNPGCluster.Spec.Bootstrap.InitDB.Import = &cnpgv1.Import{
+			Source: cnpgv1.ImportSource{
+				ExternalCluster: externalClusterName,
+			},
+			// microservice type - import only the nbcore database
+			Type: cnpgv1.MicroserviceSnapshotType,
+			Databases: []string{
+				"nbcore",
```
dannyzaken, line 289: remove "magic strings" and use consts\options instead

dannyzaken:
I added comments based on an online review with @jackyalbo and @liranmauda. Will handle them in a different PR

PR #1572, opened 2025-03-26
Cluster reconcile changes

dannyzaken:
### Explain the changes
- added support for node affinity and tolerations
- added PostgreSQL config values to cluster spec
  - Added `dbConf` field to `dbSpec` to allow overriding default parameters
 
#### Changes to install flow
- Wait for noobaa-operator and cnpg-operator to be ready before creating a system. 

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/install/install.go
+++ b/pkg/install/install.go
@@ -121,6 +121,12 @@ func RunInstall(cmd *cobra.Command, args []string) {
 		log.Printf("")
 	}
 
+	// wait for the operators to be ready before creating the system
+	util.WaitForOperatorDeploymentReady(options.Namespace, "noobaa-operator")
+	if useCNPG {
+		util.WaitForOperatorDeploymentReady(options.Namespace, cnpg.CnpgDeploymentName)
+	}
+
 	log.Printf("System Create:")
 	system.RunCreate(cmd, args)
 	log.Printf("")
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -44,6 +44,7 @@ import (
 	"github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 	"golang.org/x/term"
+	appsv1 "k8s.io/api/apps/v1"
 	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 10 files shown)


pkg/util/util.go
```diff
@@ -1295,6 +1296,31 @@ func PrintThisNoteWhenFinishedApplyingAndStartWaitLoop() {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      name,
+				Namespace: namespace,
+			},
+		}
+		found := KubeCheckQuiet(deployment)
+		if found {
+			ready := deployment.Status.ReadyReplicas == deployment.Status.Replicas
+			if !ready {
+				log.Printf("⏳ Waiting for %s deployment to be ready", name)
```
liranmauda, line 1312: Will this be printed every second (till it is ready)? 
If yes, we should avoid overloading the logs

liranmauda (approved):
LGTM, one comment on the printing.

pkg/util/util.go
```diff
@@ -1295,6 +1296,31 @@ func PrintThisNoteWhenFinishedApplyingAndStartWaitLoop() {
 	log.Printf("  - You may Ctrl-C at any time to stop the loop and watch it manually.")
 }
 
+// WaitForOperatorDeploymentReady waits for a deployment to be have the desired number of replicas ready
```
jackyalbo, line 1299:
```suggestion
// WaitForOperatorDeploymentReady waits for a deployment to have the desired number of replicas ready
```

pkg/install/install.go
```diff
@@ -121,6 +121,12 @@ func RunInstall(cmd *cobra.Command, args []string) {
 		log.Printf("")
 	}
 
+	// wait for the operators to be ready before creating the system
```
jackyalbo, line 124: Do we have to have this part now? Usually, I break the install when waiting for the system to be Ready, but now breaking the install on wait won't create the system. 
Also, should(can) we add no-wait option support here as well?

pkg/util/util.go
dannyzaken: This is no to the logs. These are messages for the cli install command, same as we always had. I just wait now before creating the system

PR #1574, opened 2025-04-01
CI | Test CLI Flow | Add Sleep to Avoid Failure of Bucket Deletion in Deleting State

shirady:
### Describe the Problem
1. Currently, we might encounter failure in the CI of the job `run-cli-test` as it tries to delete buckets in a deleting state.

### Explain the changes
1. Adding a sleep to avoid seeing buckets in deleting state (in list bucket), because later we try to delete them.

### Issues: 
1. This a [link](https://github.com/noobaa/noobaa-operator/actions/runs/14154326946/job/39651579645?pr=1570) example of a failed run without code changes (below the downloaded file).
You can see:
```
09:06:05 💬 Deleting buckets
2025-03-30 09:05:52 multinamespace-obc-deleting-1743325553731
2025-03-30 09:05:43 multinamespace-obc-deleting-1743325546281
2025-03-30 08:49:31 second.bucket
2025-03-30 08:54:24 failns.bucket1
2025-03-30 08:45:05 first.bucket
2025-03-30 08:53:48 target.bucket2
2025-03-30 08:53:46 target.bucket1
2025-03-30 08:53:43 fails3.bucket
```
We see buckets whose names contain `deleting-<timestamp>`.

And the failure comes from:
```
remove_bucket failed: Unable to delete all objects in the bucket, bucket will not be deleted.
09:06:07 ❌  aws s3 rb s3://multinamespace-obc-deleting-1743325553731 --force failed, Exiting
```

The downloaded file: [0_run-tests.txt](https://github.com/user-attachments/files/19550053/0_run-tests.txt)

GAP - we shouldn't see buckets in deleting state at first place according to the code in [noobaa-core bucket server](https://github.com/noobaa/noobaa-core/blob/e4925c5ba23c112487d98bf8e63b42b33fdd3258/src/server/system_services/bucket_server.js#L1073-L1075).

### Testing Instructions:
1. Run in CI.

- [ ] Doc added/updated
- [ ] Tests added


achouhan09 (approved):
LGTM

Just one comment these changes would increase the run time of job. So, Is there any other way we need to try I am not sure of it.

shirady:
> LGTM
> 
> Just one comment these changes would increase the run time of the job. So, Is there any other way we need to try I am not sure of it.

-  concluded that the failures were due to timing and should not happen on the server side. As the test is on the client side, I suggested fixing it by giving time to the server to return the list without buckets in the deleting state.
- I didn't test it over time, but the value of the sleep was chosen arbitrarily and probably can be decreased.
- Anyway, I think we need to solve it on the server side (the core repo), I wrote it as a gap.

PR #1580, opened 2025-04-07
Block NSS creation when  SignatureVersion v4 and endpoint non-secure

naveenpaul1:
### Explain the changes
1. Block NSS creation when SignatureVersion v4 and endpoint non-secure and return with error.
2. Setting signature version v2 for non-secure endpoint to avoid body content sign in. With secure request body sigin is disabled by [default.](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/Config.html#s3DisableBodySigning-property)

No change for Backingstore since uploadPart upload(Non-file stream) is not supported by backingstore. 

1. https://issues.redhat.com/browse/DFBUGS-1035
NSS is in rejected state because the endpoint provided is non-secure and CLI is adding signature version to the endpoint request when try to access data this is creating issue in aws-sdk

related Github issue: https://github.com/aws/aws-sdk-js/issues/965

### Testing Instructions:
1. go test ./pkg/cli/

- [ ] Doc added/updated
- [X] Tests added

### DIFF
--- a/pkg/namespacestore/namespacestore.go
+++ b/pkg/namespacestore/namespacestore.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"net/url"
 	"os"
 	"strings"
 	"time"
@@ -125,10 +126,10 @@ func CmdCreateAWSSTSS3() *cobra.Command {
 // CmdCreateGoogleCloudStorage returns a CLI command
 func CmdCreateGoogleCloudStorage() *cobra.Command {
 	cmd := &cobra.Command{
-		Hidden: true, //TODO: remove once we want to expose it. 
-		Use:   "google-cloud-storage <namespace-store-name>",
-		Short: "Create google-cloud-storage namespace store",
-		Run:   RunCreateGoogleCloudStorage,
+		Hidden: true, //TODO: remove once we want to expose it.
+		Use:    "google-cloud-storage <namespace-store-name>",
+		Short:  "Create google-cloud-storage namespace store",
+		Run:    RunCreateGoogleCloudStorage,
 	}
 	cmd.Flags().String(
 		"target-bucket", "",
@@ -173,7 +174,7 @@ func CmdCreateS3Compatible() *cobra.Command {
 		"The target S3 endpoint",
 	)
 	cmd.Flags().String(
-		"signature-version", "v4",
+		"signature-version", "",
 		"The S3 signature version v4|v2",
 	)
 	cmd.Flags().String(
@@ -321,6 +322,7 @@ func createCommon(cmd *cobra.Command, args []string, storeType nbv1.NSType, popu
 	if len(args) != 1 || args[0] == "" {
 		log.Fatalf(`❌ Missing expected arguments: <namespace-store-name> %s`, cmd.UsageString())
 	}
+
 	name := args[0]
 	secretName, _ := cmd.Flags().GetString("secret-name")
 	cmdAccessMode, _ := cmd.Flags().GetString("access-mode")
@@ -557,12 +559,23 @@ func RunCreateGoogleCloudStorage(cmd *cobra.Command, args []string) {
 // RunCreateS3Compatible runs a CLI command
 func RunCreateS3Compatible(cmd *cobra.Command, args []string) {
 	createCommon(cmd, args, nbv1.NSStoreTypeS3Compatible, func(namespaceStore *nbv1.NamespaceStore, secret *corev1.Secret) {
+		log := util.Logger()
 		endpoint := util.GetFlagStringOrPrompt(cmd, "endpoint")
 		targetBucket := util.GetFlagStringOrPrompt(cmd, "target-bucket")
 		sigVer, _ := cmd.Flags().GetString("signature-version")
 		secretName, _ := cmd.Flags().GetString("secret-name")
 		mandatoryProperties := []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"}
-
+		u, _ := url.Parse(endpoint)
+		if u.Scheme == "http" {
+			if sigVer == "v4" {
+				log.Fatalf("Non-secure endpoint works only with v2 signature-version. Please select signature version v2 for namespacestore")
+			}
+			sigVer = "v2"
+		} else {
+			if sigVer == "" {
+				sigVer = "v4"
+			}
+		}
 		if secretName == "" {
 			accessKey := util.GetFlagStringOrPromptPassword(cmd, "access-key")
 			secretKey := util.GetFlagStringOrPromptPassword(cmd, "secret-key")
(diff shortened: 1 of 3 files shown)


pkg/backingstore/backingstore.go
```diff
@@ -546,7 +547,11 @@ func RunCreateS3Compatible(cmd *cobra.Command, args []string) {
 		sigVer, _ := cmd.Flags().GetString("signature-version")
 		secretName, _ := cmd.Flags().GetString("secret-name")
 		mandatoryProperties := []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"}
-
+		u, _ := url.Parse(endpoint)
+		if u.Scheme == "http" {
+			sigVer = ""
```
liranmauda, line 552: We should add here a printing that the signature-version v2 will be used.
We should also fail here if a `signature-version` is provided and it is v4

pkg/namespacestore/namespacestore.go
```diff
@@ -562,6 +563,10 @@ func RunCreateS3Compatible(cmd *cobra.Command, args []string) {
 		sigVer, _ := cmd.Flags().GetString("signature-version")
 		secretName, _ := cmd.Flags().GetString("secret-name")
 		mandatoryProperties := []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"}
+		u, _ := url.Parse(endpoint)
+		if u.Scheme == "http" {
+			sigVer = ""
```
liranmauda, line 568: We should add here a printing that the signature-version v2 will be used.
We should also fail here if a signature-version is provided and it is v4

pkg/backingstore/backingstore.go
naveenpaul1: `We should also fail here if a signature-version is provided and it is v4` -->> default value for `signature-version` is always V4, There is no way we can identify whether this value came from user or for default value.

liranmauda: So, what is the meaning of sigVer = "" ?
If the meaning is that it will not be set, and as a result it will be V2, then when a user explicitly sets it to v4 and user HTTP we should fail.

naveenpaul1: @liranmauda  So your suggestion is to make the property `signature-version` in CLI?

pkg/namespacestore/namespacestore.go
```diff
@@ -321,6 +322,12 @@ func createCommon(cmd *cobra.Command, args []string, storeType nbv1.NSType, popu
 	if len(args) != 1 || args[0] == "" {
 		log.Fatalf(`❌ Missing expected arguments: <namespace-store-name> %s`, cmd.UsageString())
 	}
+	endpoint := util.GetFlagStringOrPrompt(cmd, "endpoint")
+	sigVer, _ := cmd.Flags().GetString("signature-version")
```
jackyalbo, line 326: Does it return v4 in case the user didn't set any signature (the default)? if so, we need to handle this correctly as the user didn't explicitly asked for v4 signature version.

pkg/namespacestore/namespacestore.go
```diff
@@ -321,6 +322,12 @@ func createCommon(cmd *cobra.Command, args []string, storeType nbv1.NSType, popu
...
+	u, _ := url.Parse(endpoint)
+	if u.Scheme == "http" && sigVer == "v4" {
+		log.Fatalf("Non-secure endpoint works only with v2 signature-version. Please select signature version v2 for namespacestore")
```
jackyalbo, line 329: shouldn't we also handle this for backingstore as well?

pkg/namespacestore/namespacestore.go
```diff
@@ -321,6 +322,12 @@ func createCommon(cmd *cobra.Command, args []string, storeType nbv1.NSType, popu
...
+	}
```
jackyalbo, line 330: Please add test cases for this in the CLI tests.

pkg/namespacestore/namespacestore.go
naveenpaul1: @jackyalbo we did the test for backingstore and couldnt reproduce the issue in backingstore, looks like it's an issue only with NSS.

pkg/namespacestore/namespacestore.go
naveenpaul1: one option it mention in error message that default value for `signature-version` is v4 and user has to explicitily select v2 from non-secure endpoint, What do you think?

jackyalbo: I just think maybe our default should be v2 in this case - and not asking the user to select signature-version when maybe he is not aware of this option.

pkg/namespacestore/namespacestore.go
jackyalbo: this sounds weird, I think we are using the same connection for both namespace and backingstore. Do you know of what flow does it happen in namespacestore?

pkg/namespacestore/namespacestore.go
naveenpaul1: I got this [comment](https://github.com/noobaa/noobaa-operator/pull/1580#discussion_r2031186157) from Liran, In that comment he mentioned that we should not change the value without letting customer know

pkg/namespacestore/namespacestore.go
jackyalbo: ok, I think the issue is with uploadPart which used only by namespace_s3. So we dont need to add this to backingstore. So resolving this conversation. Thanks

pkg/backingstore/backingstore.go
jackyalbo: My suggestion is to set the default to empty (""), in case we will get empty string we will always send v4 unless it's namespacestore s3 with http - than v2. only if we will get v4 by user request  - we will fail. @naveenpaul1 @liranmauda what do you think?

pkg/namespacestore/namespacestore.go
jackyalbo: Replied in Liran comment. Resolving this.

shirady:
@naveenpaul1 - I think the same code change can also be added in the backingstore CLI (it has similar options like namespacestore CLI in some of the cases), WDYT?

pkg/namespacestore/namespacestore.go
```diff
@@ -557,12 +559,21 @@ func RunCreateGoogleCloudStorage(cmd *cobra.Command, args []string) {
...
 		mandatoryProperties := []string{"AWS_ACCESS_KEY_ID", "AWS_SECRET_ACCESS_KEY"}
-
+		u, _ := url.Parse(endpoint)
+		if u.Scheme == "http" {
+			if sigVer == "v4" {
+				log.Fatalf("Non-secure endpoint works only with v2 signature-version. Please select signature version v2 for namespacestore")
+			}
+			sigVer = "v2"
+		} else {
+			sigVer = "v4"
```
jackyalbo, line 575: I think you need to set to "v4" only if the original is an empty string ("") no? In HTTPS, we support both versions, right?
```suggestion
			if sigVer == "" {
			       sigVer = "v4"
			}
```

jackyalbo (approved):

PR #1581, opened 2025-04-07
Handle low volume space for DB import 

dannyzaken:
### Explain the changes
- Check for the used space before initiating DB import to a new cluster.
  - If the used space is over 33 percent, request double the size for the PV
- Implemented new util function `ExecCommandInPod` to execute a command inside a pod
  - Using `ExecCommandInPod` running `df` in the old DB pod to get used percentage
  - Added permission to the noobaa-operator role to execute commands in `noobaa-db-pg-0` pod
- Also handling updates of storage size to support PVC expansion (for supporting storage-classes)

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/deploy/crds/noobaa.io_noobaas.yaml
+++ b/deploy/crds/noobaa.io_noobaas.yaml
@@ -1214,8 +1214,9 @@ spec:
                     type: object
                   dbMinVolumeSize:
                     description: |-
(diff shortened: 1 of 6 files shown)


liranmauda (approved):
LGTM

PR #1588, opened 2025-04-15
Added a flag DisableRoutes to manage openshift routes in noobaa

achouhan09:
### Explain the changes
1. Added logic to disable the creation of routes based on the `DisableRoutes` flag. If the flag is set to true, the route creation is skipped, ensuring that routes are only created when necessary, based on the flag value. This allows more control over whether routes are deployed in the cluster.

1.  JIRA: https://issues.redhat.com/browse/RHSTOR-7195

### Testing Instructions:
1. Manual testing instructions:

-  Need ocp cluster as routes supported by openshift (use cluster-bot).
-  Install ODF and create storage system.
-  Update the operator image (latest including the changes) everywhere in the below csv:
`$ oc edit csv mcg-operator.v4.18.1-rhodf` (please use image from quay/docker)
-  Update the noobaa CRD using below command:
`$ oc apply -f noobaa-operator/deploy/crds/noobaa.io_noobaas.yaml`
-  Edit noobaa instance and add `disableRoute: true` into `Spec`:
`$ oc edit noobaa noobaa`
-  Try deleting routes present, they should not be reconciled after setting `disableRoute: true`:
`$ oc get routes`

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -141,6 +141,9 @@ var DevEnv = false
 // DisableLoadBalancerService is used for setting the service type to ClusterIP instead of LoadBalancer
 var DisableLoadBalancerService = false
 
+// DisableRoutes is used to disable the reconciliation of openshift route resources
+var DisableRoutes = false
+
 // CosiDriverPath is the cosi socket fs path
 var CosiDriverPath = "/var/lib/cosi/cosi.sock"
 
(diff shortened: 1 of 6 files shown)


pkg/options/options.go
```diff
@@ -141,6 +141,9 @@ var DevEnv = false
 // DisableLoadBalancerService is used for setting the service type to ClusterIP instead of LoadBalancer
 var DisableLoadBalancerService = false
 
+// DisableRoutes is used to disables the reconciliation of openShift route resources
```
jackyalbo, line 144:
```suggestion
// DisableRoutes is used to disable the reconciliation of openShift route resources
```

jackyalbo (approved):

PR #1639, opened 2025-06-30
csv: add cpu, memory requests value for operator

Nikhil-Ladha:
Add cpu, memory requests value for operator

## Summary by CodeRabbit

* **Chores**
  * Added default CPU and memory resource requests for the noobaa-operator container to ensure more predictable resource allocation during deployment.

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -54,6 +54,9 @@ spec:
             limits:
               cpu: "250m"
               memory: "512Mi"
+            requests:
+              cpu: "100m"
+              memory: "256Mi"
           env:
             - name: OPERATOR_NAME
               value: noobaa-operator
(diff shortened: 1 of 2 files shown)


liranmauda:
Hi @Nikhil-Ladha 

I am not reviewing the actual need for this. I will leave that to @dannyzaken and @nimrod-becker 
But, the place to add this is not the deploy, the deploy is auto-generated when running:
`make gen ; make gen-api ; make` 
In order to add those things, you will need to add them in the proper CR (in this case, operator deployment) and then run the command above to generate the deployment.

Moreover, you will probably need to bump noobaa on the ocs-operator to add this code. You might even need a PR to add this field to the storagecluster on the ocs-operator, to avoid a situation where an upgrade will revert those changes.

PR #1642, opened 2025-07-01
Added github workflow to automatic update noobaa-core image tag

achouhan09:
### Explain the changes
1.  Github workflow to automatic update the noobaa-core ContainerImageTag in 'options.go'.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


.github/workflows/update-noobaa-core-tag.yaml
```diff
@@ -0,0 +1,63 @@
...
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  update-tag:
+    runs-on: ubuntu-latest
+    timeout-minutes: 90
+    env:
+      GITHUB_TOKEN: ${{ secrets.GHACTION_TOKEN_WRITE }}
```
Neon-White: Is there a reason to use `GHACTION_TOKEN_WRITE` over `GITHUB_TOKEN`?

.github/workflows/update-noobaa-core-tag.yaml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Checkout master
+        uses: actions/checkout@v4
+        with:
+          ref: master
+      
+      - name: Update ContainerImageTag in options.go
+        run: |
+          tag=${{ github.event.inputs.container_image_tag }}
+          safe_tag=$(printf '%s\n' "$tag" | sed 's/[&/]/\\&/g')
+          sed -i "s/ContainerImageTag = \".*\"/ContainerImageTag = \"${safe_tag}\"/" pkg/options/options.go
```
Neon-White: I'm not a fan of `sed` in general, and especially when escapes are making the patterns night impossible to understand.

I think it'll be better to use an action for this part, something along those lines -

```suggestion
    - name: Replace ContainerImageTag
      uses: jacobtomlinson/gha-find-replace@v3
      with:
        find: 'ContainerImageTag = ".*"'
        replace: 'ContainerImageTag = "${{ inputs.container_image_tag }}"'
        include: 'pkg/options/options.go'
        regex:   true

```

.github/workflows/update-noobaa-core-tag.yaml
aayushchouhan09: We need to merge the PR in this action, not sure if `GHACTION_TOKEN_WRITE` has such access. 
@tangledbytes Can you pls confirm if it is possible to merge the PR with this token.

tangledbytes: Honestly, I am not sure. But from the documentation it seems that `GITHUB_TOKEN` when given `permissions.pull-requests = true` does grant access to the `/merge` endpoint. Let's try just with `GITHUB_TOKEN` and let's see if that works, if not we can fix it in a subsequent PR, WDYT?

aayushchouhan09: Sure Thanks, I will try that

aayushchouhan09: @tangledbytes I have tested it we need a token with below permissions:
1. repo (Full control of private repositories)
2. workflow (Update GitHub Action workflows)
3. read:org (Read org and team membership, read org projects)

.github/workflows/update-noobaa-core-tag.yml
```diff
@@ -0,0 +1,63 @@
...
+          echo "pr_url=$PR_URL" >> "$GITHUB_OUTPUT"
+
+      
+      - name: Wait for PR checks and merge
+        run: |
+          echo "Waiting for checks to pass on: ${{ steps.create_pr.outputs.pr_url }}"
+          gh pr checks "${{ steps.create_pr.outputs.pr_url }}" --watch
+
+          echo "Merging PR..."
+          gh pr merge "${{ steps.create_pr.outputs.pr_url }}" --merge --admin --delete-branch
```
dannyzaken: Shouldn't you check if a rebase is needed before merging?

.github/workflows/update-noobaa-core-tag.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Commit & Push changes
+        run: |
+          git config --global user.email "github-action@noobaa.io"
+          git config --global user.name "NooBaa GitHub Action"
+          git checkout -b update-core-tag-${{ github.event.inputs.container_image_tag }}
+          git diff
+          git add pkg/options/options.go
+          git commit -m "chore: update noobaa-core image tag to ${{ github.event.inputs.container_image_tag }}"
+          git push origin update-core-tag-${{ github.event.inputs.container_image_tag }}
+
```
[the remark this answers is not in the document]
aayushchouhan09: This won't happen, as we plan to run it periodically (e.g., weekly) with a dynamically updated tag. Thanks

.github/workflows/update-noobaa-core-tag.yml
aayushchouhan09: Done, Thanks

.github/workflows/update-noobaa-core-tag.yml
```diff
@@ -0,0 +1,67 @@
...
+          replace: 'ContainerImageTag = "${{ github.event.inputs.container_image_tag }}"'
+          include: 'pkg/options/options.go'
+          regex: true
+
+      - name: Commit & Push changes
+        run: |
+          git config --global user.email "github-action@noobaa.io"
+          git config --global user.name "NooBaa GitHub Action"
+          git checkout -b update-core-tag-${{ github.event.inputs.container_image_tag }}
+          git diff
```
Neon-White: Does this serve any purpose?

.github/workflows/update-noobaa-core-tag.yml
```diff
@@ -0,0 +1,67 @@
+name: Update Noobaa-Core container image tag
+
+on:
```
Neon-White, line 3: Have we considered adding a `schedule:` option that runs this automatically on set time intervals?

.github/workflows/update-noobaa-core-tag.yml
```diff
@@ -0,0 +1,67 @@
...
+          ref: master
+
+      - name: Replace ContainerImageTag
+        uses: jacobtomlinson/gha-find-replace@v3
+        with:
+          find: 'ContainerImageTag = "[^"]*"'
+          replace: 'ContainerImageTag = "${{ github.event.inputs.container_image_tag }}"'
+          include: 'pkg/options/options.go'
+          regex: true
+
```
Neon-White, line 35: We might want to verify there's actually a diff before we try to create an empty commit. Something like -

```suggestion
- name: Check for real changes
  run: |
    if git diff --quiet pkg/options/options.go; then
      echo "Image tag is already about to date, canceling job'
      gh run cancel ${{ github.run_id }}
    fi
```

.github/workflows/update-noobaa-core-tag.yml
aayushchouhan09: I am planning to do it in separate PR once we are sure with the working, so for now it is manual.

.github/workflows/update-noobaa-core-tag.yml
aayushchouhan09: We will be having dynamic tags based on the current date so there is no such case of tag already upto date. But for safer side I will add this check. Thanks

.github/workflows/update-noobaa-core-tag.yml
aayushchouhan09: Not sure, I didn't try without it, I think git needs author identity to create commits.

Neon-White: Please note that my comment only addresses line 40 :) but the way GitHub shows it is indeed confusing

aayushchouhan09: We can remove it and add a check in place of it. Thanks

.github/workflows/update-noobaa-core-tag.yml
```diff
@@ -32,12 +32,19 @@ jobs:
           include: 'pkg/options/options.go'
           regex: true
 
-      - name: Commit & Push changes
+      - name: Check for changes
+        run: |
+          if git diff --quiet pkg/options/options.go; then
+            echo "Noobaa Core image tag is already up to date"
+            echo "No changes needed - skipping remaining steps"
+            exit 0
```
Neon-White: `exit` is insufficient, the workflow will simply move on to the next step - my original example utilizes `gh cancel`

aayushchouhan09: yes right, I will update it. Thanks

PR #1653, opened 2025-07-09
Added TopologyKey to noobaa CRD for multizone support

dannyzaken:
### Explain the changes
- Added `TopologyKey` to Noobaa Spec. This can be used for example to set `topology.kubernetes.io/zone` as the topology domain
- If a topologyKey is used and is different from the default `kubernetes.io/hostname` TopologyKey, adding another TopologySpreadConstraint with the additional key.
- Passing the TopologyKey to the DB cluster CR under affinity. This will be set to the podAntiAffinity of the DB pods.

1. https://issues.redhat.com/browse/DFBUGS-2826 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/apis/noobaa/v1alpha1/noobaa_types.go
+++ b/pkg/apis/noobaa/v1alpha1/noobaa_types.go
@@ -159,6 +159,10 @@ type NooBaaSpec struct {
 	// +optional
 	Affinity *corev1.Affinity `json:"affinity,omitempty"`
 
+	// TopologyKey (optional) the TopologyKey to pass as the domain for TopologySpreadConstraint and Affinity of noobaa components
+	// It is used by the endpoints and the DB pods to control pods distribution between topology domains (host/zone)
+	TopologyKey string `json:"topologyKey,omitempty"`
+
 	// ImagePullSecret (optional) sets a pull secret for the system image
 	// +optional
 	ImagePullSecret *corev1.LocalObjectReference `json:"imagePullSecret,omitempty"`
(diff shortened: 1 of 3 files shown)


deploy/crds/noobaa.io_noobaas.yaml
```diff
@@ -984,6 +984,11 @@ spec:
                         type: array
                         x-kubernetes-list-type: atomic
                     type: object
+                  topologyKey:
```
naveenpaul1, line 987: Could you please update the doc doc/noobaa-crd.md.

dannyzaken: I will create a separate PR with changes to noobaa-crd.md, that also includes updates regarding the dbSpec

dannyzaken:
@naveenpaul1, I noticed now that we do not update the topologySpreadConstraints if it is already set. Since we want it to be updated if a topologyKey is supplied, I think we should always reconcile it unless the annotation is set, WDYT?

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -242,6 +242,23 @@ type NooBaaSpec struct {
...
+	// Describes pod affinity scheduling rules (e.g. co-locate this pod in the same node, zone, etc. as some other pod(s)).
+	// +optional
+	PodAffinity *corev1.PodAffinity `json:"podAffinity,omitempty" protobuf:"bytes,2,opt,name=podAffinity"`
+	// Describes pod anti-affinity scheduling rules (e.g. avoid putting this pod in the same node, zone, etc. as some other pod(s)).
+	// +optional
+	PodAntiAffinity *corev1.PodAntiAffinity `json:"podAntiAffinity,omitempty" protobuf:"bytes,3,opt,name=podAntiAffinity"`
+
+	// TopologyKey (optional) the TopologyKey to pass as the domain for TopologySpreadConstraint and Affinity of noobaa components
+	// It is used by the endpoints and the DB pods to control pods distribution between topology domains (host/zone)
+	TopologyKey string `json:"topologyKey,omitempty"`
```
liranmauda, line 260: Add // +optional

PR #1660, opened 2025-07-16
Added prometheus alert rules for DB volume size capacity

dannyzaken:
### Explain the changes
1. Added alert rules regarding DB capacity reaching 80% (warning) and 90% (critical)
2. The alert is triggered after 5 minutes when the usage is over the threshold. 

1. https://issues.redhat.com/browse/RHSTOR-7393 

### Testing Instructions:
1. Fill DB volume space to over 80% or 90%
2. After 5 minutes, the relevant alert should be triggered in the openshift console.

- [ ] Doc added/updated
- [ ] Tests added


deploy/internal/prometheus-rules.yaml
```diff
@@ -245,3 +245,65 @@ spec:
       for: 5m
       labels:
         severity: critical
+  - name: noobaa-db-alert.rules
+    rules:
+      - alert: NooBaaDatabaseReachingCapacity
+        annotations:
+          description: The NooBaa database on pod noobaa-db-pg-cluster-1 is using 80% of its PVC requested size.
+          message: NooBaa NooBaa database on pod noobaa-db-pg-cluster-1 is using 80% of its PVC capacity.
```
liranmauda, line 253: Is the pod name always `noobaa-db-pg-cluster-1` ?

liranmauda (commented):
There are 2 alerts and each for cluster-1 and 2. 
Could we have more than those (e.g cluster-3)? 

And will we always use the namespace openshift-storage ?

deploy/internal/prometheus-rules.yaml
dannyzaken: it's always these names. I tried generalizing with regex, but it just didn't work. I will try again after this PR as a fix.
As for the namespace, I think we can remove it. I will test it and update the PR.

PR #1687, opened 2025-08-25
Modify the `golangci-lint` pre-commit and Makefile flow

Neon-White:
### Explain the changes
1. Flows related to `golandci-lint` were moved from `Makefile` and `pre-commit` to a unified script (`run-golangci-lint.sh`) to reduce bloat in external files, and also allow a central maintenance spot (and source of truth) for anything related to the linter
2. Exit codes were changed to only block commits/`make` commands in case of linting problems in reviewed files
3. Any exit code unrelated to linting problems is now printed as a warning to the git log/console, but is not blocking

## Summary by CodeRabbit

* **Chores**
  * Introduced a centralized lint runner; pre-commit hook and build/CI lint now delegate to it.
  * Pre-commit flow now relies on the centralized runner to check staged Go changes and ensure the linter is available.
  * CI/build linting runs through the centralized runner and reports clear pass/fail status.

* **Refactor**
  * Removed inline linter install/version checks and PATH modifications from hooks and Makefile.

* **Note**
  * No user-facing features changed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -184,18 +184,7 @@ test: lint test-go
 .PHONY: test
 
 lint: gen
-	@echo ""
-	@if [ ! -f "$(GOBIN)/golangci-lint" ] || [ "v2.1.6" != "$$($(GOBIN)/golangci-lint --version 2>/dev/null | grep -o 'v[0-9]\+\.[0-9]\+\.[0-9]\+' || echo '')" ]; then \
-		echo "Installing golangci-lint v2.1.6..."; \
-		go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6; \
-	fi
-	@if [ -n "$(FILES)" ]; then \
-		echo "Running lint on changed files: $(FILES)"; \
-		$(GOBIN)/golangci-lint run --config .golangci.yml $(FILES); \
-	else \
-		echo "Running lint on all files"; \
-		$(GOBIN)/golangci-lint run --config .golangci.yml; \
-	fi
+	@./scripts/run-golangci-lint.sh makefile
 	@echo "✅ lint"
 .PHONY: lint
 
--- a/scripts/run-golangci-lint.sh
+++ b/scripts/run-golangci-lint.sh
@@ -1,0 +1,130 @@
+#!/bin/sh
+
+# Unified golangci-lint runner script
+# Usage: ./scripts/run-golangci-lint.sh <mode>
+# Modes: precommit, makefile
+
+# Uncomment this to see the commands being run
+# Note: We don't use 'set -e' because we need to handle golangci-lint's
+# exit codes manually to distinguish between linting issues and other errors
+# set -x
+
+# Exit code used by golangci-lint when linting issues are found
+# Allows us to distinguish between linting issues and other errors
+readonly ISSUES_EXIT_CODE=42
+
+MODE="${1}"
+
+# Safely check whether Go is installed and set the GOPATH
+# Also check for GOBIN and add it to the PATH in case the binary was installed to a non-standard location
+ if command -v go >/dev/null 2>&1; then
+     GOBIN="$(go env GOBIN)"
+     GOPATH_BIN="$(go env GOPATH)/bin"
+     [ -n "${GOBIN}" ] && export PATH="${PATH}:${GOBIN}"
+     export PATH="${PATH}:${GOPATH_BIN}"
+ fi
+
+# Function to install golangci-lint (for makefile mode)
+install_golangci_lint() {
+    echo "Installing the latest golangci-lint with local toolchain"
+    if ! go install -a "github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest"
+    then
+        echo "⚠️ Failed to install golangci-lint"
+        exit 0
+    fi
+}
+
+# Function to run golangci-lint in precommit mode
+run_precommit_lint() {
+    echo "Running golangci-lint in precommit mode..."
+    
+    # Check if golangci-lint is available - exit gracefully if it isn't
+    if ! command -v golangci-lint >/dev/null 2>&1
+    then
+        echo "⚠️ golangci-lint not found – run 'make lint' to install it automatically"
+        exit 0
+    fi
+    
+    # Collect staged Go files as positional args
+    # cached: retrieve only staged files
+    # name-only: retrieve only the names of the files
+    # diff-filter=ACMR: retrieve only the files that have been added, copied, modified, or renamed
+    # grep -E '\.go$': retrieve only the files that end with .go
+    # || true: if there are no staged Go files, return a zero-status exit code
+    set -- $(git diff --cached --name-only --diff-filter=ACMR | grep -E '\.go$' || true)
+    # If there are no staged Go files, exit successfully
+    if [ "$#" -eq 0 ]; then
+        echo "✅ No staged Go files were found, exiting"
+        exit 0
+    fi
+    
+    echo "Running golangci-lint on staged files..."
+    BASE=$(git rev-parse --verify HEAD 2>/dev/null || echo "")
+    
+    if [ -z "${BASE}" ]
+    then
+        # Initial commit – lint only staged files
+        golangci-lint run --issues-exit-code=${ISSUES_EXIT_CODE} --config .golangci.yml "${@}"
+    else
+        # Lint only staged changes vs HEAD
+        golangci-lint run --issues-exit-code=${ISSUES_EXIT_CODE} --config .golangci.yml --new-from-rev="${BASE}" "${@}"
+    fi
+    
+    # Store the exit code
+    LINT_EXIT_CODE=${?}
+    
+    # Handle exit codes properly
+    if [ ${LINT_EXIT_CODE} -eq ${ISSUES_EXIT_CODE} ]
+    then
+        echo "❌ golangci-lint found linting issues in staged files. Please fix them before committing."
+        exit 1
+    elif [ ${LINT_EXIT_CODE} -ne 0 ]
+    then
+        echo "⚠️  golangci-lint encountered an error (exit code: ${LINT_EXIT_CODE})"
+        exit 0
+    fi
+    
+    echo "✅ golangci-lint passed!"
+}
+
+# Function to run golangci-lint in makefile mode
+run_makefile_lint() {
+    echo "Running golangci-lint in makefile mode..."
+    
+    # Install golangci-lint if needed
+    install_golangci_lint
+    
+    echo "Running lint on all files"
+    golangci-lint run --issues-exit-code=${ISSUES_EXIT_CODE} --config .golangci.yml
+    
+    # Store the exit code
+    LINT_EXIT_CODE=${?}
+    
+    # Handle exit codes properly
+    if [ ${LINT_EXIT_CODE} -eq ${ISSUES_EXIT_CODE} ]
+    then
+        echo "❌ golangci-lint found linting issues. Please fix them."
+        exit 1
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -190,9 +190,9 @@ golangci-lint: gen
 
 lint: gen
 	@echo ""
-	@if [ ! -f "$(GOBIN)/golangci-lint" ] || [ "v2.1.6" != "$$($(GOBIN)/golangci-lint --version 2>/dev/null | grep -o 'v[0-9]\+\.[0-9]\+\.[0-9]\+' || echo '')" ]; then \
-		echo "Installing golangci-lint v2.1.6..."; \
-		go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6; \
+	@if [ ! -f "$(GOBIN)/golangci-lint" ] || ! $(GOBIN)/golangci-lint --version 2>/dev/null | grep -q "version 2.4.0"; then \
```
liranmauda: Why do we pinpoint the version to a specific version?
Why don't we use the latest?

Makefile
```diff
@@ -190,9 +190,9 @@ golangci-lint: gen
...
+		echo "Installing golangci-lint v2.4.0..."; \
+		go install -a github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.4.0; \
```
liranmauda: Why don't we have else for upgrades?

Makefile
Neon-White: We pin the version to ensure reproducibility and stability across all environments. Using `latest` can introduce unexpected lint errors or even break the build if upstream changes their rules or drops support. Pinning gives us consistency, and we can explicitly upgrade when we're ready to handle any new rules or changes

Makefile
Neon-White: Upgrades (and downgrades) are handled in this `if` - if the version is not 2.4.0, then 2.4.0 is installed (which replaces the previous version), and the `-a` flag forces rebuilding packages that are already up-to-date, to ensure we get the desired result

Makefile
liranmauda: I think we should not pin it.
If a new rule is popping and failing, we should handle it on demand.

Makefile
liranmauda: I am not sure I follow.
The if condition will only happen when the version is not 2.4.0 explicitly.
If you have a v2.4.0 it will not meet the condition and will not rebuild.
Regarding upgrades, I think it should be to the latest and the check for the version should not be there.
So eventually, in any case that you already have the `golangci-lint` it will try to upgrade it to the latest.

And still this will not help for a rebuild. so we need to figure out a way to rebuild it regardless.

Neon-White: >And still this will not help for a rebuild. so we need to figure out a way to rebuild it regardless.

According to my observations, the error is fixed by installing a newer version of `golangci-lint` from GitHub

You can reproduce it by deleting the `golangci-lint` directory as you did yesterday, and then running `go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6` - the error should occur

Then, without deleting any files, you can run `go install -a github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest` (which then mentions `go: github.com/golangci/golangci-lint/v2@v2.4.0 requires go >= 1.24.0; switching to go1.24.6`) - the problem should then be fixed. I don't think a separate path for rebuilding is required - the rebuild takes place naturally with `go install`

liranmauda: It is not helping when switching branches.
The issue is not a new version of `golangci-lint`, the issue is the way it was built.

We cannot perform a manual step every time we switch branches or update a go version.

liranmauda (changes requested):
Need to make sure that when we already have a `golangci-lint` binary, we will not decline commits just because the linter was built using an older version of golang.

It applies when switching branches and when updating golang version.
We need to make sure that the latest golinet is always installed, and the golang it was built with will not create that error.

scripts/run-golangci-lint.sh
```diff
@@ -0,0 +1,133 @@
...
+    # Install golangci-lint if needed
+    install_golangci_lint
+    
+    GOBIN=$(go env GOBIN)
+    if [ -z "$GOBIN" ]; then
+        GOBIN=$(go env GOPATH)/bin
+    fi
+    
+    # Check if specific files were provided
+    if [ $# -gt 0 ]; then
```
liranmauda: Why do we need this if?
If ${@} is empty, it will just run it without files.

liranmauda (commented):
2 general styling comments:
1. All variable calls should be wrapped with {}, for example: a=${variable}
2. The syntax of `if... ; then` never made sense to me, `;` means new line, then it is better to do a new line.

scripts/run-golangci-lint.sh
Neon-White: The `if` is required for us to know whether we need to lint specific files that were provided by the user, or all files in the project -

At this point in the code, `$#` represents the amount of parameters passed to the script after the `mode` parameter.
The only expected parameters are files to lint - meaning that if `$#` is greater than 0, there are specific files to lint - and the names of these files can be accessed via `$@`.

If `$#` is 0, we have no specific files to lint, we do not pass any specific files to lint - and thus run the linter on all of the project files.
If `$#` is greater than 0, `$@` is a list of filename that we then run the linter on.

scripts/run-golangci-lint.sh
```diff
@@ -0,0 +1,147 @@
...
+    
+    # Check if specific files were provided
+    if [ ${#} -gt 0 ];
+    then
+        echo "Running lint on specified files: ${*}"
+        "${GOBIN}/golangci-lint" run --issues-exit-code=${ISSUES_EXIT_CODE} --config .golangci.yml "${@}"
+    else
+        echo "Running lint on all files"
+        "${GOBIN}/golangci-lint" run --issues-exit-code=${ISSUES_EXIT_CODE} --config .golangci.yml
+    fi
```
liranmauda: If the purpose is just to keep the printing different, then just keep the printing in the if statement.
When the ${@} is empty, the command is the same. 
```suggestion
    # Check if specific files were provided
    if [ ${#} -gt 0 ];
    then
        echo "Running lint on specified files: ${*}"
    else
        echo "Running lint on all files"
    fi
     "${GOBIN}/golangci-lint" run --issues-exit-code=${ISSUES_EXIT_CODE} --config .golangci.yml "${@}"
```

Neon-White: >When the ${@} is empty, the command is the same.

Good point, I didn't realize it earlier. I'll change the code

Makefile
```diff
@@ -185,16 +185,12 @@ test: lint test-go
...
-		go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6; \
-	fi
 	@if [ -n "$(FILES)" ]; then \
 		echo "Running lint on changed files: $(FILES)"; \
-		$(GOBIN)/golangci-lint run --config .golangci.yml $(FILES); \
+		./scripts/run-golangci-lint.sh makefile $(FILES); \
 	else \
 		echo "Running lint on all files"; \
-		$(GOBIN)/golangci-lint run --config .golangci.yml; \
+		./scripts/run-golangci-lint.sh makefile; \
```
liranmauda: Where do we get the FILES into the variable? 
I do not see any place where we define the FILES variable or get it.

If there is such an intention, we can add it to the script and remove this condition from the makefile

Neon-White: `FILES` is passed externally - e.g.
```
make lint FILES="pkg/api.go pkg/util.go"
```
or 
```
export FILES="pkg/api.go pkg/util.go"
make lint
```

If we'll remove it from the Makefile, we won't be able to run `make lint` on specific files - only on the entire project; so I don't think we can move this logic to the script. 

I renamed the variable to be more clear, and added a comment that explains its origin. What do you think?

liranmauda: @Neon-White If no one will know about it then no one will use it.
We are not calling it with specific files anywhere, and I do not think it is needed. 
Moreover, in the event of a new linter, we should not limit it per a file.

scripts/run-golangci-lint.sh
```diff
@@ -0,0 +1,146 @@
...
+# Function to install golangci-lint (for makefile mode)
+install_golangci_lint() {
+    GOBIN=$(go env GOBIN)
+    if [ -z "${GOBIN}" ]
+    then
+        GOBIN=$(go env GOPATH)/bin
+    fi
+    
+    echo "Installing the latest golangci-lint with local toolchain"
+    if ! GOTOOLCHAIN=local go -a install "github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest"
```
liranmauda:
```suggestion
    if ! GOTOOLCHAIN=local go install -a "github.com/golangci/golangci-lint/v2/cmd/golangci-lint@latest"
```

Makefile
```diff
@@ -185,16 +185,13 @@ test: lint test-go
...
 lint: gen
 	@echo ""
-	@if [ ! -f "$(GOBIN)/golangci-lint" ] || [ "v2.1.6" != "$$($(GOBIN)/golangci-lint --version 2>/dev/null | grep -o 'v[0-9]\+\.[0-9]\+\.[0-9]\+' || echo '')" ]; then \
-		echo "Installing golangci-lint v2.1.6..."; \
-		go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.1.6; \
-	fi
-	@if [ -n "$(FILES)" ]; then \
-		echo "Running lint on changed files: $(FILES)"; \
-		$(GOBIN)/golangci-lint run --config .golangci.yml $(FILES); \
+# FILES_TO_LINT is passed externally to the makefile, e.g. make lint FILES="pkg/..."
```
liranmauda: The example is misleading:
```suggestion
# FILES_TO_LINT is passed externally to the makefile, e.g. make lint FILES_TO_LINT="pkg/..."
```

Makefile
Neon-White: Alright, it could be a nice extra, but it'll likely remain unused and it's better to have fewer code paths to maintain. Thus, I removed the option to `make lint` on specific files 👍🏻

Makefile
Neon-White: Correct; since we opted to remove the option to pass specific files, I removed the comment as well

liranmauda:
@coderabbit review

scripts/run-golangci-lint.sh
```diff
@@ -0,0 +1,140 @@
...
+# Uncomment this to see the commands being run
+# Note: We don't use 'set -e' because we need to handle golangci-lint's
+# exit codes manually to distinguish between linting issues and other errors
+# set -x
+
+# Exit code used by golangci-lint when linting issues are found
+# Allows us to distinguish between linting issues and other errors
+readonly ISSUES_EXIT_CODE=42
+
+MODE="${1}"
```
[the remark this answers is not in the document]
Neon-White: The makefile and script no longer accept file lists

scripts/run-golangci-lint.sh
```diff
@@ -0,0 +1,130 @@
...
+    
+    # Collect staged Go files as positional args
+    # cached: retrieve only staged files
+    # name-only: retrieve only the names of the files
+    # diff-filter=ACMR: retrieve only the files that have been added, copied, modified, or renamed
+    # grep -E '\.go$': retrieve only the files that end with .go
+    # || true: if there are no staged Go files, return a zero-status exit code
+    set -- $(git diff --cached --name-only --diff-filter=ACMR | grep -E '\.go$' || true)
+    # If there are no staged Go files, exit successfully
+    if [ "$#" -eq 0 ]; then
```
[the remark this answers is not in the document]
liranmauda: @coderabbitai There is no value in this, there are no, and there will never be files with spaces in the name

PR #1698, opened 2025-09-10
Mongo pool upgrade

naveenpaul1:
### Explain the changes
1. Change default pool.ResourceType name to `backingstores`

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added


pkg/nb/types.go
```diff
@@ -286,7 +286,7 @@ type ListAccountsReply struct {
 // ListBcuketsParams is the params to account_api.list_buckets()
 type ListBucketsParams struct {
 	ContinuationToken *string `json:"continuation_token,omitempty"`
-	MaxBuckets *int `json:"max_buckets,omitempty"`
+	MaxBuckets        *int    `json:"max_buckets,omitempty"`
```
jackyalbo: is there any real change in this file? if no, can we leave it with the old format and take it away from this PR?

PR #1888, opened 2026-04-23
Bumping deps to avoid CVE (30/06/2026)

liranmauda:
- Bumping deps to avoid CVE (30/06/2026)


liranmauda:
Converting to draft.
KEDA blocks us as it is missing support, and will only add it in Keda v2.20

PR #1889, opened 2026-04-23
ObjectBucketName in OBC spec is missing

naveenpaul1:
### Explain the changes
1. Restore missing objectBucketName if OBC is bound and objectBucketName is empty

1. https://redhat.atlassian.net/browse/DFBUGS-2323

### Testing Instructions:
1. Create an OBC backed by NooBaa
2. Wait for it to be Bound
3. Remove `objectBucketName` from OBC spec and verify

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where the objectBucketName field could remain missing in Bound ObjectBucketClaim resources. The system now automatically restores this missing spec field during update operations when a claim is in Bound status, ensuring consistent and complete claim configuration.

### DIFF
--- a/pkg/controller/obc/obc_controller.go
+++ b/pkg/controller/obc/obc_controller.go
@@ -2,6 +2,7 @@ package obc
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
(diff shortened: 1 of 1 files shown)


romayalon:
@naveenpaul1 it seems like this code is fixing a symptom but what is the root cause - why the ObjectBucketName is empty? can we fix the root cause?

naveenpaul1:
@romayalon, In issue its mentioned that an external component (i.e. odf-multicluster-orchestrator) removes the `objectBucketName` field from the OBC spec. Its not happening from NooBaa side.

naveenpaul1:
After confirming with lib-bucket-provisioner team, closing this PR
Related ticket is also closed.

comment from lib-bucket-provisioner team : 
```
If odf-multicluster-orchestrator is the component that is wrongly removing the field, 
why is the fix going into lib-bucket-provisioner instead of fixing odf-multicluster-orchestrator? 
lib-bucket-provisioner can be updated to help, but it seems to me that the most appropriate fix is to
make sure ODF components are stripping critical info. With only the lib-bucket-provisioner fix, there
will be unnecessary churn as the orchestrator removes the field occasionally and Noobaa/Rook have to re-reconcile the OBC to repair it.
```

PR #1899, opened 2026-05-04
STS | Azure provider spec missing Region

naveenpaul1:
### Explain the changes
1. Added Azure provider spec missing Region
2. The OCP console does not prompt for a region, and the Azure CCO documentation hardcoded region
			

1. https://redhat.atlassian.net/browse/DFBUGS-6456

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Azure credential requests now default to the Central US region when no region is specified. The selected region is applied to STS credential provisioning so cluster credential requests consistently target Central US by default, reducing misconfiguration and improving reliability when the region environment variable is not set.


VershaAgrawal (approved):
LGTM. Please verify if any test cases need to be updated.

PR #1900, opened 2026-05-04
Add runtime-configurable OPERATOR_LOG_LEVEL and reduce log noise

kajalpareek-lab:
## Summary

- Demote Update/Generic event predicate logs from Info to Debug level, eliminating the main source of log noise at default verbosity
- Add separate `OPERATOR_LOG_LEVEL` key in `noobaa-config` ConfigMap — controls operator log level without triggering core/endpoint pod rollouts
- Supported levels: `warn`, `info` (default), `debug`
- Changes applied at runtime via reconcile, no operator restart needed

Fixes: https://redhat.atlassian.net/browse/DFBUGS-4372

## Usage

```bash
kubectl patch configmap noobaa-config -n <namespace> -p '{"data":{"OPERATOR_LOG_LEVEL":"warn"}}'
```

## Test plan

- [x] Verified `OPERATOR_LOG_LEVEL: warn` suppresses all Info/Debug messages
- [x] Verified changing `OPERATOR_LOG_LEVEL` does not cause core/endpoint pod rollouts
- [x] Verified `NOOBAA_LOG_LEVEL` still works independently for core/endpoints

## Summary by CodeRabbit

* **New Features**
  * Operator log level is independently configurable via a new startup flag and OPERATOR_LOG_LEVEL in the config; changes take effect after the next reconcile without restarting pods
  * Operator config changes do not trigger core/endpoint pod rollouts (operator key excluded from rollout hash)

* **Behavior Changes**
  * Update/Generic event logs are quieter by default (demoted to debug) unless operator log level is set to debug

* **Documentation**
  * Added docs describing OPERATOR_LOG_LEVEL, supported values (warn, info, debug) and runtime examples

[Review Change Stack](https://app.coderabbit.ai/change-stack/noobaa/noobaa-operator/pull/1900)

### DIFF
--- a/doc/operator-log-level.md
+++ b/doc/operator-log-level.md
@@ -1,0 +1,97 @@
+# Operator Log Level Configuration
+
+## Design
+
+Add a separate `OPERATOR_LOG_LEVEL` key in the `noobaa-config` ConfigMap to control
+operator log verbosity independently from core/endpoint pods. Changing this key does
+**not** trigger a pod rollout for core or endpoint pods.
+
+Additionally, noisy Update/Generic event predicate logs are demoted from Info to Debug
+level, so they only appear when `OPERATOR_LOG_LEVEL` is set to `debug`.
+
+### Supported Values
+
+| Value | logrus Level | Behavior |
+|-------|-------------|----------|
+| `warn` | WarnLevel | Minimal logging: only warnings and errors |
+| `info` | InfoLevel | **Default.** Create/Delete events, phase transitions, reconcile start/done |
+| `debug` | DebugLevel | Verbose: includes Update/Generic event predicates, all debug traces |
+
+### Config Source
+
+```yaml
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: noobaa-config
+data:
+  NOOBAA_LOG_LEVEL: "default_level"   # controls core + endpoint pods
+  OPERATOR_LOG_LEVEL: "info"           # controls operator process only
+```
+
+`OPERATOR_LOG_LEVEL` is excluded from the ConfigMap hash used in pod-template
+annotations. The operator computes a hash of the ConfigMap data and stores it in
+the `noobaa.io/configmap-hash` annotation on core/endpoint pod templates. When
+this hash changes, Kubernetes detects a spec change and triggers a rolling
+restart. By filtering out `OPERATOR_LOG_LEVEL` before computing the hash,
+changing the operator's log verbosity does not cause core/endpoint pod rollouts.
+
+### CLI Flag
+
+A separate `--operator-log-level` flag controls the initial operator log level at startup,
+independent from `--debug-level` (which controls core/endpoint via `NOOBAA_LOG_LEVEL`).
+
+```bash
+noobaa-operator operator --operator-log-level=warn
+```
+
+Once the first reconcile completes and the ConfigMap is read, `OPERATOR_LOG_LEVEL` takes over.
+
+### How to Change at Runtime
+
+```bash
+# Quiet the operator (warnings and errors only)
+kubectl patch configmap noobaa-config -n <namespace> \
+  -p '{"data":{"OPERATOR_LOG_LEVEL":"warn"}}'
+
+# Verbose debugging
+kubectl patch configmap noobaa-config -n <namespace> \
+  -p '{"data":{"OPERATOR_LOG_LEVEL":"debug"}}'
+
+# Back to default
+kubectl patch configmap noobaa-config -n <namespace> \
+  -p '{"data":{"OPERATOR_LOG_LEVEL":"info"}}'
+```
+
+The operator picks up the change on the next reconcile cycle (no restart needed).
+Core and endpoint pods are **not** affected.
+
+### Example Output
+
+**Before the change** (`OPERATOR_LOG_LEVEL` = `info`, Update events logged at Info):
+
+```
+time="2025-10-10T13:20:54Z" level=info msg="Update event detected for noobaa (openshift-storage), queuing Reconcile"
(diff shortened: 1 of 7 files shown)


pkg/options/options.go
```diff
@@ -300,7 +300,7 @@ func init() {
 	)
 	FlagSet.StringVar(
 		&DebugLevel, "debug-level",
-		DebugLevel, "The type of debug sets that the system prints (all, nsfs, warn, default_level)",
+		DebugLevel, "The type of debug sets that the system prints (all, nsfs, warn, info, debug, default_level). Also sets the initial operator log level.",
```
liranmauda: Core does not have info or debug.
https://github.com/noobaa/noobaa-core/blob/master/src/util/debug_config.js

doc/operator-log-level.md
```diff
@@ -0,0 +1,63 @@
...
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: noobaa-config
+data:
+  NOOBAA_LOG_LEVEL: "default_level"   # controls core + endpoint pods
+  OPERATOR_LOG_LEVEL: "info"           # controls operator process only
+```
+
+`OPERATOR_LOG_LEVEL` is excluded from the ConfigMap hash used in pod-template
```
shirady, line 32: > pod-template annotations

I think I missed something... I saw it in the previous YAML data -> OPERATOR_LOG_LEVEL which annotation?

doc/operator-log-level.md
```diff
@@ -0,0 +1,63 @@
...
+```bash
+# Quiet the operator (warnings and errors only)
+kubectl patch configmap noobaa-config -n <namespace> \
+  -p '{"data":{"OPERATOR_LOG_LEVEL":"warn"}}'
+
+# Verbose debugging
+kubectl patch configmap noobaa-config -n <namespace> \
+  -p '{"data":{"OPERATOR_LOG_LEVEL":"debug"}}'
+
+# Back to normal
```
shirady: normal? I thought it was the default
```suggestion
# Back to default
```

pkg/util/util.go
```diff
@@ -902,6 +902,20 @@ func InitLogger(lvl logrus.Level) {
 	})
 }
 
+// OperatorLogLevel maps a string log level name to a logrus.Level.
+// Accepted values: "warn", "info", "debug". Anything else (including
+// "default_level") maps to InfoLevel.
```
shirady: What is the "default_level"? (the default is "info" string)

pkg/options/options.go
```diff
@@ -117,6 +117,9 @@ var PostgresSSLCert = ""
 // DebugLevel can be used to override the default debug level
 var DebugLevel = "default_level"
 
+// OperatorLogLevel controls the operator process log verbosity (warn/info/debug)
+var OperatorLogLevel = "info"
```
shirady, line 121: I wonder if it would be more convenient to use the same suffix as `DebugLevel`.
For searching purposes.

```suggestion
var OperatorDebugLevel= "info"
```

pkg/system/phase2_creating.go
```diff
@@ -1634,7 +1634,14 @@ func (r *Reconciler) SetDesiredCoreAppConfig() error {
...
 		r.CoreAppConfig.Annotations = make(map[string]string)
 	}
-	r.CoreAppConfig.Annotations["noobaa.io/configmap-hash"] = util.GetCmDataHash(r.CoreAppConfig.Data)
+	operatorOnlyKeys := map[string]bool{"OPERATOR_LOG_LEVEL": true}
+	coreData := make(map[string]string, len(r.CoreAppConfig.Data))
+	for k, v := range r.CoreAppConfig.Data {
+		if !operatorOnlyKeys[k] {
+			coreData[k] = v
+		}
+	}
```
shirady, line 1645: Could you explain this part?

shirady (commented):
I had a couple of minor comments/questions

doc/operator-log-level.md
kajalpareek-lab: The annotation is noobaa.io/configmap-hash  it's stamped on core and endpoint pod templates. When this hash changes, pods restart. We exclude OPERATOR_LOG_LEVEL from the hash so changing the operator's log level doesn't restart those pods.

pkg/options/options.go
kajalpareek-lab: kept it as OperatorLogLevel since the operator supports warn/info/debug (actual log levels), while core's DebugLevel is more of a debug config preset (all/nsfs/warn/default_level). Happy to rename if you prefer consistency.

pkg/system/phase2_creating.go
kajalpareek-lab: This filters out operator-only keys before computing the ConfigMap hash. The hash (noobaa.io/configmap-hash) is stamped on core and endpoint pod templates — when it changes, those pods get a rolling restart. Without this filter, changing OPERATOR_DEBUG_LEVEL would change the hash and needlessly restart core/endpoint pods. By excluding it from the hash, only core-relevant config changes trigger pod rollouts.

doc/operator-log-level.md
shirady: @kajalpareek-lab Could you add this explanation to the doc?

pkg/system/phase2_creating.go
shirady: Could you add a comment in the code so it would be clear for other?

kajalpareek-lab: Sure.

doc/operator-log-level.md
kajalpareek-lab: added.

shirady (approved):
LGTM

PR #1910, opened 2026-05-13
fix(system): clear POSTGRES_HOST in ExternalPgSecret mode

### LINKED ISSUE #1909: POSTGRES_HOST default value not overridden in ExternalPgSecret mode
When `spec.externalPgSecret` is set on a NooBaa CR, the operator mounts the external secret and configures `POSTGRES_HOST_PATH=/etc/postgres-secret/host`, but the `POSTGRES_HOST` env var keeps the hard-coded default `"noobaa-db-pg-0.noobaa-db-pg"` from [deploy/internal/statefulset-core.yaml].

The `setDesiredCoreEnv` switch in [pkg/system/phase2_creating.go] has branches for `shouldReconcileStandaloneDB()` and `shouldReconcileCNPGCluster()` but no branch for `ExternalPgSecret != nil`, so the value leaks into the core/endpoint pods and connections fail.

Predecessor (closed): #1689.

I'd like to propose a minimal fix: mirror the CNPG branch behavior for ExternalPgSecret. Happy to open a PR.

Skiepp:
### Describe the Problem

When a `NooBaa` CR uses `spec.externalPgSecret` to connect to an externally
managed PostgreSQL database, the `POSTGRES_HOST` env var on the resulting
`noobaa-core` StatefulSet and `noobaa-endpoint` Deployment keeps the
hard-coded default `"noobaa-db-pg-0.noobaa-db-pg"` baked into
`deploy/internal/statefulset-core.yaml` (line 121-122). As a result the
core process tries to connect to a non-existent host and the system never
comes up against the external DB.

The `setDesiredCoreEnv()` switch in `pkg/system/phase2_creating.go`
(lines 513-520) already overrides `POSTGRES_HOST` for the
`shouldReconcileStandaloneDB()` and `shouldReconcileCNPGCluster()` cases,
but has no branch for `r.NooBaa.Spec.ExternalPgSecret != nil`, leaving the
YAML default untouched.

A symptom of this bug is users having to manually patch the workloads
post-install (e.g. `kubectl set env sts/noobaa-core POSTGRES_HOST=<real-host>`),
which is racy because the operator re-applies the default at every reconcile.

### Explain the changes

1. Added an `else if r.NooBaa.Spec.ExternalPgSecret != nil` branch to the
   `POSTGRES_HOST` case in `setDesiredCoreEnv()`. The branch mirrors the
   existing CNPG behavior: `Value` and `ValueFrom` are cleared so the core
   process falls back to reading the host from `POSTGRES_HOST_PATH`
   (`/etc/postgres-secret/host`), which is already wired in the same
   function (lines 591-592) and mounted from the external secret in
   `phase2_creating.go` (lines 706-710) for the core and
   `phase4_configuring.go` (lines 563-569) for the endpoint.

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved host configuration conflicts between managed and external PostgreSQL setups by clearing overridden host entries when the system is not managing the standalone database, ensuring the application reads the database host from the mounted secret.
  * Clarified configuration behavior so both managed and external database deployments consistently provide the host via secrets, reducing misconfiguration risk.

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -512,8 +512,9 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 		case "POSTGRES_HOST":
 			if r.shouldReconcileStandaloneDB() {
 				c.Env[j].Value = r.NooBaaPostgresDB.Name + "-0." + r.NooBaaPostgresDB.Spec.ServiceName + "." + r.NooBaaPostgresDB.Namespace + ".svc"
-			} else if r.shouldReconcileCNPGCluster() {
-				// clear env. it will be passed by mounting the secret
+			} else {
+				// CNPG and external-pg both ship the host via mounted secret;
+				// clear the default so noobaa-core reads POSTGRES_HOST_PATH
 				c.Env[j].Value = ""
 				c.Env[j].ValueFrom = nil
 			}


Skiepp:
Hi @dannyzaken, sorry to bother you 🙂 just wondering if you had a chance to review the PR.

Skiepp:
Hi @dannyzaken  I noticed you triggered the checks. I'm not sure the failure is directly related to this PR.

Please let me know if I can help investigate or provide any support. Thanks!

pkg/system/phase2_creating.go
```diff
@@ -517,6 +517,10 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
 				// clear env. it will be passed by mounting the secret
 				c.Env[j].Value = ""
 				c.Env[j].ValueFrom = nil
+			} else if r.NooBaa.Spec.ExternalPgSecret != nil {
```
dannyzaken: This case is the same as for `r.shouldReconcileCNPGCluster`. If I'm not mistaken, you can just remove the `if` after the else (line 516). I also think that your comment is inaccurate - for the external pg case, noobaa-core should use the `POSTGRES_CONNECTION_STRING_PATH` (check me). You can add a comment explaining this case inside the else block.

Skiepp: Thanks for the review @dannyzaken  — agreed on collapsing the branches. Here's what I'm going to push:
```
case "POSTGRES_HOST":
    if r.shouldReconcileStandaloneDB() {
        c.Env[j].Value = r.NooBaaPostgresDB.Name + "-0." + r.NooBaaPostgresDB.Spec.ServiceName + "." + r.NooBaaPostgresDB.Namespace + ".svc"
    } else {
        // CNPG and external-pg both ship the host via mounted secret;
        // clear the default so noobaa-core reads POSTGRES_HOST_PATH
        c.Env[j].Value = ""
        c.Env[j].ValueFrom = nil
    }
```
On the `POSTGRES_CONNECTION_STRING_PATH` point — I went to check and I think the comment should reference `POSTGRES_HOST_PATH` instead, here's why:

- `POSTGRES_CONNECTION_STRING_PATH` is not defined in [deploy/internal/statefulset-core.yaml](https://github.com/noobaa/noobaa-operator/blob/master/deploy/internal/statefulset-core.yaml), and `setDesiredCoreEnv` only iterates pre-existing entries in c.Env, so the case for it on [L592](https://github.com/noobaa/noobaa-operator/blob/master/pkg/system/phase2_creating.go#L592) is currently unreachable in the core pod.

- In noobaa-core's postgres_client.js, the `if (process.env.POSTGRES_CONNECTION_STRING_PATH)` guard is therefore falsy in this mode, and the client falls into the per-field branch: `host = process.env.POSTGRES_HOST || readFile(process.env.POSTGRES_HOST_PATH) || '127.0.0.1'`. That's exactly the chain this fix unblocks.

- Empirically: if the connection-string path were active, the bug this PR addresses wouldn't manifest — `POSTGRES_HOST` would simply be ignored. The fact that clearing it is the working fix confirms the per-field branch is the live one for external pg today.

Happy to be wrong here — if you'd prefer external pg to migrate to `POSTGRES_CONNECTION_STRING_PATH` it would mean adding the env to `statefulset-core.yaml` and wiring `<mount>/db_url` in the same case, which feels like a separate change. Let me know how you'd like to proceed.

dannyzaken (approved):

dannyzaken:
@Skiepp LGTM.
Please rebase and squash the commits.
Thanks for the PR!

Skiepp:
> @Skiepp LGTM. Please rebase and squash the commits. Thanks for the PR!

Hi @dannyzaken, thank you, done!

PR #1912, opened 2026-05-19
Fixed Stale External Connection on Endpoint Change

VershaAgrawal:
### Describe the Problem
Updating a NamespaceStore's endpoint leaves the internal NooBaa connection stale — the operator reports "Credentials are not valid", the connection is not cleaned up on deletion (orphaned), and re-creating the store fails with "Connection name already exists".

### Explain the changes
1. Admission webhook (`validate_namespacestore.go`): Reject endpoint mutations on UPDATE  with: *"Changing a NamespaceStore endpoint is unsupported; delete and re-create the NamespaceStore"*. 
2. Owned connection cleanup on delete(`reconciler.go`): Use `list_accounts` to check whether this store owns an external connection (name == CR name). If yes, call `delete_external_connection`. If no (shared connection reuse), skip — shared connections are not deleted.
3. Delete path (`reconciler.go`): `ReadSystemInfo` returns early during deletion after loading the namespace resource — no endpoint/connection setup on delete.
4. Secret rotation(`reconciler.go`): Allow `update_external_connection` when credentials change but the connection already exists.

image

1. https://redhat.atlassian.net/browse/DFBUGS-6006

### Testing Instructions:
1. Create a NamespaceStore pointing to endpoint A, verify it reaches Ready
3. Try to update the endpoint to B  — verify it is rejected with `"Changing a NamespaceStore endpoint is unsupported"`
4. Delete the NamespaceStore — verify the internal connection is removed 
5. Re-create the NamespaceStore pointing to endpoint B — verify it reaches Ready with no `"Connection name already exists"` error
6. Rotate credentials (same endpoint, new secret) — verify UpdateExternalConnectionAPI is called and store stays Ready
7. Delete a store that reuses a shared connection — verify the shared connection is not removed.

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Updates that change storage endpoints for S3-compatible and IBM COS namespace stores are now rejected; to change an endpoint, delete and recreate the namespace store.

* **Behavior**
  * External connection handling improved: identity/credential updates can be applied without recreate when safe; endpoint mismatches still require delete + recreate. Admission validation now rejects invalid endpoint-change updates earlier.

* **Tests**
  * Added tests covering endpoint-change validation and nil-spec scenarios.

### DIFF
--- a/pkg/namespacestore/reconciler.go
+++ b/pkg/namespacestore/reconciler.go
@@ -62,9 +62,9 @@ type Reconciler struct {
 	Secret         *corev1.Secret
 	ServiceAccount *corev1.ServiceAccount
 
-	SystemInfo             *nb.SystemInfo
-	ExternalConnectionInfo *nb.ExternalConnectionInfo
-	NamespaceResourceinfo  *nb.NamespaceResourceInfo
+	SystemInfo              *nb.SystemInfo
+	ExternalConnectionInfo  *nb.ExternalConnectionInfo
+	NamespaceResourceinfo   *nb.NamespaceResourceInfo
 
 	AddExternalConnectionParams    *nb.AddExternalConnectionParams
 	CreateNamespaceResourceParams  *nb.CreateNamespaceResourceParams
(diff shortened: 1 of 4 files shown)


pkg/namespacestore/reconciler.go
```diff
@@ -741,7 +799,7 @@ func (r *Reconciler) fixAlternateKeysNames() {
 // ReconcileExternalConnection handles the external connection using noobaa api
 func (r *Reconciler) ReconcileExternalConnection() error {
 
-	if r.ExternalConnectionInfo != nil {
+	if r.ExternalConnectionInfo != nil && r.UpdateExternalConnectionParams == nil {
```
naveenpaul1, line 784: move this check to next if condition

pkg/namespacestore/reconciler.go
```diff
@@ -470,6 +496,34 @@ func (r *Reconciler) ReadSystemInfo() error {
 		return err
 	}
 
+	r.ownedExternalConnection = nil
+
+	accountsList, err := r.NBClient.ListAccountsAPI(nb.ListAccountsParams{})
```
naveenpaul1: remove this validation from `ReadSystemInfo()`

pkg/namespacestore/reconciler.go
```diff
@@ -470,6 +496,34 @@ func (r *Reconciler) ReadSystemInfo() error {
 		return err
 	}
 
+	r.ownedExternalConnection = nil
```
naveenpaul1: var ownedExternalConnection = nil

pkg/namespacestore/reconciler.go
VershaAgrawal: The first condition can’t be if r.ExternalConnectionInfo != nil { return nil } — that matches initial code and skips update_external_connection when credentials change but the store reuses an existing connection. We need && r.UpdateExternalConnectionParams == nil so secret/credential updates still run. The second condition can stay as AddExternalConnectionParams == nil since add params are always populated for the check/update paths.

naveenpaul1 (approved):
LGTM

PR #1918, opened 2026-05-26
Deep Archive NamespaceStore

romayalon:
### Explain the changes
1. Added archive boolean to S3 compatible NamespaceStore CRD
2. Added namespacestore reconciler, admission and validations related changes.
3. Added CLI changes needed for supporting the new namespacestore archive field.
4. Added deep archive section to the namespacestore documentation.

1. Gap - tests will be added in the following PR 
2. validate we don't set archive: true on non IBM deep archive s3 compatible namespacestores
3. re-consider adding the immutability validation for the `archive:boolean`

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Support for marking NamespaceStores as cold-storage archives (e.g., IBM Deep Archive) via a new archive option for S3-compatible stores.
  * Added a CLI flag to create archive-enabled NamespaceStores.

* **Bug Fixes / Validation**
  * Validation now enforces that archive can only be enabled for S3-compatible stores and that the archive setting cannot be changed after creation.

* **Documentation**
  * Added guidance and examples for S3-compatible deep-archive configuration.

[Review Change Stack](https://app.coderabbit.ai/change-stack/noobaa/noobaa-operator/pull/1918?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/deploy/crds/noobaa.io_namespacestores.yaml
+++ b/deploy/crds/noobaa.io_namespacestores.yaml
@@ -54,6 +54,11 @@ spec:
               accessMode:
                 description: AccessMode is an enum of supported access modes
                 type: string
+              archive:
+                description: |-
+                  Archive marks the namespace store for cold-storage archive use (e.g. IBM Deep Archive).
+                  Only valid with type s3-compatible.
+                type: boolean
               awsS3:
                 description: AWSS3Spec specifies a namespace store of type aws-s3
                 properties:
--- a/doc/namespace-store-crd.md
+++ b/doc/namespace-store-crd.md
@@ -100,6 +100,37 @@ spec:
   type: s3-compatible
 ```
 
+### S3 Compatible - Deep Archive
+Uses an S3-compatible API to write objects directly to a tape-based cold storage endpoint (e.g. IBM Deep Archive).
+This store type is intended for S3-compatible long-term archive where data is written once and read infrequently.
+
+`archive: true` - A flag that indicates the endpoint is a tape-based or cold-storage target (e.g. IBM Deep Archive).
+Archive stores can only be referenced via `archivePolicy` in a BucketClass; they cannot be used inside a `namespacePolicy`.
+
+```shell
+noobaa namespacestore create s3-compatible <NAMESPACESTORE NAME> --archive \
+  --endpoint <> --target-bucket <> --access-key <> --secret-key <> 
+```
+
+```yaml
+apiVersion: noobaa.io/v1alpha1
+kind: NamespaceStore
+metadata:
+  finalizers:
+  - noobaa.io/finalizer
+  name: <>
+  namespace: <>
+spec:
+  archive: true
+  s3Compatible:
+    endpoint: <>
+    secret:
+      name: <>
+      namespace: <>
+    targetBucket: <>
+  type: s3-compatible
+```
+
 ## IBM COS
 Uses the IBM COS API for IO operations on plain data in IBM COS buckets
 ```shell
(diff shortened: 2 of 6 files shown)


deploy/crds/noobaa.io_namespacestores.yaml
```diff
@@ -247,6 +252,10 @@ spec:
             required:
             - type
             type: object
+            x-kubernetes-validations:
```
jackyalbo: @romayalon has it's the first time we are using this. Is this fully supported in OCP in the relevant version? Was it checked? Do we still need both the Admission webhook and this API constraint? Isn't it redundant?

doc/namespace-store-crd.md
```diff
@@ -100,6 +100,37 @@ spec:
...
 
+### S3 Compatible - Deep Archive
+Uses an S3-compatible API to write objects directly to a tape-based cold storage endpoint (e.g. IBM Deep Archive).
+This store type is intended for S3-compatible long-term archive where data is written once and read infrequently.
+
+`archive: true` - A flag that indicates the endpoint is a tape-based or cold-storage target (e.g. IBM Deep Archive).
+Archive stores can only be referenced via `archivePolicy` in a BucketClass; they cannot be used inside a `namespacePolicy`.
+
+```shell
+noobaa namespacestore create s3-compatible <NAMESPACESTORE NAME> \
```
jackyalbo: I would maybe move `--archive` to the beginning of the command variables list... it looks a bit lost at the end of the command.

pkg/validations/namespacestore_validations.go
```diff
@@ -411,6 +414,37 @@ func ValidateTargetNSBucketChange(ns nbv1.NamespaceStore, oldNs nbv1.NamespaceSt
...
+		}
+	}
+	return nil
+}
+
+// ValidateNSArchiveImmutable rejects any update that changes the archive flag.
+// Archive is an immutable property set at creation time; toggling it in either
+// direction would silently change how the store is treated by bucket classes and
+// the core, which is not safe.
+func ValidateNSArchiveImmutable(ns nbv1.NamespaceStore, oldNs nbv1.NamespaceStore) error {
```
jackyalbo: double check here if not redundant - If it does, remove only if we don't have issues with supporting restrictions.

deploy/crds/noobaa.io_namespacestores.yaml
romayalon: I added it in case the admission server turned off, do you think there is no such case where the admission server is turned off? 
tested it ofc,
for example - 
I created a regular s3 compatible namespacestore and then tried to add `archive:true` to its yaml, 
when trying to save the yaml I got the following error - 
```
....
# namespacestores.noobaa.io "nss2" was not valid:
# * spec: Invalid value: "object": spec.archive is immutable after creation
#
```

pkg/validations/namespacestore_validations.go
romayalon: As mentioned before, maybe not redundant for IBM deep archive namespacestore, but this immutability might cause issues in the future if we would like to have the same s3 target bucket for both standard and archive storage classes, let me remove this restriction for now

deploy/crds/noobaa.io_namespacestores.yaml
romayalon: nvm, I removed it as a start, we can re consider it later

jackyalbo (approved):

PR #1930, opened 2026-06-04
Harden CNPG PostgreSQL config for replica HA stability

dannyzaken:
### Describe the Problem

CNPG's default PostgreSQL configuration is not suitable for NooBaa's workload. Specifically:
- `wal_keep_size` defaults to 512MB (CNPG default), which is insufficient for replica bootstrap windows at NooBaa's WAL generation rate (~150-200 MB/min), causing permanent stuck replicas.
- `wal_level` defaults to `logical`, adding ~10-30% WAL volume overhead with no benefit (NooBaa does not use logical replication).
- `wal_sender_timeout` and `wal_receiver_timeout` default to 5s (CNPG override of PG's 60s default), causing premature WAL streaming disconnects on transient hiccups.
- `failoverDelay` defaults to 0, triggering unnecessary failovers on brief blips.
- `setPostgresConfig` had two divergent code paths (calculated vs. static fallback) that were hard to maintain.

### Explain the Changes
1. Simplify `setPostgresConfig` — remove the static fallback map; always call `calculatePGConfig` using defaults of 16GB memory and 4 CPUs when `DBResources` are not specified (for calculation only, not set on the CR).
2. Move static/constant PG parameters to the top of `setPostgresConfig` as a single map literal; `calculatePGConfig` now only handles dynamically-computed values.
3. Set `wal_level: replica` — eliminates ~10-30% WAL overhead from unused logical decoding.
4. Set `wal_sender_timeout: 60s` and `wal_receiver_timeout: 60s` — match PostgreSQL upstream defaults; prevents premature streaming disconnects.
5. Add `calculateWalKeepSize` — computes `wal_keep_size` as 12% of the data PVC size, capped at 16GB (min 1GB), protecting replica bootstrap from WAL recycling.
6. Set `failoverDelay: 30` on the CNPG Cluster spec — gives transient issues 30s to resolve before triggering a failover.
7. Always set `huge_pages: off` — removes conditional logic that set `"try"` for large shared_buffers.
8. Move `calculateMaxConnections` call inside `calculatePGConfig` — function now takes `numEndpoints` directly.
9. Track `FailoverDelay` in `wasClusterSpecChanged` so changes propagate to existing clusters.

1. https://redhat.atlassian.net/browse/DFBUGS-6943

### Testing Instructions:
1. Run unit tests: `go test ./pkg/system/ -run "TestCalculatePGConfig|TestCalculateWalKeepSize|TestCalculateMaxConnections|TestFormatBytesKB" -v`
2. Deploy operator on a cluster with a CNPG-managed NooBaa DB and verify the Cluster CR parameters are updated on reconcile.
3. Verify `DBConf` user overrides still take precedence (set `dbConf.wal_level: logical` in NooBaa CR and confirm it wins).

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Improved PostgreSQL auto-tuning: computes settings from container limits (with request fallbacks), uses endpoint count for max connections, applies static overrides (e.g., huge_pages off, checkpoint target), and enforces defaults and DB-config overrides; clusters now honor failover-delay changes.

* **Tests**
  * Expanded unit tests for endpoint-based tuning and WAL keep-size parsing, capping, and fallback behavior.

### DIFF
--- a/pkg/system/db_reconciler.go
+++ b/pkg/system/db_reconciler.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"reflect"
 	"slices"
-	"strconv"
 	"strings"
 	"time"
 
@@ -29,6 +28,10 @@ const (
 
 	noobaaDBUser = "noobaa"
 	noobaaDBName = "nbcore"
+
+	defaultCalcMemoryKB     = int64(16 * 1024 * 1024) // 16GB in KB — used when DBResources not specified
+	defaultCalcCPU          = int64(4)                // used when DBResources not specified
+	defaultFailoverDelaySec = int32(30)
 )
 
 // ReconcileCNPGCluster reconciles the CNPG cluster
@@ -300,6 +303,8 @@ func (r *Reconciler) reconcileClusterSpec(dbSpec *nbv1.NooBaaDBSpec) error {
 		Enabled: true,
 	}
 
+	r.CNPGCluster.Spec.FailoverDelay = defaultFailoverDelaySec
+
 	r.setPostgresConfig()
 
 	// Configure backup settings if specified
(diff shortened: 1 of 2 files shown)


dannyzaken:
The fix is based on new changes in 5.22 that cannot be backported to 5.21 and below. If we want to backport we need a slightly modified fix

pkg/system/db_reconciler.go
```diff
@@ -625,26 +627,21 @@ func (r *Reconciler) setPostgresConfig() {
...
 // Fixed assumptions: DB Version: 16, OS: Linux, DB Type: OLTP, Storage: SAN.
 // totalMemoryKB is the total DB memory in kilobytes, cpuNum is the number of CPU
-// cores (0 means unknown), and maxConnections is the desired max_connections value.
-func calculatePGConfig(totalMemoryKB int64, cpuNum int64, maxConnections int) map[string]string {
+// cores (0 means unknown), numEndpoints is the number of endpoint pods used to
+// derive max_connections, and dataVolumeSize is the PVC size string used to derive wal_keep_size.
+func calculatePGConfig(totalMemoryKB int64, cpuNum int64, numEndpoints int, dataVolumeSize string) map[string]string {
 	params := map[string]string{}
 
+	maxConnections := calculateMaxConnections(numEndpoints)
```
tangledbytes, line 639: Do we not want allow user specified max_connections anymore? What about the previously backward support?

karthikps97:
Postgres also has a `checkpoint_timeout` which controls how often the WAL is flushed to disk. We support cnpg and standalone PG deployments.

For cnpg, the default is 15 mins and standalone postgres uses 5 mins as default.
Since we are setting a relatively high `min_wal_size`, should we also tweak the `checkpoint_timeout` to a relatively higher value ~30 mins?

Also making it uniform for both cnpg and standalone pg would be better and would result in a uniform behavior.

pkg/system/db_reconciler.go
dannyzaken: Users should still be able to set a different `max_connections` value. This happens on line 614 (verify me)

dannyzaken:
> Postgres also has a `checkpoint_timeout` which controls how often the WAL is flushed to disk. We support cnpg and standalone PG deployments.
> 
> For cnpg, the default is 15 mins and standalone postgres uses 5 mins as default. Since we are setting a relatively high `min_wal_size`, should we also tweak the `checkpoint_timeout` to a relatively higher value ~30 mins?
> 
> Also making it uniform for both cnpg and standalone pg would be better and would result in a uniform behavior.

@karthikps97 I think that CNPG does not change the default value of `checkpoint_timeout` and it is still 5 min. Where did you see it's 15 min?
Can you explain the relation between `min_wal_size` to `checkpoint_timout`? did you mean `max_wal_size` maybe? In any case, this PR did not change `min_wal_size` or `max_wal_size` from their previous values, which are the same as those we also used for standalone. Increasing checkpoint_timeout will results with fewer checkpoints, but its a tradeoff with longer recovery times. 
I can see a case for reducing max_wal_size to better align with checkpoint_timeout, but this is out of scope for this PR. We can have a separate discussion if this change is needed.

karthikps97:
> > Postgres also has a `checkpoint_timeout` which controls how often the WAL is flushed to disk. We support cnpg and standalone PG deployments.
> > For cnpg, the default is 15 mins and standalone postgres uses 5 mins as default. Since we are setting a relatively high `min_wal_size`, should we also tweak the `checkpoint_timeout` to a relatively higher value ~30 mins?
> > Also making it uniform for both cnpg and standalone pg would be better and would result in a uniform behavior.
> 
> @karthikps97 I think that CNPG does not change the default value of `checkpoint_timeout` and it is still 5 min. Where did you see it's 15 min? Can you explain the relation between `min_wal_size` to `checkpoint_timout`? did you mean `max_wal_size` maybe? In any case, this PR did not change `min_wal_size` or `max_wal_size` from their previous values, which are the same as those we also used for standalone. Increasing checkpoint_timeout will results with fewer checkpoints, but its a tradeoff with longer recovery times. I can see a case for reducing max_wal_size to better align with checkpoint_timeout, but this is out of scope for this PR. We can have a separate discussion if this change is needed.

@dannyzaken, Sorry, I meant max_wal_size which since we have a higer value, we could also increase the `checkpoint_timeout`. The resource I looked at was [Link](https://docs.azure.cn/en-us/aks/deploy-postgresql-ha?tabs=azuredisk), and you're right it's mentioned as the "recommended" value.

alphaprinz (approved):
I didn't find the way to set max connections mentioned in https://github.com/noobaa/noobaa-operator/pull/1930/changes#r3363104688

Otherwise LGTM :)

PR #1923, opened 2026-05-31
GCP WIF (STS) | Adding  a New Type of Backingstore

shirady:
### Describe the Problem
As part of [RHSTOR-8661](https://redhat.atlassian.net/browse/RHSTOR-8661), this PR sets the new endpoint type `GOOGLE_STS` (we had `GOOGLE`) and starts implementing code changes in parts of the flows related to the GCP WIF (STS) backing store. This PR relies on the changes in PR noobaa/noobaa-core#9644.

### Explain the changes
1. In Backingstore, reconcile the use of the new endpoint type `EndpointTypeGoogleSTS`.
Note: In the `IsStringGraphicOrSpacesCharsOnly` I excluded GCP and GCP STS as it seems to be only for AWS-style properties.
2. Added util functions for the new type, and also added AI-generated tests for the new functions.

### Issues: 
List of GAPS:
1. In this PR we do not handle the CLI changes and default backingstore changes.
2. GCP WIF (STS) namespace store is not handled as part of this PR.
3. Currently, in the secret we still have the field of `GoogleServiceAccountPrivateKeyJson` in the data; also, for the case of GCP WIF (STS), we might change it later for readability.

#### Unit Tests
1. Please run: `go test -mod=vendor -v ./pkg/util -run 'TestParseGoogleCredentials|TestGoogleIdentityFromJSON|TestGoogleIdentityFromImpersonationURL'`

#### Manual Test:
More details can be found in [comment](https://github.com/noobaa/noobaa-operator/pull/1923#issuecomment-4590173559)
1. Deploy NooBaa on Minikube after adding GCP WIF configurations - see guide "Create GCP WIF (STS) Setup On Minikube With NooBaa" ([link](https://github.com/noobaa/noobaa-operator/blob/master/doc/dev_guide/create_gcp_wif_setup_on_minikube_with_noobaa.md)).
2. Create a secret with the details of the credential.json for GCP WIF (STS).
3. Create a backing store to use the secret of the kind GCP WIF (STS), expect it to be in `Ready`.
4. Write to the bucket and read from it - for this need to create a bucketclass on top of the backingstore and an OBC and then use the credentials to put object and get object from the bucket.

- [ ] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **New Features**
  * Google Cloud Storage backing stores now detect and support Google STS (Workload Identity Federation) alongside service-account auth, deriving and storing a consistent masked identity and using the appropriate endpoint type.

* **Tests**
[...]

PR #1931, opened 2026-06-04
GCP WIF (STS) | Backingstore CLI

shirady:
### Describe the Problem
As part of [RHSTOR-8661](https://redhat.atlassian.net/browse/RHSTOR-8661), this PR sets the new endpoint type `GOOGLE_STS` (we had `GOOGLE`) and starts implementing code changes in parts of the flows related to the GCP WIF (STS) backing store.

### Explain the changes
1. Added a new subcommand `noobaa backingstore google-cloud-storage-sts <backing-store-name>` that either get the required parameters for creating the JSON in the secret or gets the secret with the JSON inside.
2. Added in utils the token path const and set it where it was already used. Also, added helper function for creating the JSON that would be set as a value in the GCP WIF (STS) value.
3. Added AI-generated tests for the helper functions.
4. Added documentation about the GCP WIF (STS) type.
5. Added tests on the help to test the existing flags for GCP and GCP WIF (STS) in the `noobaa backingstore` command.

### Issues:
List of GAPs:
1. GCP WIF (STS) namespace store is not handled as part of this PR.
2. Currently, in the secret we still have the field of `GoogleServiceAccountPrivateKeyJson` in the data; also,  for the case of GCP WIF (STS), we might change it later for readability.
3. I didn't add validation per required parameter in the CLI - would probably add in one of the following PRs (link to the issue #1933).

#### Unit Tests:
Please run:
1. CLI help test: 
- First, build the image and the CLI (can be done by `make all DOCKER_DEFAULT_PLATFORM=linux/arm64 GOARCH=arm64`).
- Export the operator image tag that was build: `export OPERATOR_IMAGE=noobaa/noobaa-operator:5.23.0` 
- Run the test: `go test -mod=vendor -v ./pkg/cli -ginkgo.focus "google-cloud-storage"`
2. Helper functions unit tests: `go test -mod=vendor -count=1 -v ./pkg/util -run 'TestGoogleWIFAudience|TestBuildGoogleWIFCredentialsJSON'`

#### Manual Tests:
1. Deploy NooBaa on Minikube after adding GCP WIF configurations - see guide "Create GCP WIF (STS) Setup On Minikube With NooBaa" ([link](https://github.com/noobaa/noobaa-operator/blob/master/doc/dev_guide/create_gcp_wif_setup_on_minikube_with_noobaa.md)).
2. Create a secret with the details of the credential.json for GCP WIF (STS).
3. Create a backing using the CLI (expect the backingstore to be `Ready`):
- (should ask for required parameters): `nooba backingstore create google-cloud-storage-sts <backingstore-name> -n <namespace>`
- (should not ask for required parameters): `nooba backingstore create google-cloud-storage-sts <backingstore-name> --secret-name  <secret-name> -n <namespace>` (the instructions for creating the secret [here](https://github.com/noobaa/noobaa-operator/pull/1923#issuecomment-4590173559)).

- [X] Doc added/updated
- [X] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added Google Cloud Storage Workload Identity Federation (WIF/STS) backing store support and new CLI options to configure WIF parameters; backing store creation accepts STS credentials.
  * Reconciler now uses the projected web identity token path by default for STS flows.

* **Documentation**
  * Clarified GCS credential types and WIF/STS usage with CLI and BackingStore examples.

* **Tests**
  * Added tests for WIF audience/credentials generation and updated CLI help tests.

### DIFF
--- a/doc/backing-store-crd.md
+++ b/doc/backing-store-crd.md
@@ -71,7 +71,7 @@ However, the difference between the two backingstore types lies in the authentic
 
 This type of backingstore is useful in cases where the user wishes to limit access to their AWS cloud for a specific amount of time, and for easier management of the cloud's security.
 
-Prior to using this backingstore, an OpenIDConnect provider needs to be set up, which is outside the scope of these docs.
+Before using this backingstore, an OpenIDConnect provider needs to be set up, which is outside the scope of these docs.
 
 ```shell
 noobaa backingstore create aws-sts-s3 <BACKINGSTORE NAME> --target-bucket <> --aws-sts-arn <>
@@ -143,7 +143,7 @@ spec:
   type: ibm-cos
 ```
 
-## Google Cloud Storage
+## Google Cloud Storage (GCP)
 Uses the Google Cloud Storage API for storing encrypted chunks of data in Google Cloud buckets
 ```shell
 noobaa backingstore create google-cloud-storage <BACKINGSTORE NAME> --private-key-json-file <PATH TO credentials.json> --target-bucket <>
@@ -172,6 +172,36 @@ Permissions: the service account should have the `storage.buckets.list` permissi
   - `Browser` role: provides the `storage.buckets.list` permission.
   - `Storage Admin` role: provides full control over buckets and objects.
 - The `Owner` role (for development; not recommended for production environments).
+
+## Google Cloud Storage Workload Identity Federation (GCP WIF, STS)
+Similarly to `Google Cloud Storage` this backing store uses the Google Cloud Storage API for storing encrypted chunks of data in GCP buckets.
+However, the difference between the two backingstore types lies in the authentication method:
+- `google-cloud-storage` uses long-lived `service_account` JSON keys.
+- `google-cloud-storage-sts` uses a short-lived token via GCP Workload Identity Federation: the JSON type is `external_account`, which is built from parameters (`PROJECT_NUMBER`, `POOL_ID`, `PROVIDER_ID`, `SERVICE_ACCOUNT_EMAIL`).
+
+This type of backingstore is useful in cases where the user wishes to limit access to their GCP cloud for a specific amount of time, and for easier management of the cloud's security.
+
+Before using this backingstore, there are configurations that need to be done on the GCP account and on the cluster (for example, an OpenIDConnect provider needs to be set up) which is outside the scope of these docs.
+
+```shell
+noobaa backingstore create google-cloud-storage-sts <BACKINGSTORE NAME> --target-bucket <> --project-number <> --pool-id <> --provider-id <> --service-account-email <>
+```
+```yaml
+apiVersion: noobaa.io/v1alpha1
+kind: BackingStore
+metadata:
+  finalizers:
+  - noobaa.io/finalizer
+  name: <>
+  namespace: <>
+spec:
+  googleCloudStorage:
+    secret:
+      name: <>
+      namespace: <>
+    targetBucket: <>
+  type: google-cloud-storage
+```
 
 ## Azure Blob Storage
 Uses the Azure Blob API for storing encrypted chunks of data in an Azure container
--- a/pkg/backingstore/backingstore.go
+++ b/pkg/backingstore/backingstore.go
@@ -64,6 +64,7 @@ func CmdCreate() *cobra.Command {
 		CmdCreateIBMCos(),
 		CmdCreateAzureBlob(),
 		CmdCreateGoogleCloudStorage(),
+		CmdCreateGoogleCloudStorageSTS(),
 		CmdCreatePVPool(),
 		CmdCreateAzureSTS(),
 	)
@@ -246,7 +247,7 @@ func CmdCreateAzureBlob() *cobra.Command {
 func CmdCreateGoogleCloudStorage() *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "google-cloud-storage <backing-store-name>",
-		Short: "Create google-cloud-storage backing store",
+		Short: "Create google-cloud-storage backing store (using long-lived credentials)",
 		Run:   RunCreateGoogleCloudStorage,
 	}
 	cmd.Flags().String(
@@ -260,6 +261,40 @@ func CmdCreateGoogleCloudStorage() *cobra.Command {
 	cmd.Flags().String(
 		"secret-name", "",
 		`The name of a secret for authentication - should have GoogleServiceAccountPrivateKeyJson property`,
+	)
+	return cmd
+}
+
+// CmdCreateGoogleCloudStorageSTS returns a CLI command
+func CmdCreateGoogleCloudStorageSTS() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "google-cloud-storage-sts <backing-store-name>",
+		Short: "Create google-cloud-storage backing store using GCP WIF (STS, short-lived credentials)",
+		Run:   RunCreateGoogleCloudStorageSTS,
+	}
+	cmd.Flags().String(
+		"target-bucket", "",
+		"The target bucket name on Google cloud storage",
+	)
+	cmd.Flags().String(
+		"service-account-email", "",
+		"The GCP service account email to impersonate",
+	)
+	cmd.Flags().String(
+		"project-number", "",
+		"The GCP project number (numeric string, e.g. 123456789; not the project ID)",
+	)
+	cmd.Flags().String(
+		"pool-id", "",
+		"The GCP workload identity pool ID",
+	)
+	cmd.Flags().String(
+		"provider-id", "",
+		"The GCP workload identity provider ID",
+	)
+	cmd.Flags().String(
+		"secret-name", "",
+		`The name of a secret for authentication - should have GoogleServiceAccountPrivateKeyJson property (with type of external_account)`,
 	)
 	return cmd
 }
@@ -483,7 +518,7 @@ func RunCreate(cmd *cobra.Command, args []string) {
 	if len(args) != 1 || args[0] == "" {
 		log.Fatalf(`❌ Missing expected arguments: <backing-store-type> %s`, cmd.UsageString())
 	}
-	if args[0] != "aws-s3" && args[0] != "aws-sts-s3" && args[0] != "google-cloud-storage" &&
+	if args[0] != "aws-s3" && args[0] != "aws-sts-s3" && args[0] != "google-cloud-storage" && args[0] != "google-cloud-storage-sts" &&
 		args[0] != "azure-blob" && args[0] != "ibm-cos" && args[0] != "pv-pool" && args[0] != "s3-compatible" && args[0] != "azure-sts-blob" {
 		log.Fatalf(`❌ Unsupported <backing-store-type> -> %s %s`, args[0], cmd.UsageString())
 	}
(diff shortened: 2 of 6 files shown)


doc/backing-store-crd.md
```diff
@@ -173,6 +173,36 @@ Permissions: the service account should have the `storage.buckets.list` permissi
...
+  namespace: <>
+spec:
+  googleCloudStorage:
+    secret:
+      name: <>
+      namespace: <>
+    targetBucket: <>
+  type: google-cloud-storage
+```
+
```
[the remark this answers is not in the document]
shirady: For the first suggestion - I think it might be confusing, as the type they write is still `google-cloud-storage` the option `google-cloud-storage-sts` is in the CLI for authentication. It should be like in AWS and Azure, and both do not appear in the list.

About the comma, thanks, I will fix it.

pkg/backingstore/backingstore.go
```diff
@@ -741,6 +776,41 @@ func RunCreateGoogleCloudStorage(cmd *cobra.Command, args []string) {
...
+				poolID, providerID, serviceAccountEmail)
+			if err != nil {
+				log.Fatalf("Failed to build GCP WIF credentials: %v", err)
+			}
+			secret.StringData["GoogleServiceAccountPrivateKeyJson"] = credentialsJSON
+		} else {
+			util.VerifyCredsInSecret(secretName, options.Namespace, mandatoryProperties)
+			secret.Name = secretName
+			secret.Namespace = options.Namespace
+		}
```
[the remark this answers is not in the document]
shirady: It would be validated as part of backingstore reconcile, right? So why changing the style of the function? (I wanted it to be like it is in `RunCreateGoogleCloudStorage`.

pkg/util/util.go
```diff
@@ -1253,6 +1275,51 @@ func googleIdentityFromImpersonationURL(impersonationURL string) (string, error)
...
+	poolID = strings.TrimSpace(poolID)
+	providerID = strings.TrimSpace(providerID)
+	serviceAccountEmail = strings.TrimSpace(serviceAccountEmail)
+
+	if projectNumber == "" || poolID == "" || providerID == "" || serviceAccountEmail == "" {
+		return "", fmt.Errorf("project number, pool id, provider id, and service account email are required")
+	}
+	if !strings.Contains(serviceAccountEmail, "@") {
+		return "", fmt.Errorf("invalid service_account_impersonation_url: malformed service account email %q", serviceAccountEmail)
+	}
```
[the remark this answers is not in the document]
shirady: I might add validations to all the parameters later; it is a gap.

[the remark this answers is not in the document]
shirady: Sure, thanks.

pkg/backingstore/backingstore.go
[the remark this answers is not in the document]
shirady: You can add it in the issue you already created and I will consider it - 
https://github.com/noobaa/noobaa-operator/issues/1933

pkg/backingstore/backingstore.go
```diff
@@ -741,6 +776,41 @@ func RunCreateGoogleCloudStorage(cmd *cobra.Command, args []string) {
...
 
+// RunCreateGoogleCloudStorageSTS runs a CLI command
+func RunCreateGoogleCloudStorageSTS(cmd *cobra.Command, args []string) {
+	log := util.Logger()
+	createCommon(cmd, args, nbv1.StoreTypeGoogleCloudStorage, func(backStore *nbv1.BackingStore, secret *corev1.Secret) {
+		targetBucket := util.GetFlagStringOrPrompt(cmd, "target-bucket")
+		secretName, _ := cmd.Flags().GetString("secret-name")
+		mandatoryProperties := []string{"GoogleServiceAccountPrivateKeyJson"}
+
+		if secretName == "" {
```
naveenpaul1, line 787: In backingstore reconsile loop how are you identify its a nomal backingstore and STS backingstore?

pkg/util/util.go
```diff
@@ -1253,6 +1275,51 @@ func googleIdentityFromImpersonationURL(impersonationURL string) (string, error)
...
+
+	googleExternalAccountCredentialType := "external_account"
```
naveenpaul1, line 1294: when googleExternalAccountCredentialType is empty that is normal backingstore, is that right?

pkg/util/util.go
```diff
@@ -118,6 +121,25 @@ type googleCredentialsJSON struct {
...
+	CredentialSource               googleWIFCredentialSource `json:"credential_source"`
+}
+
+type googleWIFCredentialSource struct {
+	File   string                    `json:"file"`
+	Format googleWIFCredentialFormat `json:"format"`
+}
+
+type googleWIFCredentialFormat struct {
+	Type string `json:"type"`
```
naveenpaul1: Do we really need a struct only for one property?

pkg/backingstore/backingstore.go
```diff
@@ -741,6 +776,41 @@ func RunCreateGoogleCloudStorage(cmd *cobra.Command, args []string) {
...
+			providerID := util.GetFlagStringOrPrompt(cmd, "provider-id")
+			serviceAccountEmail := util.GetFlagStringOrPrompt(cmd, "service-account-email")
+			credentialsJSON, err := util.BuildGoogleWIFCredentialsJSON(projectNumber,
+				poolID, providerID, serviceAccountEmail)
+			if err != nil {
+				log.Fatalf("Failed to build GCP WIF credentials: %v", err)
+			}
+			secret.StringData["GoogleServiceAccountPrivateKeyJson"] = credentialsJSON
+		} else {
+			util.VerifyCredsInSecret(secretName, options.Namespace, mandatoryProperties)
```
naveenpaul1, line 799: Are we checking the secret type `googleExternalAccountCredentialType`, if type missing or wrong, Backingstore will consider as normal and result in error?

pkg/util/util.go
```diff
@@ -1253,6 +1275,51 @@ func googleIdentityFromImpersonationURL(impersonationURL string) (string, error)
...
+	credentials := googleCredentialsJSONFullSTS{
+		Type:                           googleExternalAccountCredentialType,
+		Audience:                       audience,
+		SubjectTokenType:               "urn:ietf:params:oauth:token-type:jwt",
+		TokenURL:                       "https://sts.googleapis.com/v1/token",
+		ServiceAccountImpersonationURL: GoogleImpersonationURLPrefix + serviceAccountEmail + GoogleImpersonationURLSuffix,
+		CredentialSource: googleWIFCredentialSource{
+			File: WebIdentityTokenPath,
+			Format: googleWIFCredentialFormat{
+				Type: "text",
```
naveenpaul1: If we are hardcoding this value why we need it? Are we using it anywhere else?

pkg/backingstore/backingstore.go
shirady: The changes in the backingstore reconcile we done in the previous PR - 
https://github.com/noobaa/noobaa-operator/blob/6a1a28605242a57377499c978f503b5527611059/pkg/backingstore/reconciler.go#L833-L850
So we set a different `EndpointType` and send different  `Identity`, in both cases we send the JSON file as a secret, but the content (the fields inside) is different.

pkg/backingstore/backingstore.go
shirady: Yes, it would result in an error - 
`util.VerifyCredsInSecret` verifies that the secret contains the `mandatoryProperties` which are `GoogleServiceAccountPrivateKeyJson` currently.

In the flow, we `ParseGoogleCredentials` there; if the type is missing, it would be an empty string, and then in `googleIdentityFromCredentials` the empty string would result in an error (also for any other type than `external_account` and `service_account`).

pkg/util/util.go
shirady: The structure in the JSON is ([link](https://docs.redhat.com/en/documentation/openshift_container_platform/4.21/html/authentication_and_authorization/managing-cloud-provider-credentials#cco-short-term-creds-format-gcp_cco-short-term-creds) to the docs):
```json
   "credential_source": {
      "file": "<path_to_token>", 
      "format": {
         "type": "text"
      }
   }
```
In JSON the format is an object, so the equivalent in Go is to have it as a struct.
I will change it to use map, thanks.

pkg/util/util.go
shirady: No, it should be either `external_account` (STS case) or `service_account` (as it was before).

pkg/util/util.go
shirady: Related to the previous [comment](https://github.com/noobaa/noobaa-operator/pull/1931/changes#r3361640484) - 
I will change it to use a map, thanks.

doc/backing-store-crd.md
```diff
@@ -173,6 +173,36 @@ Permissions: the service account should have the `storage.buckets.list` permissi
...
 
+## Google Cloud Storage Workload Identity Federation (GCP WIF, STS)
+Similarly to `Google Cloud Storage` this backing store uses the Google Cloud Storage API for storing encrypted chunks of data in GCP buckets.
+However, the difference between the two backingstore types lies in the authentication method:
+- `google-cloud-storage` uses long-lived `service_account` JSON keys.
+- `google-cloud-storage-sts` uses a short-lived token via GCP Workload Identity Federation: the JSON type is `external_account`, which is built from parameters (`PROJECT_NUMBER`, `POOL_ID`, `PROVIDER_ID`, `SERVICE_ACCOUNT_EMAIL`).
+
+This type of backingstore is useful in cases where the user wishes to limit access to their GCP cloud for a specific amount of time, and for easier management of the cloud's security.
+
+Prior to using this backingstore, there are configurations that needs to be done on the GCP account and on the cluster (for example, an OpenIDConnect provider needs to be set up) which is outside the scope of these docs.
```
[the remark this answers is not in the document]
shirady: I will change it, but the style was copied from the AWS STS docs in this file.

naveenpaul1 (approved):

pkg/backingstore/backingstore.go
kajalpareek-lab, line 802: : When a user passes --secret-name, the Azure STS path does:
util.VerifyCredsInSecret(secretName, options.Namespace, mandatoryProperties)
secret.Name = secretName
secret.Namespace = options.Namespace
util.KubeCheck(secret)  // <-- loads the actual secret data
But the GCP STS path does:
util.VerifyCredsInSecret(secretName, options.Namespace, mandatoryProperties)
secret.Name = secretName
secret.Namespace = options.Namespace
// Missing: util.KubeCheck(secret)
Without KubeCheck, the secret object passed to createCommon may not have its StringData populated from the cluster. Check whether createCommon handles this downstream — if not, the backing store gets created referencing a secret whose contents haven't been verified to actually load.

shirady: In other backingstore functions in this file (AWS S3, S3-compatible, IBM COS, Azure Blob, and GCP), we do not have the call to `util.KubeCheck(secret)`.
We have the function `VerifyCredsInSecret` and part of its implementation is to check `KubeCheck(secret)`:
https://github.com/noobaa/noobaa-operator/blob/6a1a28605242a57377499c978f503b5527611059/pkg/util/util.go#L1564-L1578

In `createCommon` we also check that `!util.KubeCheck(secret)`:
https://github.com/noobaa/noobaa-operator/blob/6a1a28605242a57377499c978f503b5527611059/pkg/backingstore/backingstore.go#L414-L419

So I think it is covered.

The case of Azure STS (from what I found with AI):

> Azure STS is the exception — it needs KubeCheck inside populate because it reads azure_client_id from the secret to set ClientId on the spec.

https://github.com/noobaa/noobaa-operator/blob/6a1a28605242a57377499c978f503b5527611059/pkg/backingstore/backingstore.go#L601-L612

kajalpareek-lab (approved):
looks good to me.

PR #1936, opened 2026-06-09
Fixed BucketClass quota ignore during OBC update

VershaAgrawal:
### Describe the Problem
When both a BucketClass and an OBC define quota, the intended behavior is to apply the minimum (most restrictive) value per dimension. This works correctly on initial OBC creation, but breaks on OBC update - the BucketClass quota is silently ignored, and only OBC quota values are applied.
### Explain the changes
1. Added getBucketClassByName to look up a BucketClass by name and namespace.
2. Kept getBucketClass as a thin wrapper for the create/provision path — no behavior change there.
3. Fixed the update path in NewBucketRequest to call getBucketClassByName using:
bucket class name from ObjectBucket.spec.additionalState.bucketclass
OBC namespace from ObjectBucket.spec.claimRef.namespace
4. Added unit tests for getBucketClassByName and a GetQuotaConfig test to confirm the minimum quota is applied when both BC and OBC set limits.

1. https://redhat.atlassian.net/browse/DFBUGS-7167

### Testing Instructions:
1. Create a BucketClass with quota, for example maxObjects=1000, maxSize=20Gi.
2. Create an OBC using that BucketClass with different quota, for example maxObjects=2000, maxSize=10Gi.
3. Confirm the bucket quota after create uses the minimum (1000 objects, 10Gi).
4. Patch the OBC quota, for example: `oc patch obc quota-test-obc -n openshift-storage --type=merge \ -p '{"spec":{"additionalConfig":{"maxSize":"12Gi","maxObjects":"2000","bucketclass":"quota-test-bc"}}}'`
5. Confirm the bucket quota after reconcile is 1000 objects and 12Gi (not 2000).

Run unit tests:

go test ./pkg/obc/...

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Tests**
  * Added Ginkgo/Gomega coverage for bucket class lookup with namespace precedence (OBC vs system), empty/no-match scenarios, and quota selection when both BucketClass and OBC specify `maxObjects` (smaller value wins).

* **Refactor**
  * Improved bucket class resolution for existing bucket operations by using the claim’s namespace when available and applying consistent by-name lookup semantics with clear fallback behavior.

### DIFF
--- a/pkg/obc/obc_suite_test.go
+++ b/pkg/obc/obc_suite_test.go
@@ -1,4 +1,4 @@
-package obc_test
+package obc
 
 import (
 	"testing"
--- a/pkg/obc/provisioner.go
+++ b/pkg/obc/provisioner.go
@@ -454,7 +454,11 @@ func NewBucketRequest(
 		r.AccountName = ob.Spec.AdditionalState["account"]
 		bucketClassName := ob.Spec.AdditionalState["bucketclass"]
 
-		bucketClass, exists := getBucketClass(r.OBC, bucketOptions, p.Namespace, util.KubeCheck)
+		obcNamespace := ""
+		if ob.Spec.ClaimRef != nil {
+			obcNamespace = ob.Spec.ClaimRef.Namespace
+		}
+		bucketClass, exists := getBucketClassByName(bucketClassName, obcNamespace, p.Namespace, util.KubeCheck)
 		if !exists {
 			p.Logger.Warnf("BucketClass %q not found in namespace %q", bucketClassName, p.Namespace)
 		}
@@ -910,16 +914,10 @@ func (r *BucketRequest) updateReplicationPolicy(ob *nbv1.ObjectBucket) error {
 	return nil
 }
 
-// getBucketClass takes an OBC, bucketoptions and provisioner namespace and returns the bucketClass
-//
-// If BucketClass name is not specified in the OBC, then the empty string is returned with exists=false
-// If BucketClass name is specified in the OBC, then:
-// - if the bucketclass is found in the obc namespace, then that bucketclass is returned
-// with exists=true
-// - if the bucketclass is found in the provisioner namespace, then that buckeclass is
-// returned with exists=true
-// - if the bucketclass is not found in the obc namespace or the provisioner namespace, then the
-// bucketclass with namespace set to provisioner namespace is returned with exists=false
+// getBucketClass extracts the bucket class name from an OBC (or StorageClass parameters)
+// and delegates to getBucketClassByName. Used by the create/provision path where a full
+// ObjectBucketClaim is available. On the update path (ObjectBucket only), callers should
+// use getBucketClassByName directly with values from AdditionalState and ClaimRef.
 func getBucketClass(
 	obc *nbv1.ObjectBucketClaim,
 	bucketOptions *obAPI.BucketOptions,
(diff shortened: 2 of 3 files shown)


pkg/obc/provisioner.go
```diff
@@ -454,7 +454,11 @@ func NewBucketRequest(
 		r.AccountName = ob.Spec.AdditionalState["account"]
 		bucketClassName := ob.Spec.AdditionalState["bucketclass"]
 
-		bucketClass, exists := getBucketClass(r.OBC, bucketOptions, p.Namespace, util.KubeCheck)
```
naveenpaul1, line 457: Why getBucketClass() method is not working when we update?

pkg/obc/provisioner.go
naveenpaul1: update the comment

pkg/obc/provisioner.go
```diff
@@ -963,6 +959,31 @@ func getBucketClass(
 	return bucketClass, false
 }
 
+func getBucketClass(
```
naveenpaul1: Add comment and move it to before `getBucketClassByName()`  it will help to review

pkg/obc/provisioner.go
VershaAgrawal: On update, we only get an ObjectBucket, not an ObjectBucketClaim. So r.OBC is never set and stays nil.

The old code still called getBucketClass(r.OBC, ...). When obc is nil, that function returns right away without looking up the BucketClass. That’s why we saw “BucketClass not found” and only OBC quota was applied.

On create, r.OBC is set, so getBucketClass works as expected.

The fix uses getBucketClassByName() on the update path, with the bucket class name from ob.Spec.AdditionalState["bucketclass"] and the namespace from ob.Spec.ClaimRef.Namespace

pkg/obc/obc_suite_test.go
```diff
@@ -1,4 +1,4 @@
-package obc_test
+package obc
```
naveenpaul1, line 1: why updated this?

pkg/obc/provisioner.go
```diff
@@ -454,7 +454,11 @@ func NewBucketRequest(
 		r.AccountName = ob.Spec.AdditionalState["account"]
 		bucketClassName := ob.Spec.AdditionalState["bucketclass"]
 
-		bucketClass, exists := getBucketClass(r.OBC, bucketOptions, p.Namespace, util.KubeCheck)
+		obcNamespace := ""
+		if ob.Spec.ClaimRef != nil {
+			obcNamespace = ob.Spec.ClaimRef.Namespace
+		}
+		bucketClass, exists := getBucketClassByName(bucketClassName, obcNamespace, p.Namespace, util.KubeCheck)
```
naveenpaul1, line 461: So my understanding is, bucketClass was missing because namespace was not correct and this PR fix that by getting the namespace from ob.Spec.ClaimRef.Namespace, is that right?

pkg/obc/obc_suite_test.go
VershaAgrawal: So tests can call unexported helpers like getBucketClass and getBucketClassByName

naveenpaul1 (approved):

pkg/obc/provisioner.go
VershaAgrawal: Not exactly. The main issue wasn’t a wrong namespace — on the update path r.OBC was nil, so getBucketClass returned immediately and never looked up the BucketClass, even though the name was in ob.Spec.AdditionalState["bucketclass"].

This PR fixes that by calling getBucketClassByName with that name.

PR #1951, opened 2026-06-12
[Backport to 5.21 ] Harden CNPG PostgreSQL config for replica HA stability

dannyzaken:
### Explain the changes

Backport #1930 
**5.22 Postgres configuration is handled differently, so the actual commit could not be cherry-picked. Changed the config values directly in 5.21 code**

Add WAL retention, replication timeout, and failover delay settings to prevent stuck-replica and bootstrap-gap failures observed in production.

- Add wal_keep_size computed as 12% of data PVC size (min 1GB, max 16GB)
- Set wal_level: replica to reduce ~10-30% WAL overhead from unused logical decoding
- Set wal_sender_timeout and wal_receiver_timeout to 60s (PostgreSQL upstream defaults)
- Set failoverDelay: 30 to avoid unnecessary failovers on brief blips
- Track FailoverDelay in wasClusterSpecChanged for propagation to existing clusters
- Add unit tests for calculateWalKeepSize and formatBytesKB
- `jit` and `wal_compression` are additional enhancements that are less related to the root cause fixed here. Not backporting these values to 5.21

1. https://redhat.atlassian.net/browse/DFBUGS-7330

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

### DIFF
--- a/pkg/system/db_reconciler.go
+++ b/pkg/system/db_reconciler.go
@@ -28,6 +28,8 @@ const (
 
 	noobaaDBUser = "noobaa"
 	noobaaDBName = "nbcore"
+
+	defaultFailoverDelaySec = int32(30)
 )
 
 // ReconcileCNPGCluster reconciles the CNPG cluster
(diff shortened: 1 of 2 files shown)


liranmauda (approved):

liranmauda (changes requested):
The CI is failing.

liranmauda (approved):

PR #1954, opened 2026-06-16
Keycloak | Created cli for keycloak configuration and saved it in a secret

VershaAgrawal:
### Describe the Problem
There was no CLI way to configure Keycloak OIDC for NooBaa STS. Users had to manually create the Kubernetes secret with the Keycloak provider config.

### Explain the changes
1. Added `noobaa system oidc --type keycloak --configure <<configuration in json format>` CLI command.
2. Accepts inline JSON or a file path using file:// (.json or .txt with JSON content).
3. Validates the config (required fields: issuer, client_id, client_secret, jwks_uri, token_introspection_endpoint).
4. Saves the config to secret noobaa-oidc-keycloak-config under key config.json.
5. Triggers operator reconcile so endpoint pods mount the config at /etc/noobaa-server/oidc/keycloak_config/config.json.
6. Added CLI tests for help, invalid input, and missing file cases.

1. https://redhat.atlassian.net/browse/RHSTOR-8956

### Testing Instructions:
1. Run unit tests: `go test ./pkg/cli/... -ginkgo.focus "system oidc"`
2. Create keycloak_config.json with a valid providers array.
3. Run:
```
noobaa system oidc \
  --type keycloak \
  --configure file://keycloak_config.json
```

- [ ] Doc added/updated
- [x] Tests added

## New Features
- Added OIDC configuration support for NooBaa systems using Keycloak.
- Introduced the `noobaa system oidc` CLI command with `--type` and `--configure`.
- Accepts inline JSON or `file://` configuration input, validates it, and saves the result to a Kubernetes Secret.

## Tests
- Added Ginkgo coverage for `--help` output and `--configure` validation failures (unsupported type, invalid JSON, missing required provider fields, and missing config file).

### DIFF
--- a/pkg/system/system.go
+++ b/pkg/system/system.go
@@ -55,6 +55,7 @@ func Cmd() *cobra.Command {
 		CmdList(),
 		CmdReconcile(),
 		CmdYaml(),
+		CmdOidc(),
 	)
 	return cmd
 }
@@ -154,6 +155,45 @@ func CmdYaml() *cobra.Command {
 		Run:   RunYaml,
 		Args:  cobra.NoArgs,
 	}
+	return cmd
+}
+
+const keycloakOIDCConfigKey = "config.json"
+
+// KeycloakOIDCProvider represents a single Keycloak OIDC provider configuration.
+type KeycloakOIDCProvider struct {
+	Issuer                     string `json:"issuer"`
+	ClientID                   string `json:"client_id"`
+	ClientSecret               string `json:"client_secret"`
+	JWKSURI                    string `json:"jwks_uri"`
+	TokenIntrospectionEndpoint string `json:"token_introspection_endpoint"`
+}
+
+// KeycloakOIDCConfig represents the Keycloak OIDC configuration stored in the secret.
+type KeycloakOIDCConfig struct {
+	Providers []KeycloakOIDCProvider `json:"providers"`
+}
+
+// CmdOidc returns a CLI command for managing OIDC configuration.
+func CmdOidc() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "oidc",
+		Short: "Manage OIDC configuration",
+		Long: "Configure OIDC providers for NooBaa STS.\n" +
+			"The configuration is stored in a Kubernetes secret and mounted into endpoint pods.",
+		Run: RunOidc,
+	}
+	cmd.Flags().String("type", "", "OIDC provider type (keycloak)")
+	cmd.Flags().String(
+		"configure",
+		"",
+		"OIDC configuration in JSON format, or a path to a JSON/txt file prefixed with file:// (e.g. file://keycloak_config.json).\n"+
+			"For Keycloak, the expected structure is:\n"+
+			`{"providers":[{"issuer":"<kc-server>:<kc-port>/realms/<realm-name>",`+
+			`"client_id":"<client-id>","client_secret":"<client-secret>",`+
+			`"jwks_uri":"http://<kc-server>:<kc-port>/realms/<realm-name>/protocol/openid-connect/certs",`+
+			`"token_introspection_endpoint":"http://<kc-server>:<kc-port>/realms/<realm-name>/protocol/openid-connect/token/introspect"}]}`,
+	)
 	return cmd
 }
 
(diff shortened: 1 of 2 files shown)


pkg/system/system.go
```diff
@@ -691,6 +731,128 @@ func RunYaml(cmd *cobra.Command, args []string) {
...
+	case "keycloak":
+		configJSON, err := readOIDCConfigInput(configure)
+		if err != nil {
+			log.Fatalf(`❌ %v`, err)
+		}
+		if err := configureKeycloakOIDC(configJSON); err != nil {
+			log.Fatalf(`❌ %v`, err)
+		}
+	default:
+		log.Fatalf(`❌ Unsupported OIDC provider type %q. Supported types: keycloak`, providerType)
```
naveenpaul1: can we make sure its ageneral message, If LDAP wanted to use this that should be easy

pkg/system/system.go
```diff
@@ -691,6 +731,128 @@ func RunYaml(cmd *cobra.Command, args []string) {
...
+		return fmt.Errorf("failed to marshal configuration: %w", err)
+	}
+
+	secret := util.KubeObject(bundle.File_deploy_internal_secret_empty_yaml).(*corev1.Secret)
+	secret.Name = options.SystemName + "-oidc-keycloak-config"
+	secret.Namespace = options.Namespace
+	secret.StringData = map[string]string{
+		keycloakOIDCConfigKey: string(normalizedJSON),
+	}
+	secret.Data = nil
```
naveenpaul1, line 841: Do we need this?

pkg/system/system.go
```diff
@@ -691,6 +731,128 @@ func RunYaml(cmd *cobra.Command, args []string) {
...
+	log.Printf("⏳ Reconciling system to apply the configuration...")
+
+	klient := util.KubeClient()
+	req := reconcile.Request{
+		NamespacedName: types.NamespacedName{
+			Namespace: options.Namespace,
+			Name:      options.SystemName,
+		},
+	}
+	if _, err := NewReconciler(req.NamespacedName, klient, scheme.Scheme, nil).Reconcile(); err != nil {
```
naveenpaul1: Without NewReconciler() our changes are not reflecting in noobaa endpoint?

pkg/system/system.go
```diff
@@ -691,6 +731,128 @@ func RunYaml(cmd *cobra.Command, args []string) {
...
+	if !json.Valid([]byte(configJSON)) {
+		return nil, fmt.Errorf("the provided configuration is not valid JSON")
+	}
+
+	var config KeycloakOIDCConfig
+	if err := json.Unmarshal([]byte(configJSON), &config); err != nil {
+		return nil, fmt.Errorf("failed to parse configuration: %w", err)
+	}
+
+	if len(config.Providers) == 0 {
```
naveenpaul1, line 784: can you aslo check there is provider with empty array

pkg/system/system.go
VershaAgrawal: Yes. The operator doesn't watch this secret directly. Without calling Reconcile(), the endpoint deployment may not get the volume mount promptly . Manual reconcile ensures the mount is applied immediately after the secret is saved.

pkg/system/system.go
VershaAgrawal: Yes — this follows the existing NooBaa pattern when creating/updating secrets with StringData. It avoids stale Data when updating an existing secret.

pkg/system/system.go
VershaAgrawal: Empty provider objects  are already rejected via required-field checks. I've add an explicit empty-provider check in the for loop now.

pkg/system/system.go
naveenpaul1: but in next reconcile loop it will get pick that secret, right? It wont that much time, Could you please check that too

VershaAgrawal: Yes, right .that it would eventually get picked up on a later reconcile . Creating the secret via CLI doesn't enqueue a reconcile by itself. The next reconcile depends on another trigger, so timing is unpredictable. Calling Reconcile() immediately after saving the secret ensures the endpoint mount is applied right away.

VershaAgrawal: Removed the NecReconcile() trigger after the save secret.

naveenpaul1 (approved):

PR #1961, opened 2026-06-18
GCP WIF (STS) | Analyze Resource | Exclude GCP WIF (STS)

shirady:
### Describe the Problem
Exclude GCP WIF (STS) from analyze resource, as the job does not have the projected token (and added permission in the configuration from the user).

### Explain the changes
1. Rename the functions `IsSTSClusterBS` and `IsSTSClusterNS` and add AWS to improve readability since STS clusters have more types today (`IsAWSSTSClusterBS`and `IsAWSSTSClusterNS` respectively).
Note: I intentionally saved it in a separate commit. 
2. Exclude the STS types from analyze resource:
- In `checkIfBackingStoreTypeIsSupported` added GCP WIF (STS), and also noticed that Azure STS was missing.
- In `checkIfNamespaceStoreTypeIsSupported` added.
3. Added helper functions: `isGoogleSTSClusterStore`, `IsGoogleSTSClusterBS`, `IsGoogleSTSClusterNS`, and `isGoogleSTSSecret` to check if a namespacestore and backingstore with GCP WIF (STS).
4. In `setBackingStoreDetailsInJob` and `setNamespacetoreDetailsInJob` exit and do not create the job if the secret doe not exist (fail fast).
5. Removed unused flag of bucket in analyze backingstore and namespacestore.

### Issues: 
1. none

### Testing Instructions:

1. Deploy NooBaa on Minikube after adding GCP WIF configurations - see guide "Create GCP WIF (STS) Setup On Minikube With NooBaa" ([link](https://github.com/noobaa/noobaa-operator/blob/master/doc/dev_guide/create_gcp_wif_setup_on_minikube_with_noobaa.md)).
2. Create GCP WIF (STS) backingstore using `nooba backingstore create google-cloud-storage-sts <backingstore-name> -n <namespace>`.
3. Create GCP WIF (STS) namespacestore with a YAML and apply it:
- Copy the name of the secret from the created backingstore that was created with the sub-command of `google-cloud-storage-sts`
- Create a target bucket on the GCP web console

```yaml
apiVersion: noobaa.io/v1alpha1
kind: NamespaceStore
metadata:
  name: <namesapcestore-name>
  namespace: test1
  finalizers:
    - noobaa.io/finalizer
spec:
  type: google-cloud-storage
  googleCloudStorage:
    targetBucket: <target-bucket-name>
    secret:
      name: <secret-name>
      namespace: test1
```

apply it: `kubectl apply -f <path-to-file>`

4. Create GCP backingstore using `nooba backingstore create google-cloud-storage <backingstore-name> -n <namespace>`.
5. Create GCP namespacestore with a YAML and apply it:
- Copy the name of the secret from the created backingstore that was created with the sub-command of `google-cloud-storage` and Create a target bucket on the GCP web console
6. Run the analyze resource: `nb diagnostics analyze resource -n test1`.
7. Check that the job runs and you can see the output from the commands and the created logs: GCP types should be analyzed, GCP WIF (STS) should not.
Note: you can also test it separately with:
- `nb diagnostics analyze backingstore <backingstore-name> -n test1`
- `nb diagnostics analyze namesapcestore <namespacestore-name> -n test1`

- [ ] Doc added/updated
- [ ] Tests added

## Summary

* **New Features**
  * Added detection for Google Cloud Workload Identity Federation (WIF) credentials for both backing stores and namespace stores.

* **Improvements**
  * Improved AWS/AWS-ST S handling by using AWS-specific STS detection when selecting secret behavior and external connection parameters.
  * Strengthened analyze diagnostics: fails earlier when secret references are missing/invalid and includes clearer resource and secret-identifying details.

* **Chores**
  * Removed the unused `--bucket` flag from diagnostics analyze commands.

### DIFF
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -1192,16 +1192,16 @@ func IsFusionHCIWithScale() bool {
 	return fusionHCIWithScale
 }
 
-// IsSTSClusterBS returns true if it is running on an STS cluster
-func IsSTSClusterBS(bs *nbv1.BackingStore) bool {
+// IsAWSSTSClusterBS returns true if it is running on an STS cluster
+func IsAWSSTSClusterBS(bs *nbv1.BackingStore) bool {
 	if bs.Spec.Type == nbv1.StoreTypeAWSS3 {
 		return bs.Spec.AWSS3.AWSSTSRoleARN != nil
 	}
 	return false
 }
 
-// IsSTSClusterNS returns true if the namespace store uses AWS STS (short-lived credentials). For Azure STS use IsAzureSTSClusterNS().
-func IsSTSClusterNS(ns *nbv1.NamespaceStore) bool {
+// IsAWSSTSClusterNS returns true if the namespace store uses AWS STS (short-lived credentials). For Azure STS use IsAzureSTSClusterNS().
+func IsAWSSTSClusterNS(ns *nbv1.NamespaceStore) bool {
 	if ns.Spec.Type == nbv1.NSStoreTypeAWSS3 {
 		return ns.Spec.AWSS3.AWSSTSRoleARN != nil
 	}
--- a/pkg/util/util.go
+++ b/pkg/util/util.go
@@ -1224,6 +1224,55 @@ func IsAzureSTSClusterNS(ns *nbv1.NamespaceStore) bool {
 	return false
 }
 
+// isGoogleSTSClusterStore is a helper function that checks if a store uses GCP WIF (STS) (short-lived credentials).
+func isGoogleSTSClusterStore(isGoogleCloudStorage bool, secretRef *corev1.SecretReference, err error) bool {
+	if !isGoogleCloudStorage {
+		return false
+	}
+	if err != nil || secretRef == nil || secretRef.Name == "" {
+		return false
+	}
+	secret, err := GetSecretFromSecretReference(secretRef)
+	if err != nil {
+		return false
+	}
+	return isGoogleSTSSecret(secret)
+}
+
+// IsGoogleSTSClusterBS returns true if the backing store uses GCP WIF (STS) (short-lived credentials).
+func IsGoogleSTSClusterBS(bs *nbv1.BackingStore) bool {
+	isGoogleCloudStorage := bs.Spec.Type == nbv1.StoreTypeGoogleCloudStorage && bs.Spec.GoogleCloudStorage != nil
+	secretRef, err := GetBackingStoreSecret(bs)
+	return isGoogleSTSClusterStore(isGoogleCloudStorage, secretRef, err)
+}
+
+// IsGoogleSTSClusterNS returns true if the namespace store uses GCP WIF (STS) (short-lived credentials).
+func IsGoogleSTSClusterNS(ns *nbv1.NamespaceStore) bool {
+	isGoogleCloudStorage := ns.Spec.Type == nbv1.NSStoreTypeGoogleCloudStorage && ns.Spec.GoogleCloudStorage != nil
+	secretRef, err := GetNamespaceStoreSecret(ns)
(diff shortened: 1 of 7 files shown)


pkg/util/util.go
```diff
@@ -1219,6 +1224,51 @@ func IsAzureSTSClusterNS(ns *nbv1.NamespaceStore) bool {
...
+	secretRef, err := GetNamespaceStoreSecret(ns)
+	return isGoogleSTSClusterStore(isGoogleCloudStorage, secretRef, err)
+}
+
+// isGoogleSTSSecret returns true if the secret contains a GCP WIF (STS) (short-lived credentials).
+func isGoogleSTSSecret(secret *corev1.Secret) bool {
+	if secret == nil {
+		return false
+	}
+	if wifJSON := secret.StringData[GoogleCredentialsJson]; wifJSON != "" {
```
kajalpareek-lab: isGoogleSTSSecret reads from secret.StringData, but when you fetch a secret from the Kubernetes API, the values are in secret.Data (base64 []byte), not StringData. StringData is write-only — it's never populated on read. So I guess this function will always return false, and GCP WIF backing stores won't actually get excluded from the analyze job.

Should use string(secret.Data[GoogleCredentialsJson]) instead of secret.StringData[GoogleCredentialsJson].

pkg/util/util.go
```diff
@@ -1219,6 +1224,51 @@ func IsAzureSTSClusterNS(ns *nbv1.NamespaceStore) bool {
...
+		return false
+	}
+	return isGoogleSTSSecret(secret)
+}
+
+// IsGoogleSTSClusterBS returns true if the backing store uses GCP WIF (STS) (short-lived credentials).
+func IsGoogleSTSClusterBS(bs *nbv1.BackingStore) bool {
+	isGoogleCloudStorage := bs.Spec.Type == nbv1.StoreTypeGoogleCloudStorage && bs.Spec.GoogleCloudStorage != nil
+	secretRef, err := GetBackingStoreSecret(bs)
+	return isGoogleSTSClusterStore(isGoogleCloudStorage, secretRef, err)
```
naveenpaul1, line 1246: Is there any dchance isGoogleCloudStorage can have false value? we will always call this method for GCP flow right?

naveenpaul1 (commented):
LGTM

pkg/util/util.go
shirady: The function in the flow is working as expected, since currently `isGoogleSTSSecret` (the function you pointed to) is only called from `isGoogleSTSClusterStore` which loads the secret via `GetSecretFromSecretReference` which calls `KubeCheck` which calls `SecretResetStringDataFromData` copies `secret.Data` into `secret.StringData` before the secret is returned. The GCP WIF (STS) backingstore was excluded from the analyze resource as expected.

Since the function was added in the `util` pkg and I don't want any confusion, I added a helper function `secretDataString`.

So in the case of GCP WIF (STS) backingstore the output is (as expected):
`nb diagnostics analyze backingstore shira-gcp-sts -n test1`
```
INFO[0000] ✅ Exists: BackingStore "shira-gcp-sts"
INFO[0000] ✅ Exists: Secret "backing-store-google-cloud-storage-shira-gcp-sts"
INFO[0000] BackingStore shira-gcp-sts type is not supported in analyze
```

pkg/util/util.go
shirady: The function `IsGoogleSTSClusterBS` is currently used only by the analyze resource command.
`isGoogleCloudStorage` returns `false` when it is GCP (not STS), `true` for GCP WIF (STS).
Therefore, the GCP BS is analyzed, and the GCP BS STS is not analyzed (excluded).

naveenpaul1 (approved):

kajalpareek-lab (approved):
Looks good to me.

PR #1969, opened 2026-06-23
Add dev-env and mini-env performance profiles for CLI deployments

dannyzaken:
### Describe the Problem

The `--dev` and `--mini` CLI flags hardcoded resource values inline in
`system.go`, bypassing the performance profile system. This meant they
couldn't leverage profile-only settings like `pvPoolNumVolumes` and
`dbInstances`, and the resource definitions were duplicated outside the
centralized profile map.

### Explain the Changes
1. Added `dev-env` and `mini-env` performance profile types to `noobaa_types.go`
   with kubebuilder enum validation.
2. Added corresponding profile entries in `performance_profiles.go` with resource
   values matching the previous inline definitions (mini-env: minimal footprint
   with 100m CPU / 500Mi memory for most components; dev-env: moderate resources
   with 500m-1 CPU / 500Mi-2Gi memory). Both profiles set `dbInstances: 1` and
   `pvPoolNumVolumes: 1`.
3. Added a `logResources` field to the `performanceProfile` struct (as a pointer,
   only set by mini-env) and a `getLogResources()` accessor function.
4. Updated `phase2_creating.go` to use `getLogResources()` so the log-processor
   container picks up profile-defined log resources during reconciliation.
5. Replaced ~60 lines of inline resource construction in `system.go` with
   single-line `sys.Spec.PerformanceProfile` assignments.  
6. Change default CNPG image to `quay.io/rhceph-dev/odf4-odf-cloudnative-pg-rhel9-operator:v4.22` 
7. Added an option `--postgres-instances` to set the number of DB instances and changed the dev/mini envs to run with a single instance.

1.

### Testing Instructions:
1. Deploy with `noobaa install --mini` and verify core/db/endpoint pods use
   mini-env resource values (100m CPU, 500Mi-1Gi memory) and log-processor
   uses 50m CPU / 200Mi memory.
2. Deploy with `noobaa install --dev` and verify pods use dev-env resource
   values (500m-1 CPU, 500Mi-2Gi memory).
3. Deploy without flags and verify the default profile is applied unchanged.
4. Verify the default pv-pool backingstore is created with `NumVolumes: 1`
   when using `--mini` or `--dev`.
5. Verify that setting explicit resource overrides on the CR spec still takes
   precedence over the profile values.

- [ ] Doc added/updated
- [ ] Tests added

## Summary

* **New Features**
  * Added two new performance profile options: `dev-env` and `mini-env`.
  * Updated the CRD and bundled deployment schema to accept the new profile values.

* **Improvements**
  * Enhanced log resource configuration for the log processor to follow the selected performance profile (or use explicitly provided log settings when set).
  * Updated mini/dev default resource sizing to follow the chosen performance profile automatically.

* **Updates**
  * Changed the default managed Postgres instance count to `1` and added a `--postgres-instances` CLI option.
  * Updated the default CloudNativePG operator image tag.

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -76,7 +76,7 @@ var DBImage = "quay.io/sclorg/postgresql-16-c9s"
 var PostgresMajorVersion = 16
 
 // PostgresInstances is the default number of postgres instances in a managed postgres cluster
-var PostgresInstances = 2
+var PostgresInstances = 1
 
 // Psql12Image is the default postgres12 db image url
 // currently it can not be overridden.
@@ -187,7 +187,7 @@ var AWSSTSARN = ""
 var CnpgVersion = "1.27.0"
 
 // CnpgImage is the container image url of cloudnative-pg operator
-var CnpgImage = "quay.io/noobaa/cloudnative-pg-noobaa:v1.25.0"
+var CnpgImage = "quay.io/rhceph-dev/odf4-odf-cloudnative-pg-rhel9-operator:v4.22"
 
 // UseCnpgApiGroup indicates if the original CloudNativePG API group should be used for the installation manifests
 // Relevant for the CLI commands. during reconciliation we look at the env variable USE_CNPG_API_GROUP to determine
@@ -292,6 +292,10 @@ func init() {
 	FlagSet.IntVar(
 		&DBVolumeSizeGB, "db-volume-size-gb",
 		DBVolumeSizeGB, "The database volume size in GB",
+	)
+	FlagSet.IntVar(
+		&PostgresInstances, "postgres-instances",
+		PostgresInstances, "The number of postgres instances",
 	)
 	FlagSet.StringVar(
 		&DBStorageClass, "db-storage-class",
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -783,8 +783,8 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 				c.Image = r.NooBaa.Status.ActualImage
 			}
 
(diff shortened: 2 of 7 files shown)


pkg/options/options.go
```diff
@@ -187,7 +187,7 @@ var AWSSTSARN = ""
 var CnpgVersion = "1.27.0"
 
 // CnpgImage is the container image url of cloudnative-pg operator
-var CnpgImage = "quay.io/noobaa/cloudnative-pg-noobaa:v1.25.0"
+var CnpgImage = "quay.io/rhceph-dev/odf4-odf-cloudnative-pg-rhel9-operator:v4.22"
```
shirady, line 190: Why do you change it as part of the addition of the flags?

pkg/system/phase2_creating.go
```diff
@@ -783,8 +783,8 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 				c.Image = r.NooBaa.Status.ActualImage
 			}
 
-			if r.NooBaa.Spec.LogResources != nil {
-				c.Resources = *r.NooBaa.Spec.LogResources
+			if logRes := getLogResources(r.NooBaa); logRes != nil {
```
shirady: I suggest renaming the `logRes` to the full name `logResources`, since when reading without context, the `res` suffix can be many things.

shirady (approved):
LGTM

pkg/options/options.go
dannyzaken: This is related to dev envs, as `CnpgImage` is mainly used in this flow and in the tests. I changed it to align with the acutal version we use.

PR #1975, opened 2026-06-28
Update Archive Policy of buckets via NB CLI and bucketclass reconciler

romayalon:
### Explain the changes
1. Added to `nb bucket update` an option to update/remove archive policy that calls update_bucket api.
2. Added changes to bucketclass reconciler for updating archive policy.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary

* **New Features**
  * Added deep-archive policy updates to bucket updates via `--deep-archive-resource`.
  * Added `--remove-archive-policy` to explicitly remove an archive policy.
  * `bucket status` now shows deep-archive archive resource/path details when configured.
  * BucketClass reconciliation now applies deep-archive policy to existing buckets that don’t have one yet.
* **Bug Fixes**
  * Deep-archive setup now validates archive-enabled S3-compatible namespace stores only.
  * Rejects using deep-archive setup and archive-policy removal flags together.
* **Tests**
  * Added coverage for new flags and deep-archive eligibility/update behavior.

### DIFF
--- a/pkg/bucket/bucket.go
+++ b/pkg/bucket/bucket.go
@@ -5,14 +5,15 @@ import (
 	"os"
 	"strconv"
 
+	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	"github.com/noobaa/noobaa-operator/v5/pkg/bucketclass"
 	"github.com/noobaa/noobaa-operator/v5/pkg/nb"
 	"github.com/noobaa/noobaa-operator/v5/pkg/options"
 	"github.com/noobaa/noobaa-operator/v5/pkg/system"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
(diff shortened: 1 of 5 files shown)


pkg/bucket/bucket.go
```diff
@@ -172,6 +188,30 @@ func RunUpdate(cmd *cobra.Command, args []string) {
 		updateParams.Quota = &quota
 	}
 
+	if deepArchiveResource != "" {
+		archivePolicy, err := buildArchivePolicyForUpdate(deepArchiveResource)
```
karthikps97: Do we need the error here since `buildArchivePolicyForUpdate` always returns `nil`. If we will add some validations later on, then we can keep this

romayalon: it's a leftover, removed it and did some improvements in that area, thanks!

karthikps97 (approved):

PR #2015, opened 2026-07-09
Analyze Resource | Fix CA Changes (To Use Unsecured Agent)

shirady:
### Describe the Problem
While I worked on another feature, I noticed an error in the setup when using s3-compatible (OBM COS) with the changes that were made regarding the TLS (noobaa/noobaa-core#9395).

### Explain the changes
1. Add `automountServiceAccountToken: false` in the analyze resource job.
2. Clean up the env `NODE_EXTRA_CA_CERTS` as it is not used in noobaa-core anymore.

### Issues: 
1. Before the change, I noticed that in IBM COS, it had an error: `self-signed certificate in certificate chain`.
2. After the change, all the tests are passing. The explanation for this (using AI): 

> With `automountServiceAccountToken: false`:
> - `service-ca.crt` is not mounted → `INTERNAL_CA_CERTS` read fails → no CA loaded
> - No OCP CA bundle on this job either → `EXTERNAL_CA_CERTS` also missing
> - `https_agent.options.ca` is empty → `get_unsecured_agent` skips TLS verification for non-AWS endpoints
> That matches block-store agents, which also set `automountServiceAccountToken: false` and therefore never see `service-ca.crt`.

reference: 
https://github.com/noobaa/noobaa-core/blob/646bdc10b8565d4486becda601642eb7cb753ce0/src/util/http_utils.js#L39-L40

### Testing Instructions:
1. Use the Simple Jenkins Deploy with OCP deployed with ODF (includes NooBaa).
2. Use the CLI and run: `nb diagnostics analyze backingstore noobaa-default-backing-store -n openshift-storage` (expect all tests to pass).

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Disabled automatic service account token mounting for resource analysis jobs to enhance security.

* **Chores**
  * Simplified environment variable handling by removing unnecessary certificate configuration while preserving proxy settings for resource analysis operations.

### DIFF
--- a/deploy/job_analyze_resource.yml
+++ b/deploy/job_analyze_resource.yml
@@ -12,6 +12,7 @@ spec:
   ttlSecondsAfterFinished: 10
   template:
     spec:
+      automountServiceAccountToken: false
       volumes:
       - name: cloud-credentials
         secret:
@@ -30,7 +31,6 @@ spec:
           - name: HTTP_PROXY
           - name: HTTPS_PROXY
           - name: NO_PROXY
-          - name: NODE_EXTRA_CA_CERTS
         command: 
             - /bin/bash
             - -c
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -5958,7 +5958,7 @@ const File_deploy_internal_text_system_status_readme_rejected_tmpl = `
 	NooBaa Operator Version: {{.OperatorVersion}}
 `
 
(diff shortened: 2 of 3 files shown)


deploy/job_analyze_resource.yml
```diff
@@ -12,6 +12,7 @@ spec:
   ttlSecondsAfterFinished: 10
   template:
     spec:
+      automountServiceAccountToken: false
```
jackyalbo, line 15: Shouldn't we support the service account token for internal servers? Let's say in case RGW is the resource?

shirady:
Closing, as it was not planned.
I can reopen when time permits.

PR #2024, opened 2026-07-14
Add PV pool resources to perf profiles + ibm z profile

dannyzaken:
### Describe the Problem
PV pool backingstore agent pods used hardcoded minimum CPU/memory values
that varied only by environment variables (TEST_ENV, DEV_ENV). These
values were enforced with a PersistentError that rejected backingstores
if resources fell below the minimum, and were not tied to the
performance profile system used by all other NooBaa components.

Additionally, the CLI validation for CPU/memory requests vs. limits was
insufficient — if a user provided only a request flag without a
corresponding limit flag, CLI validation passed but the reconciler would
apply the request against profile defaults with lower limits, creating
invalid Kubernetes resources (request > limit).

### Explain the Changes
1. Added `pvPoolResources` field to the `performanceProfile` struct with
   per-profile CPU/memory values (requests=limits):
   default (400m/800Mi), default-ibm-z (200m req, 400m lim / 800Mi),
   mixed-workload and small-objects (1/2Gi), dev-env (500m/500Mi),
   mini-env (100m/400Mi).
2. Added new `PerformanceProfileDefaultIBMZ` (`"default-ibm-z"`) profile
   with the same values as the default profile but CPU **requests** halved
   (0.5x factor) to accommodate IBM Z (s390x) environments.
3. Added exported `GetPVPoolResources()` function that returns profile-based
   defaults, with a test-env override (50m/200Mi).
4. Refactored `updatePodResourcesTemplate` in the backingstore reconciler to
   use profile defaults instead of env-based minimums, and replaced
   `reconcileResources` (which returned PersistentError on under-minimum
   values) with `applyResourceOverrides` that selectively merges
   user-specified VolumeResources over profile defaults.
5. Added `validateResourceRequestsVsLimits` in the reconciler to catch
   request > limit after merging user overrides with profile defaults.
6. Removed hardcoded `minCPUString`/`minMemoryString` constants and
   `getMinimalResourcesByEnv()` function.
7. Updated CLI `CmdCreatePVPool` flags to default to empty (profile-determined)
   instead of hardcoded values; CPU/memory are only set in VolumeResources
   when explicitly provided by the user. CLI-level request vs. limit
   validation only runs when both values are explicitly provided.
8. Added `ValidatePvpoolResources` admission validation in
   `backingstore_validations.go` that checks for non-negative resource
   values and request <= limit for CPU and memory, wired into
   `ValidateBackingStore` for `StoreTypePVPool`.
9. Added `TestGetPVPoolResources` covering all profiles and fallback, updated
   `TestProfileResourcesValues` and `assertPerformanceProfile` to validate
   pvPool resources including the new `default-ibm-z` profile.

1.

### Testing Instructions:
1. Run `go test ./pkg/system/...` — all profile tests pass including new
   `TestGetPVPoolResources` and `default-ibm-z` profile values.
2. Run `go test ./pkg/validations/...` — all validation tests pass.
3. Run `go build ./...` — clean build with no errors.
4. Deploy with different performance profiles and verify PV pool agent pods
   get the expected CPU/memory from the profile.
5. Deploy with `performanceProfile: default-ibm-z` and verify CPU requests
   are halved compared to the default profile while limits remain unchanged.
6. Create a PV pool backingstore via CLI without `--request-cpu`/`--request-memory`
   flags and verify it uses profile defaults.
7. Create a PV pool backingstore via CLI with explicit resource flags and verify
   the user values override the profile defaults.
8. Create a PV pool backingstore with `--request-cpu` exceeding the profile's
   default limit (without `--limit-cpu`) and verify the reconciler rejects it
   with a PersistentError.

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

- **New Features**
  - Added the `default-ibm-z` performance profile option.
  - PV pool CPU and memory resources now come from the selected performance profile.
  - PV pool CPU/memory requests and limits support overrides when explicitly configured.

- **Bug Fixes**
  - Improved PV pool validation to reject negative CPU/memory requests or limits.
  - Enforced CPU/memory request-vs-limit correctness (request must not exceed limit when both are set).

- **Tests**
  - Added/updated unit tests to cover PV pool resource computation and new profile expectations.

### DIFF
--- a/deploy/crds/noobaa.io_noobaas.yaml
+++ b/deploy/crds/noobaa.io_noobaas.yaml
@@ -1946,6 +1946,7 @@ spec:
                   Explicit per-component resource/count fields take precedence over the profile.
                 enum:
                 - default
+                - default-ibm-z
                 - mixed-workload
                 - small-objects
                 - dev-env
--- a/pkg/apis/noobaa/v1alpha1/noobaa_types.go
+++ b/pkg/apis/noobaa/v1alpha1/noobaa_types.go
@@ -262,7 +262,7 @@ type NooBaaSpec struct {
 	// settings for the core, db and endpoint components.
 	// Explicit per-component resource/count fields take precedence over the profile.
 	// +optional
-	// +kubebuilder:validation:Enum=default;mixed-workload;small-objects;dev-env;mini-env
+	// +kubebuilder:validation:Enum=default;default-ibm-z;mixed-workload;small-objects;dev-env;mini-env
 	PerformanceProfile PerformanceProfileType `json:"performanceProfile,omitempty"`
 }
 
@@ -890,6 +890,8 @@ type PerformanceProfileType string
 const (
 	// PerformanceProfileDefault is the default profile
 	PerformanceProfileDefault PerformanceProfileType = "default"
+	// PerformanceProfileDefaultIBMZ is the default profile adjusted for IBM Z (s390x) with halved CPU requests
+	PerformanceProfileDefaultIBMZ PerformanceProfileType = "default-ibm-z"
 	// PerformanceProfileMixedWorkload is for mixed workload
 	PerformanceProfileMixedWorkload PerformanceProfileType = "mixed-workload"
 	// PerformanceProfileSmallObjects is for small objects workload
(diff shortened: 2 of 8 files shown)


alphaprinz (approved):
LGTM

PR #2036, opened 2026-07-21
core HA - add wire leader election to core pod

nadavMiz:
### Describe the Problem
Today the NooBaa core pod runs as a single active instance. If it fails or is rescheduled, there is no built-in way for a standby core pod to take over cleanly. We need active-passive high availability so only one core pod serves at a time, with automatic failover when the leader is lost.

### Explain the changes
change noobaa core init command to run kubernetes leader elect module before calling core_init
1. add kubernetes leader elect pkg
2. create a new init script `leaderelect.go` to acquire the lease before starting the core pod and then renew it
3. 

### Issues:
1.part of https://redhat.atlassian.net/browse/RHSTOR-7491

### GAPs
1. currently still uses 1 core pod. scaling to 2 and enabling and disabled referred to next PR

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added Kubernetes Lease–based leader election for core pods.
  * Exposed an advanced `leader-elect` CLI subcommand for configuring election behavior.
* **Improvements**
  * Core pod startup now uses a shared leader-election binary and updates core lease naming.
  * Expanded RBAC to allow Lease operations.
  * Improved core reconciliation convergence for commands, environment, and init containers.
  * Updated the default operator image tag.
* **Tests**
  * Added unit tests for `leader-elect` argument parsing, validation, and grace/termination calculations.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -8,7 +8,7 @@ ARG NOOBAA_BIN_PATH=build/_output/bin/noobaa-operator
 # tar is needed for kubectl cp
 RUN microdnf -y install tar
 
-# install operator binary
+# install operator binary (includes Hidden leader-elect subcommand for core pods)
 COPY ${NOOBAA_BIN_PATH} ${OPERATOR}
 
 USER ${USER_UID}
--- a/deploy/internal/statefulset-core.yaml
+++ b/deploy/internal/statefulset-core.yaml
@@ -27,6 +27,8 @@ spec:
       volumes:
         - name: logs
           emptyDir: {}
+        - name: shared-bin
+          emptyDir: {}
         - name: mgmt-secret
           secret:
             secretName: noobaa-mgmt-serving-cert
@@ -48,6 +50,20 @@ spec:
       securityContext:
         runAsUser: 10001
         runAsGroup: 0
+      initContainers:
+        - name: copy-leader-elect
+          image: NOOBAA_OPERATOR_IMAGE
+          command:
+            - cp
+            - /usr/local/bin/noobaa-operator
+            - /noobaa-shared/noobaa-operator
+          volumeMounts:
+            - name: shared-bin
+              mountPath: /noobaa-shared
+          securityContext:
+            runAsNonRoot: true
+            allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
       containers:
         #----------------#
         # CORE CONTAINER #
@@ -60,9 +76,20 @@ spec:
             timeoutSeconds: 2
           image: NOOBAA_CORE_IMAGE
           terminationMessagePolicy: FallbackToLogsOnError
+          command:
+            - /noobaa-shared/noobaa-operator
+            - leader-elect
+            - --
+            - /bin/bash
+            - -ec
+            - |
+              exec /usr/local/bin/node /root/node_modules/noobaa-core/src/cmd/core_init.js
           volumeMounts:
             - name: logs
               mountPath: /log
+            - name: shared-bin
+              mountPath: /noobaa-shared
+              readOnly: true
             - name: mgmt-secret
               mountPath: /etc/mgmt-secret
               readOnly: true
@@ -119,6 +146,8 @@ spec:
                 configMapKeyRef:
                   name: noobaa-config
                   key: NOOBAA_VERSION_AUTH_ENABLED
+            - name: NOOBAA_CORE_LEASE_NAME
+              value: ""
             - name: POSTGRES_HOST
               value: "noobaa-db-pg-0.noobaa-db-pg"
             - name: POSTGRES_HOST_RO
--- a/deploy/role_core.yaml
+++ b/deploy/role_core.yaml
@@ -56,3 +56,11 @@ rules:
   - update
   - list
   - watch
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
+  verbs:
+  - get
+  - create
+  - update
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -5603,7 +5603,7 @@ spec:
       noobaa-s3-svc: "true"
 `
 
-const Sha256_deploy_internal_statefulset_core_yaml = "cc38a4b3cd26c43156962823645c43be26e39ed57a7205d6b390c7849c39651b"
+const Sha256_deploy_internal_statefulset_core_yaml = "1f60f4995b291944111536d9f70b2678ec75dd60f53c0340d2bf6adb5bdd1325"
 
 const File_deploy_internal_statefulset_core_yaml = `apiVersion: apps/v1
 kind: StatefulSet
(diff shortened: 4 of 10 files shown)


build/Dockerfile
```diff
@@ -1,15 +1,18 @@
 FROM registry.access.redhat.com/ubi9/ubi-minimal:latest
 
 ENV OPERATOR=/usr/local/bin/noobaa-operator \
+    LEADER_ELECT=/usr/local/bin/noobaa-leader-elect \
     USER_UID=1001 \
     USER_NAME=noobaa-operator
 ARG NOOBAA_BIN_PATH=build/_output/bin/noobaa-operator
+ARG LEADER_ELECT_BIN_PATH=build/_output/bin/noobaa-leader-elect
```
dannyzaken: Any reason to build a separate binary for `noobaa-leader-elect` instead of using the noobaa-operator binary as the CLI?
Building and using a separate binary requires changes in the DS build (a different Dockerfile), which makes it a bit more complicated. 
a separate binary make sense if the CLI binary is much larger, but I don't think this is the case, is it?

dannyzaken: If you're worried about exposing the command in the CLI menu, you can use `Hidden: true` property for the `cobra.Command`

cmd/leader-elect/main.go
```diff
@@ -0,0 +1,23 @@
...
+import (
+	"fmt"
+	"os"
+
+	"github.com/noobaa/noobaa-operator/v5/pkg/leaderelect"
+	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+	"github.com/sirupsen/logrus"
+)
+
+func main() {
```
dannyzaken: If we use the CLI to run the command, we can drop this.

deploy/internal/statefulset-core.yaml
```diff
@@ -60,9 +75,23 @@ spec:
...
+            # Prefer core_init.js (newer core images); fall back to supervisord (older images).
+            - /noobaa-shared/noobaa-leader-elect
+            - --
+            - /bin/bash
+            - -ec
+            - |
+              if [ -f /root/node_modules/noobaa-core/src/cmd/core_init.js ]; then
+                exec /usr/local/bin/node /root/node_modules/noobaa-core/src/cmd/core_init.js
+              fi
+              exec /usr/bin/supervisord start
```
dannyzaken: I'm not completely convinced we should fall back to the older script. Maybe, to avoid possible edge cases, we should just not run the core until the new image is used. WDYT?

deploy/role_core.yaml
```diff
@@ -47,7 +47,7 @@ rules:
   verbs:
   - use
 - apiGroups:
-  - route.openshift.io
+    - route.openshift.io
```
dannyzaken: Seems unintentional. revert

deploy/internal/statefulset-core.yaml
nadavMiz: the problem is that the CI uses the old image. tests don't pass otherwise

build/Dockerfile
nadavMiz: that what I was worried about. it seemed more correct this way. sure I will use the existing cli and make it hidden

pkg/leaderelect/leaderelect.go
```diff
@@ -0,0 +1,442 @@
...
+		return
+	case <-timer.C:
+		r.log.Infof("leader-elect: grace %v elapsed, SIGKILL process group %d", grace, pid)
+		if err := syscall.Kill(-pid, syscall.SIGKILL); err != nil && err != syscall.ESRCH {
+			r.log.Errorf("leader-elect: kill pgid %d SIGKILL: %v", pid, err)
+		}
+		select {
+		case <-done:
+		case <-time.After(5 * time.Second):
+		}
```
dannyzaken, line 453: If the grace period is reached and we also hit the `SIGKILL` 5 second timeout, the 30 second `terminationGracePeriodSeconds` (in phase2_creating) doesn't leave time to release the lease on cancel. 
We should either increase `terminationGracePeriodSeconds` or reduce the `ShutdownGrace`.

pkg/system/phase2_creating.go
```diff
@@ -647,9 +650,10 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 	r.CoreApp.Spec.ServiceName = r.ServiceMgmt.Name
 
 	podSpec := &r.CoreApp.Spec.Template.Spec
-	// set the termination grace period for noobaa-core pod.
-	// For now we set it to 1 second. A better approach should be to implement a graceful shutdown for the noobaa-core pod when SIGTERM is received.
-	terminationGracePeriodSeconds := int64(1)
+	// Accommodate leader-elect wrapper drain + lease release on SIGTERM.
+	// Previously 1s; raised so the wrapper can stop the child process group
+	// and ReleaseOnCancel the Lease within the pod termination window.
+	terminationGracePeriodSeconds := int64(30)
```
dannyzaken: Set `terminationGracePeriodSeconds` to be `ShutdownGrace` + some margin for releasing the lease (maybe 5-10 seconds?).

pkg/bundle/deploy.go
```diff
@@ -5667,9 +5683,24 @@ spec:
...
+            - /noobaa-shared/noobaa-operator
+            - leader-elect
+            - --
+            - /bin/bash
+            - -ec
+            - |
+              if [ -f /root/node_modules/noobaa-core/src/cmd/core_init.js ]; then
+                exec /usr/local/bin/node /root/node_modules/noobaa-core/src/cmd/core_init.js
+              fi
+              exec /usr/bin/supervisord start
```
dannyzaken: How does `supervisord` behave with SIGTERM? Does it fit with the way we switch over on shutdown? I still think it's safer to just fail here instead of silently running the old image in an HA configuration. 
In any case, you should update the core image in options.go to use a noobaa-core veresion with the new core_init

pkg/leaderelect/leaderelect.go
```diff
@@ -0,0 +1,442 @@
...
+		LostGrace:     defaultLostGrace,
+	}
+}
+
+func bindFlags(cmd *cobra.Command, cfg *Config) {
+	cmd.Flags().StringVar(&cfg.LeaseName, "lease-name", cfg.LeaseName, "Lease object name (default: $NOOBAA_CORE_LEASE_NAME)")
+	cmd.Flags().DurationVar(&cfg.LeaseDuration, "lease-duration", cfg.LeaseDuration, "Lease duration")
+	cmd.Flags().DurationVar(&cfg.RenewDeadline, "renew-deadline", cfg.RenewDeadline, "Leadership renew deadline")
+	cmd.Flags().DurationVar(&cfg.RetryPeriod, "retry-period", cfg.RetryPeriod, "Leadership retry period")
+	cmd.Flags().DurationVar(&cfg.ShutdownGrace, "shutdown-grace", cfg.ShutdownGrace, "Time to wait for child exit after SIGTERM/SIGINT before SIGKILL")
```
dannyzaken, line 115: If we don't plan to pass these values as flags, we should probably remove them. This is more relevant to your next PR maybe

nadavMiz: yes I will remove it in the next PR. in case we will end keeping this

dannyzaken (approved):

PR #2038, opened 2026-07-21
allow rwo in mcg only, use pvc_access_mode_rwo

sakshimunjal:
### Describe the Problem

[DFBUGS-8863](https://redhat.atlassian.net/browse/DFBUGS-8863)

MCG-only (ODF 4.19+) hardcodes NooBaa DB PVC access mode to ReadWriteOncePod. CSI drivers without RWOP support (e.g. HPE) fail PVC provisioning and NooBaa never becomes Ready.

### Explain the changes
1. Add boolean NooBaa CR annotation noobaa.io/pvc_access_mode_rwo (true → RWO; missing/other → RWOP).
2. CNPG path (setDesiredStorageConf) reads the annotation and sets DB PVC access mode accordingly.
3. Add unit tests for the annotation and storage conf wiring.

1. DFBUGS-8863 — Allow RWO via NooBaa CR annotation when storage lacks RWOP (MCG-only).

### Testing Instructions:
1. Unit: go test -mod=mod ./pkg/system/ -run 'TestGetDesiredDBPVCAccessMode|TestSetDesiredStorageConfAccessMode' -count=1 -v
2. Cluster: MCG-only on non-RWOP storage (or lab deny policy) → annotate as below → confirm CNPG accessModes=["ReadWriteOnce"], DB PVCs Bound RWO, NooBaa Ready:
`oc -n openshift-storage annotate noobaa noobaa \
  noobaa.io/pvc_access_mode_rwo=true --overwrite`

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added an annotation to control the database PVC access mode.
  * When the annotation value is exactly `"true"`, the PVC access mode is set to **ReadWriteOnce**; otherwise it defaults to **ReadWriteOncePod**.

* **Bug Fixes**
  * Improved database storage configuration so the selected access mode (and the configured storage class) is applied correctly.

* **Tests**
  * Added unit tests covering default behavior and annotation-driven access-mode selection.

### DIFF
--- a/pkg/system/db_reconciler.go
+++ b/pkg/system/db_reconciler.go
@@ -270,7 +270,7 @@ func (r *Reconciler) reconcileClusterSpec(dbSpec *nbv1.NooBaaDBSpec) error {
 
 	// update db volume resources
 	// update the storage configuration
-	err := setDesiredStorageConf(&r.CNPGCluster.Spec.StorageConfiguration, dbSpec)
+	err := setDesiredStorageConf(&r.CNPGCluster.Spec.StorageConfiguration, dbSpec, r.NooBaa.Annotations)
 	if err != nil {
 		r.cnpgLogError("got error getting desired storage configuration for cnpg cluster. error: %v", err)
 		return err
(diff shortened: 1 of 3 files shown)


pkg/system/db_reconciler.go
```diff
@@ -771,14 +771,18 @@ func getDesiredDbImage(dbSpec *nbv1.NooBaaDBSpec) string {
...
-func setDesiredStorageConf(storageConfiguration *cnpgv1.StorageConfiguration, dbSpec *nbv1.NooBaaDBSpec) error {
+func setDesiredStorageConf(storageConfiguration *cnpgv1.StorageConfiguration, dbSpec *nbv1.NooBaaDBSpec, annotations map[string]string) error {
 	if storageConfiguration == nil {
 		return fmt.Errorf("storage configuration is nil")
 	}
 	if storageConfiguration.PersistentVolumeClaimTemplate == nil {
 		storageConfiguration.PersistentVolumeClaimTemplate = &corev1.PersistentVolumeClaimSpec{}
 	}
-	storageConfiguration.PersistentVolumeClaimTemplate.AccessModes = []corev1.PersistentVolumeAccessMode{corev1.ReadWriteOncePod}
+	accessMode, err := getDesiredDBPVCAccessMode(annotations)
```
dannyzaken: This should only be used as a workaround if the user can't use RWOP, so we should avoid using it as a free form input that can have an invalid value and fail. I think the annotation should be a boolean (something like `noobaa.io/pvc_access_mode_rwo`). If it's `"true"`, we should set RWO, otherwise, set RWOP.

dannyzaken (approved):

srekkas:
Do this annotation works for standalone DB too? Or only for CNPG one?

PR #2058, opened 2026-07-27
Replace default-ibm-z profile with in-place CPU request adjustment on s390x

malayparida2000:
### Describe the Problem
A dedicated `default-ibm-z` performance profile is not the right approach for IBM Z. By agreement, IBM Z should have all the same features as a normal ODF deployment, so NooBaa resource profiles should be usable on IBM Z as well. IBM Z's situation is unique in that its CPUs are very powerful, so having half as much CPU request compared to other architectures makes sense—rather than maintaining a separate profile.

### Explain the changes
1. Remove the `default-ibm-z` performance profile from the API, CRD, and profile map
2. On `runtime.GOARCH == s390x`, halve CPU requests (factor 0.5) for all profile-managed components—core, log, db, endpoint, and pvPool—once at startup in `init()`, matching ocs-operator / ocs-client-operator. Limits are unchanged; explicit resource overrides still take precedence
3. Add unit tests for `adjustCpuResourcesForIbmZ` (whole, fractional, even/odd milli CPU)

Follow Up to
https://github.com/red-hat-storage/ocs-operator/pull/4039

### Testing Instructions:
1. Unit: `go test ./pkg/system/ -run 'TestAdjustCpuResourcesForIbmZ|TestLookupProfile|TestGetCoreResources|TestGetDBResources|TestGetEndpointResources|TestGetLogResources|TestGetPVPoolResources'`
2. Confirm `performanceProfile: default-ibm-z` is no longer a valid enum value on the NooBaa CRD
3. On s390x, verify core/log/db/endpoint/pvPool CPU requests from any profile are halved; explicit resource overrides are unchanged

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added performance profiles for mixed workloads, small objects, development environments, and minimal environments.
  * IBM Z systems now automatically apply optimized CPU request sizing across supported performance profiles.

* **Breaking Changes**
  * Removed the `default-ibm-z` performance profile option. Use the standard `default` profile on IBM Z systems.

### DIFF
--- a/pkg/apis/noobaa/v1alpha1/noobaa_types.go
+++ b/pkg/apis/noobaa/v1alpha1/noobaa_types.go
@@ -262,7 +262,7 @@ type NooBaaSpec struct {
 	// settings for the core, db and endpoint components.
 	// Explicit per-component resource/count fields take precedence over the profile.
 	// +optional
-	// +kubebuilder:validation:Enum=default;default-ibm-z;mixed-workload;small-objects;dev-env;mini-env
+	// +kubebuilder:validation:Enum=default;mixed-workload;small-objects;dev-env;mini-env
 	PerformanceProfile PerformanceProfileType `json:"performanceProfile,omitempty"`
 }
 
@@ -894,8 +894,6 @@ type PerformanceProfileType string
 const (
 	// PerformanceProfileDefault is the default profile
 	PerformanceProfileDefault PerformanceProfileType = "default"
-	// PerformanceProfileDefaultIBMZ is the default profile adjusted for IBM Z (s390x) with halved CPU requests
-	PerformanceProfileDefaultIBMZ PerformanceProfileType = "default-ibm-z"
 	// PerformanceProfileMixedWorkload is for mixed workload
 	PerformanceProfileMixedWorkload PerformanceProfileType = "mixed-workload"
 	// PerformanceProfileSmallObjects is for small objects workload
--- a/pkg/system/performance_profiles.go
+++ b/pkg/system/performance_profiles.go
@@ -1,12 +1,19 @@
 package system
 
 import (
+	"runtime"
+
 	"github.com/sirupsen/logrus"
 
 	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+const (
+	ibmZCpuArch         = "s390x"
+	ibmZCpuAdjustFactor = 0.5
 )
 
 type performanceProfile struct {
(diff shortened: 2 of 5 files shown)


pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -262,7 +262,7 @@ type NooBaaSpec struct {
 	// settings for the core, db and endpoint components.
 	// Explicit per-component resource/count fields take precedence over the profile.
 	// +optional
-	// +kubebuilder:validation:Enum=default;default-ibm-z;mixed-workload;small-objects;dev-env;mini-env
+	// +kubebuilder:validation:Enum=default;mixed-workload;small-objects;dev-env;mini-env
```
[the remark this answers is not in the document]
malayparida2000: Since the value default-ibm-z was added & this filed was made a thing there has not been any release. So there's no upgrade migration to worry about

pkg/system/performance_profiles.go
```diff
@@ -110,13 +106,23 @@ var performanceProfiles = map[nbv1.PerformanceProfileType]performanceProfile{
...
+		profile = performanceProfiles[nbv1.PerformanceProfileDefault]
 	}
-	if profileType != "" {
-		logrus.Warnf("Unknown performanceProfile %q, falling back to %q", profileType, nbv1.PerformanceProfileDefault)
+	// Halve CPU requests on IBM Z (s390x) for core/db/endpoint, matching
+	// ocs-operator's IbmZCpuAdjustFactor. See https://redhat.atlassian.net/browse/RHSTOR-9067
+	if runtime.GOARCH == ibmZCpuArch {
+		profile.coreResources = adjustCpuResourcesForIbmZ(profile.coreResources, ibmZCpuAdjustFactor)
+		profile.dbResources = adjustCpuResourcesForIbmZ(profile.dbResources, ibmZCpuAdjustFactor)
+		profile.endpointResources = adjustCpuResourcesForIbmZ(profile.endpointResources, ibmZCpuAdjustFactor)
```
dannyzaken, line 113: you skipped `pvPoolResources`

pkg/system/performance_profiles.go
```diff
@@ -147,6 +153,21 @@ func getEndpointResources(nb *nbv1.NooBaa) corev1.ResourceRequirements {
 	return lookupProfile(nb).endpointResources
 }
 
+// adjustCpuResourcesForIbmZ multiplies CPU requests by adjustFactor (limits are unchanged).
+func adjustCpuResourcesForIbmZ(rr corev1.ResourceRequirements, adjustFactor float64) corev1.ResourceRequirements {
```
dannyzaken, line 159: I suggest running this adjustment on the profile map in an init function in this file. I think it's simpler, although a bit harder to unit-test.

pkg/system/performance_profiles.go
malayparida2000: Added now.

malayparida2000:
/hold
Should go in after https://github.com/noobaa/noobaa-operator/pull/2059

pkg/system/performance_profiles.go
malayparida2000: Shifted to the init func approach now

dannyzaken (approved):

PR #2060, opened 2026-07-27
core ha | increase replica count to 2 and add configuration variables

nadavMiz:
### Describe the Problem

NooBaa core ran as a single pod. If that pod died or was restarted, management was unavailable until it came back. We want a standby core that can take over quickly.

### Explain the changes
1. Increase noobaa-core state full set replica count to have 2 pods (one active, one passive)
2. Set leader election time configuration through noobaa config-map
3. Add `spec.coreHA` on the NooBaa CR to turn core HA on/off (controls replica count: off = 1, on = 2). On by default.
4. Change pod update from `RollingUpdate` to `OnDelete`, since `RollingUpdate` can get stuck waiting for passive pod to be ready. this can cause so active pod will not get updated as well
5. Delete core pods after timing configuration flags, so that configuration will be updated on delete. and new pods will come up with correct configuration
6. Add option to the cli installation  `--core-ha` to set the `coreHA` configuration before the installation
7. Set ReleaseOnCancel to be false because of leader election module bug (`https://github.com/kubernetes/kubernetes/pull/1399`). control the release on our own once core_init is dead and renew loop stop   
8. Change operator cli to check who is the ready pod, so it can communicate with the ready pod instead of always trying for `noobaa-core-0`

### Issues: 
1. https://redhat.atlassian.net/browse/RHSTOR-9429

###Gaps
1. set pods anti-affinity #2063  

### Testing Instructions:
1. `go test ./pkg/leaderelect/ ./pkg/system/ -count=1`

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **New Features**
  * Core high availability is enabled by default with two core pods using Kubernetes lease-based leader election.
  * Added the `--disable-core-ha` option to run a single core pod.
  * Added configurable leader-election timing and shutdown grace settings.
  * External connections dynamically select an available, ready core pod.

* **Bug Fixes**
  * Core readiness now succeeds when at least one core pod is ready.
  * Improved handling of unavailable, unready, and outdated core pods.
  * Improved lease release and leadership-loss handling.

### DIFF
--- a/deploy/internal/statefulset-core.yaml
+++ b/deploy/internal/statefulset-core.yaml
@@ -5,13 +5,16 @@ metadata:
   labels:
     app: noobaa
 spec:
-  replicas: 1
+  replicas: 2
   selector:
     matchLabels:
       noobaa-core: noobaa
   serviceName: noobaa-mgmt
+  # OnDelete: RollingUpdate waits for Ready after each pod; HA standbys never
+  # become Ready, so the roll stalls after the standby and never updates the
+  # leader. With OnDelete, operator will detect template/hash drift and delete outdated pods itself.
   updateStrategy:
-    type: RollingUpdate
+    type: OnDelete
   template:
     metadata:
       labels:
@@ -60,6 +63,13 @@ spec:
           volumeMounts:
             - name: shared-bin
               mountPath: /noobaa-shared
+          resources:
+            requests:
+              cpu: "100m"
+              memory: "256Mi"
+            limits:
+              cpu: "100m"
+              memory: "256Mi"
           securityContext:
             runAsNonRoot: true
             allowPrivilegeEscalation: false
@@ -150,6 +160,31 @@ spec:
                 configMapKeyRef:
                   name: noobaa-config
                   key: NOOBAA_VERSION_AUTH_ENABLED
+            - name: NOOBAA_CORE_LEASE_DURATION
+              valueFrom:
+                configMapKeyRef:
+                  name: noobaa-config
+                  key: NOOBAA_CORE_LEASE_DURATION
+            - name: NOOBAA_CORE_RENEW_DEADLINE
+              valueFrom:
+                configMapKeyRef:
+                  name: noobaa-config
+                  key: NOOBAA_CORE_RENEW_DEADLINE
+            - name: NOOBAA_CORE_RETRY_PERIOD
+              valueFrom:
+                configMapKeyRef:
+                  name: noobaa-config
+                  key: NOOBAA_CORE_RETRY_PERIOD
+            - name: NOOBAA_CORE_SHUTDOWN_GRACE
+              valueFrom:
+                configMapKeyRef:
+                  name: noobaa-config
+                  key: NOOBAA_CORE_SHUTDOWN_GRACE
+            - name: NOOBAA_CORE_LOST_GRACE
+              valueFrom:
+                configMapKeyRef:
+                  name: noobaa-config
+                  key: NOOBAA_CORE_LOST_GRACE
             - name: NOOBAA_CORE_LEASE_NAME
               value: ""
             - name: POSTGRES_HOST
--- a/pkg/apis/noobaa/v1alpha1/noobaa_types.go
+++ b/pkg/apis/noobaa/v1alpha1/noobaa_types.go
@@ -214,6 +214,12 @@ type NooBaaSpec struct {
 	// +optional
 	Annotations AnnotationsSpec `json:"annotations,omitempty"`
 
+	// DisableCoreHA (optional) disables noobaa-core high availability (2 replicas with Kubernetes lease leader election).
+	// When false or omitted, two core pods compete for the Lease; only the leader becomes Ready.
+	// When true, a single core pod still runs leader-elect and holds the Lease.
+	// +optional
+	DisableCoreHA bool `json:"disableCoreHA,omitempty"`
+
 	// DisableLoadBalancerService (optional) sets the service type to ClusterIP instead of LoadBalancer
 	// +nullable
 	// +optional
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -1511,7 +1511,7 @@ spec:
       status: {}
 `
 
-const Sha256_deploy_crds_noobaa_io_noobaas_yaml = "88be0872b57677c437aeb656eac747cd13871d9b2733d41b0b2ec5b2d4ea7603"
+const Sha256_deploy_crds_noobaa_io_noobaas_yaml = "cf8b0283a5d8660badba7edc2cb196fb43815f7a837eab7cb86f5fbeb204e2cf"
 
 const File_deploy_crds_noobaa_io_noobaas_yaml = `---
 apiVersion: apiextensions.k8s.io/v1
@@ -3172,6 +3172,12 @@ spec:
               denyHTTP:
                 description: DenyHTTP (optional) if given will deny access to the
                   NooBaa S3 service using HTTP (only HTTPS)
+                type: boolean
+              disableCoreHA:
+                description: |-
+                  DisableCoreHA (optional) disables noobaa-core high availability (2 replicas with Kubernetes lease leader election).
+                  When false or omitted, two core pods compete for the Lease; only the leader becomes Ready.
+                  When true, a single core pod still runs leader-elect and holds the Lease.
                 type: boolean
               disableLoadBalancerService:
                 description: DisableLoadBalancerService (optional) sets the service
@@ -4895,6 +4901,16 @@ metadata:
 data: {}
 `
 
+const Sha256_deploy_internal_lease_core_yaml = "7a53fb9c5a3d886d365104e28cefac9562b02da092d8de5f84751b4e6343f734"
+
+const File_deploy_internal_lease_core_yaml = `apiVersion: coordination.k8s.io/v1
+kind: Lease
+metadata:
+  name: noobaa-core-lease
+  labels:
+    app: noobaa
+`
+
 const Sha256_deploy_internal_nsfs_pvc_cr_yaml = "6dd65ca7d324991b813f209ec6a8a6bcf6c2c9a9f45c519ad3fba51e25042f07"
 
 const File_deploy_internal_nsfs_pvc_cr_yaml = `apiVersion: v1
@@ -5623,7 +5639,7 @@ spec:
       noobaa-s3-svc: "true"
 `
 
-const Sha256_deploy_internal_statefulset_core_yaml = "8ff071dc752f7020b32222f46713887a4f9c0d2bd4d0e6d766d470f37ab5d466"
+const Sha256_deploy_internal_statefulset_core_yaml = "6d6c6f11ff0ec75ff25a1da47f32ce22af1104d7ee64d564ef1839e26954b78c"
 
 const File_deploy_internal_statefulset_core_yaml = `apiVersion: apps/v1
 kind: StatefulSet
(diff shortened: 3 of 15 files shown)


pkg/system/phase2_creating.go
```diff
@@ -1693,6 +1745,143 @@ func (r *Reconciler) SetDesiredCoreAppConfig() error {
...
+	if !util.KubeDelete(r.CoreApp) {
+		return fmt.Errorf("failed deleting StatefulSet %q to update podManagementPolicy", r.CoreApp.Name)
+	}
+	// Clear identity so CreateOrUpdate treats this as a create with the desired spec.
+	r.CoreApp.SetUID("")
+	r.CoreApp.SetResourceVersion("")
+	r.CoreApp.SetGeneration(0)
+	r.CoreApp.Spec.PodManagementPolicy = appsv1.ParallelPodManagement
+	return nil
+}
```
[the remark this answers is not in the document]
nadavMiz: this path was removed. is this still a problem?

pkg/system/phase2_creating.go
```diff
@@ -1693,6 +1745,143 @@ func (r *Reconciler) SetDesiredCoreAppConfig() error {
...
+		toDelete = &staleLeaders[0]
+	default:
+		return nil
+	}
+
+	r.Logger.Warnf("deleting outdated core pod %q (OnDelete; standby-first) desiredHash=%q updateRevision=%q holder=%q",
+		toDelete.Name, desiredHash, updateRevision, holderIdentity)
+	util.KubeDeleteNoPolling(toDelete)
+	return nil
+}
```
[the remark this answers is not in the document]
nadavMiz: this path was removed. is this still a problem?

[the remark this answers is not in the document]
nadavMiz: changing config map values is something we do only on special situations when something went wrong or for debugging. I think for these case waiting a bit for pods to be ready again is fine.

deploy/internal/statefulset-core.yaml
```diff
@@ -5,13 +5,17 @@ metadata:
...
+  podManagementPolicy: Parallel
   selector:
     matchLabels:
       noobaa-core: noobaa
   serviceName: noobaa-mgmt
+  # OnDelete: RollingUpdate stalls on never-Ready HA standbys. The operator deletes
+  # outdated pods (standby first, then leader) when the template/hash changes.
   updateStrategy:
-    type: RollingUpdate
+    type: OnDelete
```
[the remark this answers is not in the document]
nadavMiz: thats a good point. but should be done in a separate PR. can you open an issue for this?

pkg/system/reconciler.go
```diff
@@ -160,6 +162,7 @@ func NewReconciler(
 		NooBaa:                    util.KubeObject(bundle.File_deploy_crds_noobaa_io_v1alpha1_noobaa_cr_yaml).(*nbv1.NooBaa),
 		ServiceAccount:            util.KubeObject(bundle.File_deploy_service_account_yaml).(*corev1.ServiceAccount),
 		CoreApp:                   util.KubeObject(bundle.File_deploy_internal_statefulset_core_yaml).(*appsv1.StatefulSet),
+		CoreLease:                 util.KubeObject(bundle.File_deploy_internal_lease_core_yaml).(*coordinationv1.Lease),
```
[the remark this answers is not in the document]
nadavMiz: but doesn't  "k8s.io/client-go/kubernetes/scheme" already registers built in APIs like coordinationv1.Lease

pkg/apis/noobaa/v1alpha1/noobaa_types.go
```diff
@@ -101,6 +101,12 @@ type NooBaaSpec struct {
 	// +optional
 	CoreResources *corev1.ResourceRequirements `json:"coreResources,omitempty"`
 
+	// CoreHA (optional) enables noobaa-core high availability (2 replicas with Kubernetes lease leader election).
+	// When true or omitted, two core pods compete for the Lease; only the leader becomes Ready.
+	// When false, a single core pod still runs leader-elect and holds the Lease.
+	// +optional
+	CoreHA *bool `json:"coreHA,omitempty"`
```
dannyzaken: The meaning of `CoreHA` is confusing (`nil`=`true`). 
I suggest changing it to `disableCoreHA` and `nil`=`false`.

deploy/internal/statefulset-core.yaml
```diff
@@ -60,6 +63,13 @@ spec:
...
             - name: shared-bin
               mountPath: /noobaa-shared
+          resources:
+            requests:
+              cpu: "10m"
+              memory: "32Mi"
+            limits:
+              cpu: "50m"
+              memory: "64Mi"
           securityContext:
```
dannyzaken, line 73: To qualify for the `Guaranteed` Quality of Service (QoS), the values for `requests` and `limits` must be identical. Currently, we are using `Burstable` QoS for the core pod across all our performance profiles. However, we want to ensure that the initContainer does not restrict us in this aspect.

pkg/system/phase2_creating.go
```diff
@@ -203,6 +207,10 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 		return err
 	}
 
+	if err := r.ReconcileObject(r.CoreLease, r.SetDesiredCoreLease); err != nil {
```
dannyzaken, line 210: why do we need to reconcile the lease here? Isn't it created automatically by the first leader?

deploy/internal/statefulset-core.yaml
```diff
@@ -5,13 +5,16 @@ metadata:
...
   selector:
     matchLabels:
       noobaa-core: noobaa
   serviceName: noobaa-mgmt
+  # OnDelete: RollingUpdate waits for Ready after each pod; HA standbys never
+  # become Ready, so the roll stalls after the standby and never updates the
+  # leader. With OnDelete, operator will detect template/hash drift and delete outdated pods itself.
   updateStrategy:
-    type: RollingUpdate
+    type: OnDelete
```
dannyzaken, line 17: Can we avoid the `OnDelete` change by changing `spec.podManagementPolicy` to `Parallel`? I have some concerns with having to manage the pod deletions in the operator. 

The `podManagementPolicy` is immutable, so if we want to change it, we need to delete the statefulset and recreate it.

pkg/system/phase2_creating.go
```diff
@@ -1702,8 +1717,101 @@ func (r *Reconciler) SetDesiredCoreAppConfig() error {
...
+	if raw := strings.TrimSpace(r.CoreAppConfig.Data["NOOBAA_CORE_SHUTDOWN_GRACE"]); raw != "" {
+		if d, err := time.ParseDuration(raw); err == nil && d > 0 {
+			shutdownGrace = d
+		}
+	}
+	return leaderelect.RecommendedTerminationGracePeriod(shutdownGrace)
+}
 
+// restartStaleCorePods deletes one outdated core pod per reconcile under OnDelete.
+func (r *Reconciler) restartStaleCorePods() error {
```
dannyzaken, line 1757: If we can avoid the `OnDelete`, this should become unnecessary

pkg/system/phase2_creating.go
```diff
@@ -1683,6 +1693,11 @@ func (r *Reconciler) SetDesiredCoreAppConfig() error {
 		"NOOBAA_VERSION_AUTH_ENABLED":  "true",
 		"ENDPOINT_SYSTEM_STORE_SOURCE": "core", // by default, load the system store in the endpoint from the core instead of the DB
 		"OPERATOR_LOG_LEVEL":           "info",
+		"NOOBAA_CORE_LEASE_DURATION":   "20s",
+		"NOOBAA_CORE_RENEW_DEADLINE":   "10s",
+		"NOOBAA_CORE_RETRY_PERIOD":     "3s",
+		"NOOBAA_CORE_SHUTDOWN_GRACE":   "25s",
+		"NOOBAA_CORE_LOST_GRACE":       "8s",
```
dannyzaken, line 1700: not sure if these should be part of noobaa-config. We can evaluate it later

pkg/system/phase2_creating.go
nadavMiz: it is. its mostly just in case someone deletes the lease from the outside. not strictly needed

pkg/system/phase2_creating.go
nadavMiz: will leave it as part of noobaa-config for now we can move it if you think it belongs some place else

deploy/internal/statefulset-core.yaml
nadavMiz: the problem RollingUpdate is that it updates by replica order without checking which pod is ready and which isn't even if they are run on parallel. this can cause the update to stuck waiting for the standby to be ready. never updating the leader pod. for update policy, I couldn't find an update policy that updates both pods in parallel. so I did it this way.

having said that, its only for update policy. and not needed for the HA to work. I removed it, and will defer to a future PR

pkg/system/phase2_creating.go
nadavMiz: commented on related comment

deploy/internal/statefulset-core.yaml
nadavMiz: looks this problem causes tests in the CI to fail. so its needed in this PR

pkg/system/system.go
```diff
@@ -67,14 +68,19 @@ func CmdCreate() *cobra.Command {
...
+Use --disable-core-ha to install with a single core pod.`,
+		Run:  RunCreate,
+		Args: cobra.NoArgs,
 	}
 	cmd.Flags().String("core-resources", "", "Core resources JSON")
 	cmd.Flags().String("db-resources", "", "DB resources JSON")
 	cmd.Flags().String("endpoint-resources", "", "Endpoint resources JSON")
 	cmd.Flags().Bool("use-standalone-db", false, "Create NooBaa system with standalone DB (Legacy)")
 	cmd.Flags().Bool("use-obc-cleanup-policy", false, "Create NooBaa system with obc cleanup policy")
+	AddDisableCoreHAFlags(cmd)
```
jackyalbo: why not the same as the others? too long?

pkg/system/system.go
```diff
@@ -1440,6 +1454,54 @@ func Connect(isExternal bool) (*Client, error) {
...
+		}
+		if !isPodReady(pod) {
+			continue
+		}
+		ready = append(ready, pod.Name)
+	}
+	if len(ready) == 0 {
+		return "", fmt.Errorf("no Ready mgmt/core pod found for port-forward")
+	}
+	sort.Strings(ready)
```
jackyalbo: I think we should only have one like this - just return the first one

pkg/bundle/deploy.go
```diff
@@ -5687,6 +5706,13 @@ spec:
           volumeMounts:
             - name: shared-bin
               mountPath: /noobaa-shared
+          resources:
```
jackyalbo, line 5709: looks like you missed building or something  - you have different values in the statefulset-core.yaml and in deploy.go. please fix - run `make gen`

pkg/leaderelect/leaderelect_test.go
```diff
@@ -237,3 +304,8 @@ func TestRecommendedTerminationGracePeriod(t *testing.T) {
 		t.Fatalf("RecommendedTerminationGracePeriod(30s) = %d, want %d", got, want)
 	}
 }
+
+type fakeLogger struct{}
```
jackyalbo: Do you really need this?

pkg/system/phase2_creating.go
```diff
@@ -1702,8 +1717,108 @@ func (r *Reconciler) SetDesiredCoreAppConfig() error {
...
+	updateRevision := r.CoreApp.Status.UpdateRevision
+
+	podList := &corev1.PodList{}
+	if !util.KubeList(podList, client.InNamespace(r.Request.Namespace), client.MatchingLabels{"noobaa-core": r.Request.Name}) {
+		return fmt.Errorf("failed listing core pods to restart stale revisions")
+	}
+
+	for i := range podList.Items {
+		pod := &podList.Items[i]
+		// Only delete one outdated core pod at a time so HA can keep at least one core pod running to assume leadership.
```
jackyalbo, line 1795: Should we delete the standby first here and not just the first one first?

nadavMiz: added it. not it doesn't prevent a downtime. as we need to kill the leader pod, before the standby can get the lease. but it does shorten it

jackyalbo (approved):

PR #2062, opened 2026-07-29
fix: ensure NooBaaAccount credentials Secret exists before Ready

aayushchouhan09:
### Describe the Problem
NooBaaAccount can reach `phase: Ready` after a flaky `create_account` RPC even when the credentials Secret was never created, so later Secret reads fail.

### Explain the changes
1. After create/update, ensure the account credentials Secret exists (`ensureNoobaaAccountSecret` → KubeCheck, recreate via `ReadAccountAPI` if missing) before Ready.
2. Share Secret creation in `createNoobaaAccountSecret` using `KubeCreateSkipExisting`, and always pass the account ARN (including on restore) so recreated Secrets match the original contract.

Before changes:
NooBaaAccount can reach/stay phase: Ready even when the credentials Secret was never created (e.g. flaky create_account RPC). Later Secret reads fail; nb account status can show Ready with ❌ Not Found for the Secret.
Screenshot From 2026-07-30 13-31-51

After changes:
On reconcile, the operator ensures noobaa-account-<name> exists before Ready; if missing, it recreates the Secret from account credentials (or requeues on failure). 

**Note:** Secret delete alone does not trigger reconcile until the account is reconciled again (e.g. spec change).

1.  Fixed: https://redhat.atlassian.net/browse/DFBUGS-8859
2.  Related PR: https://github.com/noobaa/noobaa-core/pull/9904

### Testing Instructions:
1. Create a `NooBaaAccount` with `default_resource` set; wait until Ready and confirm Secret `noobaa-account-<name>` exists.
3. Delete that Secret, then patch the account (e.g. `force_md5_etag`) to force reconcile.
`$ kubectl patch noobaaaccount aayush --type=merge   -p '{"spec":{"force_md5_etag":true}}'`
4. Confirm the Secret is recreated and the account stays/returns Ready (nb account status shows connection info).

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved NooBaa account setup reliability by ensuring the credentials Secret is created before configuration completes.
  * Added recovery for missing credentials Secrets by retrieving account details and recreating the Secret when valid access keys are available.
  * Preserved the account ARN to support reliable credentials Secret management.

### DIFF
--- a/pkg/nb/types.go
+++ b/pkg/nb/types.go
@@ -44,6 +44,7 @@ type AccountInfo struct {
 	NextPasswordChange int64          `json:"next_password_change"`
 	DefaultResource    string         `json:"default_resource"`
 	AccessKeys         []S3AccessKeys `json:"access_keys"`
+	ARN                string         `json:"arn,omitempty"`
 	AllowedIPs         []struct {
 		Start string `json:"start"`
 		End   string `json:"end"`
--- a/pkg/noobaaaccount/reconciler.go
+++ b/pkg/noobaaaccount/reconciler.go
@@ -261,7 +261,8 @@ func (r *Reconciler) ReconcilePhaseConfiguring() error {
 		}
 	}
 
-	return nil
+	// Ensure credentials Secret exists before Ready
+	return r.ensureNoobaaAccountSecret()
 }
 
 // ReconcileDeletion handles the deletion of a noobaa account using the noobaa api
@@ -355,23 +356,50 @@ func (r *Reconciler) CreateNooBaaAccount() error {
 			return err
 		}
 		accessKeys = readAccountReply.AccessKeys[0]
+		if accountInfo.ARN == "" {
+			accountInfo.ARN = readAccountReply.ARN
+		}
 	} else {
 		accessKeys = accountInfo.AccessKeys[0]
 	}
-	r.Secret.StringData = map[string]string{}
-	r.Secret.StringData["AWS_ACCESS_KEY_ID"] = string(accessKeys.AccessKey)
-	r.Secret.StringData["AWS_SECRET_ACCESS_KEY"] = string(accessKeys.SecretKey)
-	r.Secret.StringData["ARN"] = string(accountInfo.ARN)
 
-	r.Own(r.Secret)
-	err = r.Client.Create(r.Ctx, r.Secret)
-	if err != nil {
-		r.Logger.Errorf("got error on NooBaaAccount creation. error: %v", err)
+	if err := r.createNoobaaAccountSecret(accessKeys, accountInfo.ARN); err != nil {
 		return err
 	}
 
 	log.Infof("✅ Successfully created account %q", r.NooBaaAccount.Name)
 	return nil
+}
+
+// createNoobaaAccountSecret creates the credentials Secret for the NooBaaAccount
+func (r *Reconciler) createNoobaaAccountSecret(accessKeys nb.S3AccessKeys, arn string) error {
+	r.Secret.StringData = map[string]string{
+		"AWS_ACCESS_KEY_ID":     string(accessKeys.AccessKey),
+		"AWS_SECRET_ACCESS_KEY": string(accessKeys.SecretKey),
+		"ARN":                   arn,
+	}
+	r.Own(r.Secret)
+	if !util.KubeCreateSkipExisting(r.Secret) {
+		return fmt.Errorf("Failed to create NoobaaAccount secret %q", r.Secret.Name)
+	}
+	return nil
+}
+
+// ensureNoobaaAccountSecret ensures the credentials Secret exists for the NooBaaAccount
+func (r *Reconciler) ensureNoobaaAccountSecret() error {
+	if util.KubeCheck(r.Secret) {
+		return nil
+	}
+
+	r.Logger.Warnf("⏳ Secret %q is missing, recreating from account credentials", r.Secret.Name)
+	accountInfo, err := r.NBClient.ReadAccountAPI(nb.ReadAccountParams{Email: r.NooBaaAccount.Name})
+	if err != nil {
+		return err
+	}
+	if len(accountInfo.AccessKeys) == 0 {
+		return fmt.Errorf("NooBaaAccount %q has no access keys", r.NooBaaAccount.Name)
+	}
+	return r.createNoobaaAccountSecret(accountInfo.AccessKeys[0], accountInfo.ARN)
 }
 
 // UpdateNooBaaAccount update an existing noobaa account


pkg/noobaaaccount/reconciler.go
```diff
@@ -358,22 +359,46 @@ func (r *Reconciler) CreateNooBaaAccount() error {
...
+// createAccountSecret creates the credentials Secret for the NooBaaAccount
+func (r *Reconciler) createAccountSecret(accessKeys nb.S3AccessKeys, arn string) error {
+	r.Secret.StringData = map[string]string{
+		"AWS_ACCESS_KEY_ID":     string(accessKeys.AccessKey),
+		"AWS_SECRET_ACCESS_KEY": string(accessKeys.SecretKey),
+		"ARN":                   arn,
+	}
+	r.Own(r.Secret)
+	if !util.KubeCreateSkipExisting(r.Secret) {
+		return fmt.Errorf("Failed to create secret %q", r.Secret.Name)
```
shirady: In the previous log message you had:
"got error on NooBaaAccount creation. error:"
and now it is about a general secret, was it intentional?
```suggestion
		return fmt.Errorf("Failed to create NoobaaAccount secret %q", r.Secret.Name)
```

pkg/noobaaaccount/reconciler.go
```diff
@@ -358,22 +359,46 @@ func (r *Reconciler) CreateNooBaaAccount() error {
...
+	if err := r.createAccountSecret(accessKeys, string(accountInfo.ARN)); err != nil {
 		return err
 	}
 
 	log.Infof("✅ Successfully created account %q", r.NooBaaAccount.Name)
 	return nil
 }
 
+// createAccountSecret creates the credentials Secret for the NooBaaAccount
+func (r *Reconciler) createAccountSecret(accessKeys nb.S3AccessKeys, arn string) error {
```
shirady: Maybe worth keeping the term `NoobaaAccount` in the function as well
```suggestion
func (r *Reconciler) createNoobaaAccountSecret(accessKeys nb.S3AccessKeys, arn string) error {
```

pkg/noobaaaccount/reconciler.go
```diff
@@ -358,22 +359,46 @@ func (r *Reconciler) CreateNooBaaAccount() error {
...
+	}
+	r.Own(r.Secret)
+	if !util.KubeCreateSkipExisting(r.Secret) {
+		return fmt.Errorf("Failed to create secret %q", r.Secret.Name)
+	}
+	return nil
+}
+
+// ensureAccountSecret ensures the credentials Secret exists for the account
+func (r *Reconciler) ensureAccountSecret() error {
```
shirady: Same here - 
Maybe worth keeping the term NoobaaAccount in the function as well
```suggestion
func (r *Reconciler) ensureNoobaaAccountSecret() error {
```

pkg/noobaaaccount/reconciler.go
```diff
@@ -358,22 +359,46 @@ func (r *Reconciler) CreateNooBaaAccount() error {
...
+
+	r.Logger.Warnf("⏳ Secret %q is missing, recreating from account credentials", r.Secret.Name)
+	accountInfo, err := r.NBClient.ReadAccountAPI(nb.ReadAccountParams{Email: r.NooBaaAccount.Name})
+	if err != nil {
+		return err
+	}
+	if len(accountInfo.AccessKeys) == 0 {
+		return fmt.Errorf("NooBaaAccount %q has no access keys", r.NooBaaAccount.Name)
+	}
+	return r.createAccountSecret(accountInfo.AccessKeys[0], "")
```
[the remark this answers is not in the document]
shirady: @aayushchouhan09 I agree with this comment.

shirady:
@aayushchouhan09 were you able to reproduce and test it?
If so, can you share the steps that you did?

aayushchouhan09:
> @aayushchouhan09 were you able to reproduce and test it? If so, can you share the steps that you did?

Updated the PR description. Thanks!

pkg/nb/types.go
```diff
@@ -44,6 +44,7 @@ type AccountInfo struct {
 	NextPasswordChange int64          `json:"next_password_change"`
 	DefaultResource    string         `json:"default_resource"`
 	AccessKeys         []S3AccessKeys `json:"access_keys"`
+	ARN                string         `json:"arn,omitempty"`
```
shirady, line 47: Do you have this ARN in the account info from noobaa-core?
I'm not sure you get it from `ReadAccountAPI`.

aayushchouhan09: You’re right, `ReadAccountAPI` does not return ARN (only `create_account` does). We no longer rely on it for ARN. On create we persist `CreateAccountReply.ARN` on the NooBaaAccount annotation `noobaa.io/arn`, and on Secret recreate we take keys from `ReadAccountAPI` and ARN from that annotation - same approach OBC already uses with `AdditionalState["arn"]`.

aayushchouhan09:
Question: the controller only watches NooBaaAccount with GenerationChangedPredicate, so deleting the credentials Secret alone does not trigger reconcile. Until the NooBaaAccount is reconciled again (e.g. a spec patch), phase stays Ready even though the Secret is missing — it is not updated to reflect that. The Secret is then recreated on that next reconcile. That was intentional to keep this fix minimal — should we also watch the owned Secret (and/or surface a non-Ready phase while it is missing), or is reconcile-on-spec-change with phase staying Ready enough?

@naveenpaul1 @shirady @karthikps97 wdyt?

shirady:
> Question: the controller only watches NooBaaAccount with GenerationChangedPredicate, so deleting the credentials Secret alone does not trigger reconcile. Until the NooBaaAccount is reconciled again (e.g. a spec patch), phase stays Ready even though the Secret is missing — it is not updated to reflect that. The Secret is then recreated on that next reconcile. That was intentional to keep this fix minimal — should we also watch the owned Secret (and/or surface a non-Ready phase while it is missing), or is reconcile-on-spec-change with phase staying Ready enough?
> 
> @naveenpaul1 @shirady @karthikps97 wdyt?

When I saw your reproduction steps I also thought about it also.
I agree with your approach, you can added in in a follow-up PR.

naveenpaul1:
@aayushchouhan09 @shirady Do we really want to watch for the secret? While this is a valid issue caused by API-side delays, a deleted secret is typically the result of human intervention. If we begin tracking all such resources, it will introduce excessive watch conditions into our code.

aayushchouhan09:
> @aayushchouhan09 @shirady Do we really want to watch for the secret? While this is a valid issue caused by API-side delays, a deleted secret is typically the result of human intervention. If we begin tracking all such resources, it will introduce excessive watch conditions into our code.

We don’t need to watch the `Secret` or clear `Ready` when it’s missing. `Secret` deletion is usually human intervention, and watching owned `Secrets` would add unnecessary watch overhead. The real issue (`DFBUGS-8859`) is that the `Secret` may never be created after a flaky `create_account`, while the account exists in core and the CR is `Ready`. This fix recreates the `Secret` on the next reconcile (before staying `Ready`); if ensure fails, reconcile errors and `Ready` is not set. That keeps the fix minimal without extra watches or phase issue.

@naveenpaul1 @shirady  
cc @liranmauda

pkg/nb/types.go
shirady: Did we have the annotation before or is it something that you're suggesting to add?

aayushchouhan09: It’s new in this PR - we didn’t have `noobaa.io/arn` on `NooBaaAccount` before. We added it to persist the account identity ARN from `CreateAccountReply` on create, and reuse it when recreating the credentials Secret, since `ReadAccountAPI` does not return ARN.

This is the same as ARN OBC already stores - OBC puts it in `AdditionalState["arn"]` because `ObjectBucket` already has that map. `NooBaaAccount` has no equivalent field, so we use an annotation to avoid a CRD/API change.

naveenpaul1 (approved):
LGTM

pkg/nb/types.go
aayushchouhan09: Added support for `ARN` in `ReadAccountAPI` reply, PTAL: [#9904](https://github.com/noobaa/noobaa-core/pull/9904)

Also updated the code as per the changes.

shirady (approved):
LGTM

shirady:
> > @aayushchouhan09 @shirady Do we really want to watch for the secret? While this is a valid issue caused by API-side delays, a deleted secret is typically the result of human intervention. If we begin tracking all such resources, it will introduce excessive watch conditions into our code.
> 
> We don’t need to watch the `Secret` or clear `Ready` when it’s missing. `Secret` deletion is usually human intervention, and watching owned `Secrets` would add unnecessary watch overhead. The real issue (`DFBUGS-8859`) is that the `Secret` may never be created after a flaky `create_account`, while the account exists in core and the CR is `Ready`. This fix recreates the `Secret` on the next reconcile (before staying `Ready`); if ensure fails, reconcile errors and `Ready` is not set. That keeps the fix minimal without extra watches or phase issue.
> 
> @naveenpaul1 @shirady cc @liranmauda

I still think we have a gap in the reconcile, you can open an issue about it.

PR #2064, opened 2026-07-30
Decrease adjust factor for IBM Z CPU to 0.2

malayparida2000:
Lower the s390x CPU request adjust factor so NooBaa/ODF can fit typical IBM Z footprints (e.g. 4 CPUs / 2 IFLs per node). This reflects higher per-core capability on Z rather than expecting pods to scale up from the reduced requests; limits are unchanged.

The Z team will run functional and performance validation on this build, and we can revisit the factor if needed based on results.

Ref-https://redhat.atlassian.net/browse/RHSTOR-9526

## Bug Fixes

* Improved CPU request scaling for performance-profile-managed components on IBM Z systems.
* Reduced the CPU adjustment factor used during initialization to better align requested resources with expected requirements.
* CPU limits and resource handling for other architectures remain unchanged.

### DIFF
--- a/pkg/system/performance_profiles.go
+++ b/pkg/system/performance_profiles.go
@@ -13,7 +13,7 @@ import (
 
 const (
 	ibmZCpuArch         = "s390x"
-	ibmZCpuAdjustFactor = 0.5
+	ibmZCpuAdjustFactor = 0.2
 )
 
 type performanceProfile struct {


dannyzaken (changes requested):
@malayparida2000, I must push back on this change. 
Were any tests performed specifically to test Noobaa that can support this scaling factor? Can you refer me to actual numbers?

malayparida2000:
> @malayparida2000, I must push back on this change. Were any tests performed specifically to test Noobaa that can support this scaling factor? Can you refer me to actual numbers?

I understand your concern — my initial description was too light on context.

This change is not about expecting pods to "scale up" from a low CPU request toward the unadjusted values. The adjust factor exists because IBM Z cores are substantially more capable than typical x86 cores (larger caches, higher memory bandwidth per core, higher IPC, aggressive branch prediction, SMT, etc.). That was the original rationale for 0.5, and it is the same rationale for moving to 0.2: we expect the adjusted request to better match average CPU consumption on Z, not to act as a temporary floor that workloads grow out of.

On the testing point: I don't yet have Noobaa-specific perf numbers I can attach here. I'll follow up with the Z team and share if they have any past numbers.

The immediate constraint is installability. IBM Z environments commonly start with a small CPU footprint (default target here: 4 CPUs / 2 IFLs per node), with vertical scaling being relatively easy later. With current requests, ODF simply does not fit that footprint. So the plan is:

1. We first lower CPU requests so the product can install on the IBM Z env
2. Let the Z team run performance validation on that build
3. Make adjustments if those results show regressions or undersizing

malayparida2000:
> > @malayparida2000, I must push back on this change. Were any tests performed specifically to test Noobaa that can support this scaling factor? Can you refer me to actual numbers?
> 
> I understand your concern — my initial description was too light on context.
> 
> This change is not about expecting pods to "scale up" from a low CPU request toward the unadjusted values. The adjust factor exists because IBM Z cores are substantially more capable than typical x86 cores (larger caches, higher memory bandwidth per core, higher IPC, aggressive branch prediction, SMT, etc.). That was the original rationale for 0.5, and it is the same rationale for moving to 0.2: we expect the adjusted request to better match average CPU consumption on Z, not to act as a temporary floor that workloads grow out of.
> 
> On the testing point: I don't yet have Noobaa-specific perf numbers I can attach here. I'll follow up with the Z team and share if they have any past numbers.
> 
> The immediate constraint is installability. IBM Z environments commonly start with a small CPU footprint (default target here: 4 CPUs / 2 IFLs per node), with vertical scaling being relatively easy later. With current requests, ODF simply does not fit that footprint. So the plan is:
> 
> 1. We first lower CPU requests so the product can install on the IBM Z env
> 2. Let the Z team run performance validation on that build
> 3. Make adjustments if those results show regressions or undersizing

@dannyzaken ^^

dannyzaken (approved):

PR #2065, opened 2026-07-30
small-objects perf profile - endpoint CPU request 1 + HPA scaling beh…

dannyzaken:
### Describe the Problem

The `small-objects` performance profile sets the endpoint CPU request to 2 cores.
But a single-process endpoint serving small objects is main-thread bound and tops
out at ~1 core, so a fully saturated pod reports only ~50% CPU utilization. With the
HPA target at 80%, the autoscaler never fires — the profile cannot scale on small
objects even under full load. Measurements across 4KiB–128KiB confirm the pod sits
at 50–70% utilization while saturated (a mixed 46KiB stream reads 53–55%).

Separately, the endpoint HPA has no `behavior` block, so scale-up/down timing falls
back to Kubernetes defaults.

### Explain the Changes
1. `small-objects` endpoint CPU **request 2 → 1** (limit unchanged at 4, memory
   unchanged). A saturated small-object pod now reports ~100% utilization, restoring
   the CPU-based HPA trigger. `mixed-workload` keeps request 2 because its per-pod
   ceiling is ~2.5 cores — it already fires.
2. Add an HPA `behavior` block to the endpoint HPA (global, applies to whichever
   profile is active): scale up promptly on sustained load (60s window, ≤2 pods/60s),
   scale down slowly (600s window, 1 pod/180s) to hold endpoints through transient
   breaks and avoid the replica flapping that the more sensitive request value can cause.
3. Update the profile unit test expectation for the `small-objects` endpoint resources.
4. Regenerate `pkg/bundle/deploy.go`.

1. [RHSTOR-8629](https://redhat.atlassian.net/browse/RHSTOR-8629) (MCG performance profiles) 

### Testing Instructions:
1. Deploy the operator with `spec.performanceProfile: small-objects`; confirm the
   `noobaa-endpoint` deployment has CPU `requests: 1` / `limits: 4`.
2. `oc get hpa noobaa-hpav2 -o yaml` — confirm the `spec.behavior` block is present.
3. Run a small-object workload against 2 endpoints (e.g. `warp mixed --obj.size 4KiB`);
   confirm a saturated pod at ~1 core now reads ~100% HPA utilization, and the HPA
   scales up (it read ~50% and never scaled before).
4. Confirm `mixed-workload` still has endpoint `requests: 2` and fires on ≥1MiB traffic.
5. `go test ./pkg/system/ -run TestProfile`.

- [x] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

* **Performance Improvements**
  * Reduced the CPU request for small-object endpoint workloads while keeping existing CPU limits and memory settings unchanged.
* **Autoscaling**
  * Updated HPA behavior to improve replica stability by tuning scale-up/scale-down stabilization windows and limiting replica change rates.
* **Documentation**
  * Added expanded performance measurement results and guidance explaining endpoint scaling behavior and the recommended profile settings.
* **Tests**
  * Updated profile resource expectations to reflect the revised small-object CPU request.

### DIFF
--- a/deploy/internal/hpav2-autoscaling.yaml
+++ b/deploy/internal/hpav2-autoscaling.yaml
@@ -9,9 +9,30 @@ spec:
     kind: Deployment
     name: noobaa-endpoint
   metrics:
-  - type: Resource
-    resource:
-      name: cpu
-      target:
-        type: Utilization
-        averageUtilization: 80
+    - type: Resource
+      resource:
+        name: cpu
+        target:
+          type: Utilization
+          averageUtilization: 80
+  # reduce replica flapping. The small objects profile endpoint CPU request is 1 core,
+  # so a saturated pod reaches ~100% utilisation and a mixed small/large object stream can push
+  # utilisation across the trigger repeatedly. Scale up gradually (2 pods every 60 seconds),
+  # but hold added endpoints through transient pauses before removing them to avoid endpoint
+  # pod churn which is expensive.
+  # These are conservative starting points, not yet validated under live burst traffic.
+  # the behavior defined here only applies to the first creation of the HPA object,
+  # If changed, it should be updated in the reconcileAdapterHPA 
+  behavior:
+    scaleUp:
+      stabilizationWindowSeconds: 60
+      policies:
+        - type: Pods
+          value: 2
+          periodSeconds: 60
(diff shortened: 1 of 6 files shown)


liranmauda (approved):
LGTM

PR #2076, opened 2026-08-05
CLI command to update endpoints on stores

karthikps97:
### Description
Introduce a new CLI command to update endpoint for select store types (IBM COS and S3 compatible). The CLI updates both backingstores and namespace stores.

Working:
1. Checks for all stores that have their endpoint set to `old-endpoint` parameter.
2. Fetch each store's secret and bucket and call `check_external_connection` to validate the new endpoint. noobaa-core checks for presence of `noobaa_blocks` key prefix for backingstores, and just a basic connection test for namespace stores.
3. Add an annotation on the store's. Reconciler now skips the stores with this annotation.
4. Update the endpoint field in spec for each store. (rollback on error)
5. Call `update_external_connection` for each unique connection identified which uses the old endpoint. (rollback on failure)
6. Remove the set annotations from each store.

### Explain the changes
1. Added a new package `endpoint` for the new CLI.
2. Inroduced a new annotation `noobaa.io/pause-reconcile`.
3. Added checks in backingstore and namespace store reconciler's to skip on presence of `pause-reconcile` annotation.

1. 

### Testing Instructions:
1. Create backing and namespace stores.
4. Execute `nb connection update --old-endpoint <old-endpoint> --new-endpoint <new-endpoint>`
5. Once updated check for validity of store with updated endpoint.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added a `connection update` command to replace endpoints for supported object stores.
  * Added endpoint validation, credential handling, rollback protection, and reconciliation pause support during updates.
  * Added optional bucket and endpoint details for connection checks and updates.
  * Added automatic registration of the connection command in the CLI’s Manage group.

* **Bug Fixes**
  * Resources marked to pause reconciliation are now skipped and retried automatically.

* **Documentation**
  * Added usage, prerequisites, workflows, rollback details, annotation guidance, and troubleshooting for connection updates.

### DIFF
--- a/pkg/connection/connection.go
+++ b/pkg/connection/connection.go
@@ -1,0 +1,577 @@
+package connection
+
+import (
+	"fmt"
+	"strings"
+
+	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	"github.com/noobaa/noobaa-operator/v5/pkg/constants"
+	"github.com/noobaa/noobaa-operator/v5/pkg/nb"
+	"github.com/noobaa/noobaa-operator/v5/pkg/options"
+	"github.com/noobaa/noobaa-operator/v5/pkg/system"
+	"github.com/noobaa/noobaa-operator/v5/pkg/util"
+
+	"github.com/spf13/cobra"
+	corev1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type StoreType string
+
+const (
+	StoreTypeS3Compatible StoreType = "s3-compatible"
+	StoreTypeIBMCOS       StoreType = "ibm-cos"
+)
+
+type matchedStore struct {
+	isBackingStore bool
+	Store          client.Object
+	identity       string
+	secret         string
+	authMethod     nb.CloudAuthMethod
+	bucket         string
+	endpointType   nb.EndpointType
+}
+
+func (m *matchedStore) name() string {
+	if m.Store == nil {
+		return ""
+	}
+	name := m.Store.GetName()
+	if m.isBackingStore {
+		return "BackingStore/" + name
+	}
+	return "NamespaceStore/" + name
+}
+
+type connEntry struct {
+	name         string
+	endpointType nb.EndpointType
+}
+
+// Cmd returns a CLI command
+func Cmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "connection",
+		Short: "Manage backingstores/namespacestores connections",
+	}
+	cmd.AddCommand(
+		CmdUpdate(),
+	)
+	return cmd
+}
+
+// CmdUpdate returns a CLI command
+func CmdUpdate() *cobra.Command {
+	cmd := &cobra.Command{
+		Use: "update",
+		Short: `Update connection properties across all matching backing stores and namespace stores, 
+		currently the only supported property is endpoint`,
+		Run: RunUpdate,
+	}
+	cmd.Flags().String("old-endpoint", "", "The current endpoint URL to replace")
+	cmd.Flags().String("new-endpoint", "", "The new endpoint URL to set")
+	return cmd
+}
+
+// RunUpdate implements the endpoint update flow
+func RunUpdate(cmd *cobra.Command, args []string) {
+	log := util.Logger()
+	oldEndpoint, _ := cmd.Flags().GetString("old-endpoint")
+	newEndpoint, _ := cmd.Flags().GetString("new-endpoint")
+	oldEndpoint = strings.TrimSpace(oldEndpoint)
+	newEndpoint = strings.TrimSpace(newEndpoint)
+
+	// Validate flags
+	if oldEndpoint == "" || newEndpoint == "" {
+		log.Fatalf("both --old-endpoint and --new-endpoint are required")
+	}
+
+	// List and filter all matching stores using oldEndpoint
+	matched := findMatchingStores(oldEndpoint)
+	if len(matched) == 0 {
+		log.Fatalf("no matching stores found for endpoint %q", oldEndpoint)
+	}
+
+	log.Infof("Found %d store(s) matching endpoint %q:", len(matched), oldEndpoint)
+	for _, m := range matched {
+		log.Infof("  - %s", m.name())
+	}
+
+	// Read secrets and buckets
+	if err := loadSecretsAndBuckets(matched); err != nil {
+		log.Fatalf("failed to read secrets: %s", err)
+	}
+
+	// Pre-validate each store's bucket on the new endpoint
+	sysClient, err := system.ConnectAuto()
+	if err != nil {
+		log.Fatalf("failed to connect to NooBaa system: %s", err)
+	}
+	nbClient := sysClient.NBClient
+
+	failedValidations := validateStores(nbClient, matched, newEndpoint)
+	if len(failedValidations) > 0 {
+		log.Errorf("Pre-validation failed for the following stores:")
+		for _, f := range failedValidations {
+			log.Errorf("  - %s", f)
+		}
+		log.Fatalf("Aborting. No changes have been made.")
+	}
+	log.Infof("All stores passed pre-validation on new endpoint %q", newEndpoint)
+
+	// Set pause annotation on all matching stores
+	if errs := setPauseAnnotation(matched, true); len(errs) > 0 {
+		removePauseAnnotations(matched)
+		log.Fatalf("failed to set pause annotation: %s", errs)
+	}
+	log.Infof("Paused reconciliation for all matching stores")
+
+	// Patch all CR specs with new endpoint
+	var patched []matchedStore
+	if err := patchEndpoints(matched, newEndpoint, &patched); err != nil {
+		log.Errorf("failed to patch endpoints: %s", err)
+		rollback(patched, oldEndpoint, matched, nil, nil)
+		log.Fatalf("Rollback complete. Endpoint update aborted.")
+	}
+	log.Infof("Patched all %d store specs with new endpoint", len(patched))
+
+	// Discover unique connections by name from core
+	uniqueConns, err := findConnections(nbClient, oldEndpoint)
+	if err != nil {
+		log.Errorf("failed to find connections: %s", err)
+		rollback(patched, oldEndpoint, matched, nil, nil)
+		log.Fatalf("Rollback complete. Endpoint update aborted.")
+	}
+	log.Infof("Found %d unique connection(s) to update in NooBaa core", len(uniqueConns))
+
+	// Update external connections, tracking successes for rollback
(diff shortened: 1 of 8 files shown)


pkg/endpoint/endpoint.go
```diff
@@ -0,0 +1,511 @@
+package endpoint
```
romayalon: endpoint CLI sounds like it's related to the noobaa-endpoint and not a backingstore's/namespacestore's connection
why not add this in system/connection/backingstore's/namespacestore's CLI?
I would add a new CLI and not use an existing one only if we think we will add more options to it in the future

pkg/endpoint/endpoint.go
```diff
@@ -0,0 +1,511 @@
...
+type connEntry struct {
+	name         string
+	endpointType nb.EndpointType
+}
+
+// Cmd returns a CLI command
+func Cmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "endpoint",
+		Short: "Manage endpoints",
```
romayalon: it sounds like it manages noobaa endpoints

pkg/connection/connection.go
```diff
@@ -0,0 +1,511 @@
...
+	// Set pause annotation on all matching stores
+	if err := setPauseAnnotation(matched, true); err != nil {
+		removePauseAnnotations(matched)
+		log.Fatalf("failed to set pause annotation: %s", err)
+	}
+	log.Infof("Paused reconciliation for all matching stores")
+
+	// Patch all CR specs with new endpoint
+	var patched []matchedStore
+	if err := patchEndpoints(matched, newEndpoint, &patched); err != nil {
```
romayalon, line 132: we should be more flexible here to allow different connection properties updates in the future

pkg/connection/connection.go
```diff
@@ -0,0 +1,511 @@
...
+		Short: "Update endpoint URL across all matching backing stores and namespace stores",
+		Run:   RunUpdate,
+	}
+	cmd.Flags().String("old-endpoint", "", "The current endpoint URL to replace")
+	cmd.Flags().String("new-endpoint", "", "The new endpoint URL to set")
+	return cmd
+}
+
+// RunUpdate implements the endpoint update flow
+func RunUpdate(cmd *cobra.Command, args []string) {
```
romayalon, line 78: I would split to functions

pkg/endpoint/endpoint.go
```diff
@@ -0,0 +1,511 @@
...
+				m.authMethod = "AWS_V2"
+			}
+		}
+	}
+}
+
+func setPauseAnnotation(stores []matchedStore, pause bool) (errors []error) {
+	for i := range stores {
+		m := &stores[i]
+		if m.isBackingStore {
```
romayalon: why do you care if it's backingstore/namespacestore
you can just call it store instead of having the same code twice

pkg/endpoint/endpoint.go
```diff
@@ -0,0 +1,511 @@
...
+
+	bucket, err := util.GetNamespaceStoreTargetBucket(ns)
+	if err != nil {
+		return fmt.Errorf("NamespaceStore %q: %w", ns.Name, err)
+	}
+	m.bucket = bucket
+	return nil
+}
+
+func extractNamespaceStoreCredentials(m *matchedStore, ns *nbv1.NamespaceStore, secret *corev1.Secret) {
```
romayalon: seems like we have double functions for namespacestore/backingstore - 
I would try to think if I can merge them to a single logic per function

pkg/nb/types.go
```diff
@@ -748,6 +748,7 @@ type CheckExternalConnectionParams struct {
 	AzureLogAccessKeys     *AzureLogAccessKeysParams `json:"azure_log_access_keys,omitempty"`
 	Region                 string                    `json:"region,omitempty"`
 	AzureSTSCredentials    *AzureSTSCredentials      `json:"azure_sts_credentials,omitempty"`
+	Bucket                 string                    `json:"bucket,omitempty"`
```
romayalon, line 752: currently in noobaa-core check_external_connection does list_buckets, are you planning to change it? also if you want to change/add new parameters they nust be changed in noobaa-core

pkg/nb/types.go
```diff
@@ -759,11 +760,18 @@ type CheckExternalConnectionReply struct {
...
+	Endpoint     string       `json:"endpoint,omitempty"`
+	EndpointType EndpointType `json:"endpoint_type,omitempty"`
+}
+
 // UpdateExternalConnectionParams is the params of account_api.update_external_connection()
 type UpdateExternalConnectionParams struct {
 	Name               string                    `json:"name"`
-	Identity           MaskedString              `json:"identity"`
-	Secret             MaskedString              `json:"secret"`
+	EndpointInfo       *EndpointInfo             `json:"endpoint_info,omitempty"`
```
romayalon, line 773: why this change is needed?

pkg/endpoint/endpoint.go
karthikps97: I can't add this under backingstore or namespace store since it acts on both.
I didn't find anything for connection.
For system we can, but it looks more like for managing noobaa CR while this CLI is just for stores.
We can add more options like updating credentials in future. It would work without any additional changes on core.

pkg/connection/connection.go
karthikps97: Split into few more functions.

pkg/endpoint/endpoint.go
karthikps97: Updated the command description

pkg/connection/connection.go
karthikps97: We can add support for secrets with minimal changes. But we dont have any other use case as of now.

pkg/endpoint/endpoint.go
karthikps97: Yes, I refactored using interface and type checks.

pkg/endpoint/endpoint.go
karthikps97: Merged into single function, but we still need 2 blocks with type checks.

pkg/nb/types.go
karthikps97: Yes, we have done changes for core too. [PR](https://github.com/noobaa/noobaa-core/pull/9701)
We check for presence of `noobaa_storage` prefix for backingstores

pkg/nb/types.go
karthikps97: To align with core changes in which we accept endpoint and its type.

pkg/connection/connection.go
```diff
@@ -0,0 +1,576 @@
...
+			params.Bucket = m.bucket
+		}
+		reply, err := nbClient.CheckExternalConnectionAPI(params)
+		if err != nil {
+			failedValidations = append(failedValidations, fmt.Sprintf("%s: %v", m.name(), err))
+			continue
+		}
+		if reply.Status != nb.ExternalConnectionSuccess {
+			failedValidations = append(failedValidations, fmt.Sprintf("%s (bucket %q): status=%s error=%s",
+				m.name(), m.bucket, reply.Status, reply.Error.Message))
```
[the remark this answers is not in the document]
karthikps97: We dont need bucket for namespace stores. Its by design that we dont pass them in API call.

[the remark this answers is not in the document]
karthikps97: We might not need that too. We identify connections based on the endpoint, secret.

pkg/endpoint/endpoint.go
romayalon: I wouldn't call this CLI endpoint as it sounds like it should handle noobaa-endpoint and not an endpoint property in connection

pkg/connection/connection.go
```diff
@@ -0,0 +1,576 @@
...
+		storeName := m.name()
+		annotations := store.GetAnnotations()
+		if annotations == nil {
+			annotations = map[string]string{}
+		}
+
+		if pause {
+			annotations[constants.PauseReconcile] = "true"
+		} else {
+			delete(annotations, constants.PauseReconcile)
```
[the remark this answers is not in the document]
karthikps97: This annotation is not used by any other controller. We preserve all other annotations that are currently set on the store.

pkg/nb/types.go
romayalon: good, you should also update the core PR on options.go so the api versions will match

pkg/connection/connection.go
```diff
@@ -0,0 +1,576 @@
...
+type connEntry struct {
+	name         string
+	endpointType nb.EndpointType
+}
+
+// Cmd returns a CLI command
+func Cmd() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "connection",
+		Short: "Manage store connections",
```
romayalon:
```suggestion
		Short: "Manage backingstores/namespacestores connections",
```

pkg/connection/connection.go
```diff
@@ -0,0 +1,576 @@
...
+		CmdUpdate(),
+	)
+	return cmd
+}
+
+// CmdUpdate returns a CLI command
+func CmdUpdate() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "update",
+		Short: "Update connection endpoint across all matching backing stores and namespace stores",
```
romayalon: we want the update command to be generic
```suggestion
		Short: "Update connection properties across all matching backing stores and namespace stores, currently the only supported property is endpoint",
```

pkg/connection/connection.go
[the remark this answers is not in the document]
karthikps97: In case of a failed update earlier, we still set the annotation, and on successful completion remove it. It's the same annotation and will have the same value `true`. Also, the user of the CLI would need to remember not to invoke the cli multiple times concurrently. So this problem wouldnt arise.

pkg/connection/connection.go
```diff
@@ -0,0 +1,576 @@
...
+	var failedValidations []string
+	for _, m := range matched {
+		params := nb.CheckExternalConnectionParams{
+			Endpoint:     newEndpoint,
+			EndpointType: m.endpointType,
+			Identity:     nb.MaskedString(m.identity),
+			Secret:       nb.MaskedString(m.secret),
+			AuthMethod:   m.authMethod,
+		}
+		if m.isBackingStore {
```
romayalon, line 229: why do you set that only for backingstore?

romayalon (approved):
Approving to not block you
I think we should have a follow up refactoring PR for reducing the generated duplicated code

pkg/connection/connection.go
karthikps97: For backingstore, we do an additional check in `check_external_connection` API to find if the bucket contains the `noobaa_storage/` key prefix in it and only then proceed with the update. This is to ensure that the store's bucket should be reachable via the new endpoint.

PR #2079, opened 2026-08-09
Core HA | add anti-affinity to core pods so they will run on different nodes

nadavMiz:
### Describe the Problem
Core HA runs two core pods for failover, but both can be scheduled on the same node. If that node dies, both go down and failover doesn’t help.

### Explain the changes
1. When Core HA is on, ask the scheduler to prefer putting the two core pods on different nodes.
2. If the user sets a topology key (e.g. zone), also prefer spreading across that topology.
3. Keep any affinity the user already configured; don’t change behavior when HA is off.

### Testing Instructions:
1. `go test ./pkg/system/ -count=1 -run 'TestDesiredCoreAffinity|TestGetDesiredCoreReplicas'`

- [ ] Doc added/updated
- [x] Tests added

## Bug Fixes

* Improved high-availability scheduling by preferring placement across different hostnames and custom topology zones.
* Preserved existing node, pod, and anti-affinity settings, including user-defined preferences.
* Prevented duplicate scheduling rules while retaining configurations safely without unintended changes.
* Ensured high-availability settings remain unchanged when the feature is disabled.
* Improved scheduling behavior when custom topology rules overlap with automatically generated preferences.

### DIFF
--- a/pkg/system/core_ha_test.go
+++ b/pkg/system/core_ha_test.go
@@ -8,6 +8,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 func TestGetDesiredCoreReplicas(t *testing.T) {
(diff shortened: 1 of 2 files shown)


pkg/system/core_ha_test.go
```diff
@@ -186,3 +186,188 @@ func TestPickStaleCorePodToDelete(t *testing.T) {
...
 	}
 }
+
+func TestDesiredCoreAffinity(t *testing.T) {
+	t.Parallel()
+
+	const coreName = "noobaa"
+	zoneKey := "topology.kubernetes.io/zone"
+
+	t.Run("ha off no affinity", func(t *testing.T) {
```
jackyalbo: Can you use capital HA? ha sounds less clear - to me at least...

pkg/system/phase2_creating.go
```diff
@@ -843,7 +843,7 @@ func (r *Reconciler) SetDesiredCoreApp() error {
 			[]corev1.LocalObjectReference{*r.NooBaa.Spec.ImagePullSecret}
 	}
 	podSpec.Tolerations = r.NooBaa.Spec.Tolerations
-	podSpec.Affinity = r.GetAffinity()
+	podSpec.Affinity = desiredCoreAffinity(r.NooBaa, r.Request.Name)
```
jackyalbo: Thinking if we can/should combine the two function - GetAffinity and the new desriedCoreAffinity.

jackyalbo: Also another thing is, maybe we should generalize this function to be not specific just to core, but to any HA we might have.

nadavMiz: sure I can use GetAffinity from inside desiredCoreAffinity

nadavMiz: regarding generalizing, the code is mostly generalized as is. I generalized the two helper functions. for the main core getaffinity function. I think we can keep it core specific for the moment. there are already specific getaffinity functions that act different for endpoints and cnpg. also this function will be only used by core. when we will add another ha that works the same as core. we can easily generalize and apply function to both

jackyalbo (approved):

PR #2081, opened 2026-08-10
Add support for network policy

tangledbytes:
### Describe the Problem

We want to add network policies for all of the noobaa components.

### Explain the changes
This PR adds the required network policies for all of the noobaa components. Currently the network policies DELIBERATELY have only ingress rules. egress rules are out of scope for now.

1. https://redhat.atlassian.net/browse/RHSTOR-8134?focusedCommentId=17758637

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary

* **New Features**
  * Added network access policies for NooBaa core, S3 endpoints, operator, database, and storage-pool components.
  * Added network policies for the CloudNative PostgreSQL controller and monitoring endpoints.
  * Policies allow required internal service, API, ingress, and monitoring traffic on designated ports.
  * Policies are automatically applied and reconciled during deployment.
  * Database and PostgreSQL controller policies are configured when applicable.
  * Added permissions required to manage these network policies.

### DIFF
--- a/deploy/internal/networkpolicy-core.yaml
+++ b/deploy/internal/networkpolicy-core.yaml
@@ -1,0 +1,35 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: noobaa-core
+  labels:
+    app: noobaa
+spec:
+  podSelector:
+    matchLabels:
+      noobaa-core: noobaa
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - podSelector: {}
+    ports:
+    - protocol: TCP
+      port: 8080
+    - protocol: TCP
+      port: 8443
+    - protocol: TCP
+      port: 8444
+    - protocol: TCP
+      port: 8445
+    - protocol: TCP
+      port: 8446
+    - protocol: TCP
+      port: 60100
+    - protocol: UDP
+      port: 5140
+  - from:
+    - namespaceSelector: {}
+    ports:
+    - protocol: TCP
+      port: 8443
--- a/deploy/internal/networkpolicy-endpoint.yaml
+++ b/deploy/internal/networkpolicy-endpoint.yaml
@@ -1,0 +1,45 @@
+apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: noobaa-endpoint
+  labels:
+    app: noobaa
+spec:
+  podSelector:
+    matchLabels:
+      noobaa-s3: noobaa
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - podSelector: {}
+    ports:
+    - protocol: TCP
+      port: 6001
+    - protocol: TCP
+      port: 6443
+    - protocol: TCP
+      port: 7443
+    - protocol: TCP
+      port: 8444
+    - protocol: TCP
+      port: 9443
+    - protocol: TCP
+      port: 13443
+    - protocol: TCP
+      port: 14443
+  - from:
+    - namespaceSelector: {}
+    ports:
+    - protocol: TCP
+      port: 6001
+    - protocol: TCP
+      port: 6443
+    - protocol: TCP
+      port: 7443
+    - protocol: TCP
+      port: 9443
+    - protocol: TCP
+      port: 13443
+    - protocol: TCP
+      port: 14443
--- a/pkg/bundle/deploy.go
+++ b/pkg/bundle/deploy.go
@@ -4914,6 +4914,146 @@ metadata:
     app: noobaa
 `
 
+const Sha256_deploy_internal_networkpolicy_core_yaml = "e03679b5dda3c850d4918bb9f82a08e4c07809b115cf40aa4e490a212aa3a662"
+
+const File_deploy_internal_networkpolicy_core_yaml = `apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: noobaa-core
+  labels:
+    app: noobaa
+spec:
+  podSelector:
+    matchLabels:
+      noobaa-core: noobaa
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - podSelector: {}
+    ports:
+    - protocol: TCP
+      port: 8080
+    - protocol: TCP
+      port: 8443
+    - protocol: TCP
+      port: 8444
+    - protocol: TCP
+      port: 8445
+    - protocol: TCP
+      port: 8446
+    - protocol: TCP
+      port: 60100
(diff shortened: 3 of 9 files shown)


pkg/bundle/deploy.go
```diff
@@ -4911,6 +4911,169 @@ metadata:
...
+const Sha256_deploy_internal_networkpolicy_cnpg_controller_yaml = "2e9187e2dd7823bd10c5b0394ff0b6fa72a5ac239ea8ffb518f5100a2c1ab764"
+
+const File_deploy_internal_networkpolicy_cnpg_controller_yaml = `apiVersion: networking.k8s.io/v1
+kind: NetworkPolicy
+metadata:
+  name: cnpg-controller-manager
+  labels:
+    app: noobaa
+spec:
+  podSelector:
```
OdedViner: @tangledbytes Based on the new Operator requirements ([HPSTRAT-104](https://redhat.atlassian.net/browse/HPSTRAT-104)), we are moving away from static NetworkPolicies.

To align with Phase 1 (ODF 4.23 / OCP 5.0), please remove the static NetworkPolicy manifests from the CSV/bundle directories and implement their creation and lifecycle management directly inside the Go reconciler.

pkg/system/phase2_network_policies.go
```diff
@@ -0,0 +1,49 @@
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// SetDesiredNetworkPolicyCore updates the core network policy pod selector
+// to match the actual NooBaa system name
+func (r *Reconciler) SetDesiredNetworkPolicyCore() error {
```
OdedViner, line 33: Could you please split Phase 1 and Phase 2 into two separate PRs?

tangledbytes: Is that going to be helpful? We would need the entire thing in one single commit or else it would be a broken feature.

travisn: To clarify, the two phases are:
1. (4.23) Create network policies for operands. 
   a. No network policies are allowed in the CSV. OCP specifically prohibits it for this release.
   b. RBAC is required for the operator to create/update the network policies
2. (5.1) Create network policies for operators in the CSV. 

My understanding from this PR is that you have implemented the policies for operands and created the needed RBAC. The implementation may just need to embed the yaml in code (e.g. with `//go:embed` instead of including it in the CSV.

dannyzaken: @tangledbytes, if I understand correctly, `NetworkPolicyOperator` and `NetworkPolicyCNPGController` are part of phase 2 and are out of scope for this PR/release.

@OdedViner @travisn, this change does not add the network policies to the CSV. They are created in the reconciliation loop of noobaa system.

deploy/internal/networkpolicy-core.yaml
```diff
@@ -0,0 +1,35 @@
...
+      port: 8446
+    - protocol: TCP
+      port: 60100
+    - protocol: UDP
+      port: 5140
+  - from:
+    - namespaceSelector: {}
+    ports:
+    - protocol: TCP
+      port: 8443
```
dannyzaken, line 35: Why do we need access to the mgmt port from all namespaces?

pkg/system/phase2_network_policies.go
```diff
@@ -0,0 +1,49 @@
...
+	r.NetworkPolicyCore.Spec.PodSelector.MatchLabels["noobaa-core"] = r.Request.Name
+	return nil
+}
+
+// SetDesiredNetworkPolicyEndpoint updates the endpoint network policy pod selector
+// to match the actual NooBaa system name
+func (r *Reconciler) SetDesiredNetworkPolicyEndpoint() error {
+	r.NetworkPolicyEndpoint.Spec.PodSelector.MatchLabels["noobaa-s3"] = r.Request.Name
```
dannyzaken, line 41: Does the network policy allow external access through the LB service? 
The external access policy should be controlled by `DisableLoadBalancerService`.

pkg/system/phase2_network_policies.go
```diff
@@ -0,0 +1,49 @@
+package system
+
+// ReconcileNetworkPolicies reconciles all network policies for the NooBaa system
+func (r *Reconciler) ReconcileNetworkPolicies() error {
```
dannyzaken, line 5: we also need a network policy for pv-pool pods

deploy/internal/networkpolicy-endpoint.yaml
```diff
@@ -0,0 +1,43 @@
...
+    matchLabels:
+      noobaa-s3: noobaa
+  policyTypes:
+  - Ingress
+  ingress:
+  - from:
+    - podSelector: {}
+    ports:
+    - protocol: TCP
+      port: 6001
```
dannyzaken, line 18: This network policy makes the http port unreachable from other namespaces. Is that on purpose? This can block applications that access this port

deploy/internal/networkpolicy-core.yaml
tangledbytes: Actually I couldn't think of a better selector -
1. Backingstore can be in other namespace as well and I assumed that they might need access to API.
2. Some cu actually use noobaa CLI from within pods which could be in other namespaces.

WDYT?

pkg/system/phase2_network_policies.go
tangledbytes: AFAIK network policies do not affect LB and nodeport as they are very CNI specific.

pkg/system/phase2_network_policies.go
tangledbytes: I misunderstood it seems. I have updated the PR.

deploy/internal/networkpolicy-core.yaml
dannyzaken: I now see that `mgmt-https` port is used in the prometheus service monitor, so I guess it's needed for that.

dannyzaken (approved):

pkg/system/phase2_network_policies.go
dannyzaken: We should probably test it.
@coderabbitai, any idea?

pkg/system/phase2_network_policies.go
travisn: @dannyzaken Thanks for clarifying, I was thinking it was related to the csv with the bundle package and phase2 filenames. Does the phase2 filename relate to a noobaa concept? If it's for phase 2 of network policies, then it seems incorrectly named. We don't expect network policy phase 2 changes to be in code or in this release. The phase 2 changes are only expected in the csv in v5.1.

PR #2083, opened 2026-08-11
Fix DB dump command to accept user preference

naveenpaul1:
### Explain the changes
1. If the pod name was explicitly provided (e.g. via the --db-pod CLI flag) it is used as-is.
Otherwise, the pod is auto-detected by looking for either:
    1. The legacy standalone postgres pod (e.g. "noobaa-db-pg-0"), labeled "noobaa-db=postgres"
    2. The primary pod of a CNPG managed cluster (e.g. "noobaa-db-pg-cluster-1"), which does not

1. https://redhat.atlassian.net/browse/DFBUGS-4113

### Testing Instructions:
1. nb diagnostics db-dump -n noobaa --db-pod noobaa-db-pg-0

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **New Features**
  * Added a `--dump-from` option to diagnostics collection and database dump commands.
  * Choose between the primary or replica database; the primary is used by default.
  * Database dumps now target the selected database and produce locally accessible SQL files.
* **Bug Fixes**
  * Improved handling and reporting of database discovery, command execution, streaming, and dump processing errors.
* **Documentation**
  * Added guidance for using the new database dump option.

### DIFF
--- a/pkg/diagnostics/dbdump.go
+++ b/pkg/diagnostics/dbdump.go
@@ -1,7 +1,10 @@
 package diagnostics
 
 import (
+	"bytes"
+	"compress/gzip"
 	"fmt"
+	"io"
 	"os"
 	"os/exec"
 	"time"
@@ -10,7 +13,22 @@ import (
 	"github.com/noobaa/noobaa-operator/v5/pkg/util"
 	"github.com/sirupsen/logrus"
 
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
 	"github.com/spf13/cobra"
+)
+
+const (
+
+	// cnpgClusterLabelKey is set (to the cluster name) on every pod managed by a CNPG Cluster.
+	cnpgClusterLabelKey = "cnpg.io/cluster=noobaa-db-pg-cluster"
+	// cnpgInstanceRoleLabelKey and cnpgLegacyRoleLabelKey mark the primary/replica role
+	// of a CNPG managed pod.
+	cnpgInstanceRoleLabelKey = "cnpg.io/instanceRole"
+	cnpgPrimaryRoleValue     = "primary"
+	cnpgReplicaRoleValue     = "replica"
 )
 
 // CollectorDbDump configuration for diagnostics
@@ -20,16 +38,18 @@ type CollectorDbDump struct {
 	remoteTarPath    string // Remote path (at pod) of tar'd db dump
 	kubeconfig       string
 	kubeCommand      string
+	dbDumpFrom       string // DB Preference. Auto-selecct 'replica' when value is empty
 	log              *logrus.Entry
 }
 
-func newCollectorDbDump(kubeconfig string) *CollectorDbDump {
+func newCollectorDbDump(kubeconfig string, dbDumpFrom string) *CollectorDbDump {
 	c := new(CollectorDbDump)
 	c.folderName = fmt.Sprintf("%s_%d", "noobaa_db_dump", time.Now().Unix())
 	c.remoteTarPath = fmt.Sprintf("%s/%s%s", "/tmp", c.folderName, ".tar.gz")
 	c.log = util.Logger()
 	c.kubeconfig = kubeconfig
 	c.kubeCommand = util.GetAvailabeKubeCli()
+	c.dbDumpFrom = dbDumpFrom
 	return c
 }
 
@@ -37,13 +57,14 @@ func newCollectorDbDump(kubeconfig string) *CollectorDbDump {
 func RunDump(cmd *cobra.Command, args []string) {
 	kubeconfig, _ := cmd.Flags().GetString("kubeconfig")
 	destDir, _ := cmd.Flags().GetString("dir")
+	dbDumpFrom, _ := cmd.Flags().GetString("dump-from")
 
-	CollectDBDump(kubeconfig, destDir)
+	CollectDBDump(kubeconfig, destDir, dbDumpFrom)
 }
 
 // CollectDBDump exposes the functionality to the diagnostics collect mechanism
-func CollectDBDump(kubeconfig string, destDir string) {
-	c := newCollectorDbDump(kubeconfig)
+func CollectDBDump(kubeconfig string, destDir string, dbDumpFrom string) {
+	c := newCollectorDbDump(kubeconfig, dbDumpFrom)
 
 	c.log.Println("Collecting db dump")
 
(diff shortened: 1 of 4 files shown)


pkg/diagnostics/dbdump.go
```diff
@@ -58,11 +79,77 @@ func CollectDBDump(kubeconfig string, destDir string) {
...
+				Namespace: options.Namespace,
+			},
+		}
+		if !util.KubeCheckQuiet(pod) {
+			return "", fmt.Errorf("Could not find noobaa db pod %q in namespace %q", c.dbPodName, options.Namespace)
+		}
+		return c.dbPodName, nil
+	}
+
+	standaloneSelector, err := labels.Parse(standalonePostgresLabelSelector)
```
dannyzaken: standalone DB is not relevant anymore, and for sure it should not be the default.
I think that the the default should be the secondary (`cnpg.io/instanceRole=replica`) if it is in a `Ready` state. If not then fallback to the primary.

pkg/diagnostics/dbdump.go
```diff
@@ -58,11 +79,77 @@ func CollectDBDump(kubeconfig string, destDir string) {
...
+	if err != nil {
+		c.log.Printf(`❌ cannot find db pod: %v`, err)
+		return err
+	}
+	c.log.Printf("Using db pod %q for db dump", dbPodName)
+
 	// In case of a postgres db, the dump is a single file so in this case we can
 	// redirect the output of the dump straight to a local file
-	cmd := exec.Command(c.kubeCommand, "exec", "-n", options.Namespace, "-it", "pod/noobaa-db-pg-0", "--", "pg_dumpall")
+	cmd := exec.Command(c.kubeCommand, "exec", "-n", options.Namespace, "-it", fmt.Sprintf("pod/%s", dbPodName), "--", "pg_dumpall")
```
dannyzaken: A few comments:
* we should remove the `t` from `-it`, as it translates `\n` to `\r\n`. so far it worked for us, but it can potentially insert silent failures into jsonb records.
* add `--no-role-passwords` flag to pg_dumpall, to exclude password hashes from the dump
* consider piping to gzip inside the `exec` command, to avoid transferring and writing the uncompressed dump file and only then compressing

```suggestion
    dumpScript := "set -o pipefail; pg_dumpall --no-role-passwords | gzip -c"
    cmd := exec.Command(c.kubeCommand, "exec", "-n", options.Namespace, "-i", fmt.Sprintf("pod/%s", dbPodName), "--", "bash", "-c", dumpScript)
```

pkg/diagnostics/dbdump.go
```diff
@@ -10,40 +10,61 @@ import (
...
 	c.kubeCommand = util.GetAvailabeKubeCli()
+	c.dbPodName = dbPodName
 	return c
 }
 
 // RunDump generates a tar file of a db dump
 func RunDump(cmd *cobra.Command, args []string) {
 	kubeconfig, _ := cmd.Flags().GetString("kubeconfig")
 	destDir, _ := cmd.Flags().GetString("dir")
+	dbPodName, _ := cmd.Flags().GetString("db-pod")
```
dannyzaken: Receiving the DB pod name has a potential for user error. I suggest maybe adding `--dump-from` flag that can receive `primary` or `replica`.

dannyzaken: Better values for this flag are `primary` and `prefer-replica`, since if the replica is not available, we fall back to the primary.

naveenpaul1: updated

pkg/diagnostics/dbdump.go
```diff
@@ -10,40 +12,59 @@ import (
...
 )
 
+const (
+
+	// cnpgClusterLabelKey is set (to the cluster name) on every pod managed by a CNPG Cluster.
+	cnpgClusterLabelKey = "cnpg.io/cluster=noobaa-db-pg-cluster"
+	// cnpgInstanceRoleLabelKey and cnpgLegacyRoleLabelKey mark the primary/replica role
+	// of a CNPG managed pod.
+	cnpgInstanceRoleLabelKeyReplica = "cnpg.io/instanceRole"
+	cnpgLegacyRoleLabelKey          = "role"
```
dannyzaken: unused

pkg/diagnostics/dbdump.go
```diff
@@ -10,40 +12,59 @@ import (
...
 	"github.com/spf13/cobra"
 )
 
+const (
+
+	// cnpgClusterLabelKey is set (to the cluster name) on every pod managed by a CNPG Cluster.
+	cnpgClusterLabelKey = "cnpg.io/cluster=noobaa-db-pg-cluster"
+	// cnpgInstanceRoleLabelKey and cnpgLegacyRoleLabelKey mark the primary/replica role
+	// of a CNPG managed pod.
+	cnpgInstanceRoleLabelKeyReplica = "cnpg.io/instanceRole"
```
dannyzaken: This is not specific to replica, rename to `cnpgInstanceRoleLabelKey`

pkg/diagnostics/dbdump.go
```diff
@@ -58,11 +79,69 @@ func CollectDBDump(kubeconfig string, destDir string) {
...
+		return err
+	}
+	c.log.Printf("Using db pod %q for db dump", dbPodName)
+
+	// pg_dumpall's output is piped through gzip inside the pod to reduce the amount of data
+	// transferred over the exec stream (important for large DBs). The gzip stream is
+	// decompressed locally below so the resulting .sql file on disk is plain SQL text, as its
+	// extension implies, rather than leaving it gzip-compressed.
+	dumpScript := "set -o pipefail; pg_dumpall --no-role-passwords | gzip -c"
+	cmd := exec.Command(c.kubeCommand, "exec", "-n", options.Namespace, "-i", fmt.Sprintf("pod/%s", dbPodName), "--", "bash", "-c", dumpScript)
```
dannyzaken, line 156: capture stderr to print on error

pkg/diagnostics/diagnostics.go
```diff
@@ -39,6 +39,7 @@ func CmdCollect() *cobra.Command {
 	}
 	cmd.Flags().String("dir", "", "collect noobaa diagnostics tar file into destination directory")
 	cmd.Flags().Bool("db-dump", false, "collect db dump in addition to diagnostics")
+	cmd.Flags().String("dump-from", "", "specifies the database type from which to collect the dump. Valid options are 'primary' and 'replica' (default: 'replica')")
```
dannyzaken: After investigating a bit, I think that it's better to actually default to the primary. Long-running DB dumps from the secondary will be canceled in some replication conditions. 
To simplify it, we should also not do any fallback - if there is a failure, just provide a clear message with the stderr output and let the user retry if needed.

For the flags definition, set the default as the second parameter instead of `""`



```suggestion
	cmd.Flags().String("dump-from", "primary", "specifies the database type from which to collect the dump. Valid options are 'primary' and 'replica' (default: 'primary')")
```

pkg/diagnostics/dbdump.go
```diff
@@ -58,11 +79,69 @@ func CollectDBDump(kubeconfig string, destDir string) {
...
+		return "", fmt.Errorf("Could not find noobaa DB pod in namespace %q, "+
+			"Make sure your DB pods are up and running ", options.Namespace)
+	}
+	if foundDBPod {
+		for _, pod := range podList.Items {
+			role := pod.Labels[cnpgInstanceRoleLabelKeyReplica]
+			if role == cnpgPrimaryRoleValue {
+				primaryDBPodName = pod.Name
+			} else if role == cnpgReplicaRoleValue && pod.Status.Phase == "Running" {
+				replicaDBPodName = pod.Name
```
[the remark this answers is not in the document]
dannyzaken: @naveenpaul1 I agree with this comment. use only a ready pod

dannyzaken (approved):

PR #2093, opened 2026-08-18
[Backport to 5.22] backporting fixes

liranmauda:
- Fix: stop PV-pool agent recreate loop on toleration superset match
- Fix DB dump command to accept user preference


naveenpaul1 (approved):
DB dump changes looks good to me

PR #2103, opened 2026-08-24
core HA | set PodManagementPolicy to Parallel

nadavMiz:
### Describe the Problem
With HA, the standby core pod is not Ready. Default StatefulSet ordering (OrderedReady) can block scaling or replacing the other core pod until the standby becomes Ready.

### Explain the changes
1. Use podManagementPolicy: Parallel on noobaa-core.
2. since podManagementPolicy is immutable, On upgrade, auto-recreate the StatefulSet (keep pods) 

1. 

### Testing Instructions:
1. `go test ./pkg/system/ -run 'CoreStatefulSet|ClearStatefulSet'`

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

- **New Features**
  - Core service pods are now created and updated concurrently.
  - New installations use parallel pod management.

- **Bug Fixes**
  - Existing deployments using sequential pod management are automatically migrated to the parallel approach while preserving running pods.
  - StatefulSet reconciliation now maintains consistent pod management settings during updates.
  - Management metrics now target only ready pods.

### DIFF
--- a/pkg/system/phase2_creating.go
+++ b/pkg/system/phase2_creating.go
@@ -231,6 +231,10 @@ func (r *Reconciler) ReconcilePhaseCreatingForMainClusters() error {
 		}
 	}
 
+	if err := r.ensureParallelCoreStatefulSet(); err != nil {
+		return err
+	}
+
 	if err := r.ReconcileObject(r.CoreApp, r.SetDesiredCoreApp); err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


pkg/system/phase2_creating.go
```diff
@@ -654,6 +658,60 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
...
+	if current == "" {
+		current = appsv1.OrderedReadyPodManagement
+	}
+	r.Logger.Infof(
+		"upgrading %s StatefulSet to parallel pod startup (delete STS, keep pods, recreate) from %q",
+		r.CoreApp.Name, current,
+	)
+
+	// delete the statefulset and keep the pods
+	orphan := metav1.DeletePropagationOrphan
```
dannyzaken, line 690: Why is this needed?

pkg/system/phase2_creating.go
```diff
@@ -654,6 +658,60 @@ func (r *Reconciler) setDesiredCoreEnv(c *corev1.Container) {
...
+// ensures that the statefulset has parallel pod management policy
+// PodManagementPolicy is immutable, so we need to delete the statefulset and create a new one
+// can happen when upgrading from older operator versions
+func (r *Reconciler) ensureParallelCoreStatefulSet() error {
+	if !util.KubeCheckQuiet(r.CoreApp) {
+		return nil
+	}
+	if !coreStatefulSetHasOrderedReadyPolicy(r.CoreApp) {
+		return nil
+	}
```
dannyzaken: I think the code is clearer if you rename the function to check for parallel and remove the `!`. Something like this:


```suggestion
	if hasParallelPodManagement(r.CoreApp) {
		return nil
	}
```

pkg/system/phase2_creating.go
nadavMiz: its not strictly needed. its just so we don't delete existing pods when deleting the sts. since we only need to change the sts itself. the pods will remain if available

PR #2104, opened 2026-08-25
Update external psql check to allow versions higher than 15

liranmauda:
Update external psql check to allow versions higher than 15

Fixes [DFBUGS-10092](https://redhat.atlassian.net/browse/DFBUGS-10092)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated PostgreSQL version validation to support version 15 and newer.
  * Improved handling and reporting of unsupported PostgreSQL versions.

### DIFF
--- a/pkg/system/phase1_verifying.go
+++ b/pkg/system/phase1_verifying.go
@@ -242,17 +242,17 @@ func (r *Reconciler) checkExternalPg(postgresDbURL string) error {
 				dbURL, err))
 	}
 	// Query the PostgreSQL version
-	var version string
+	var version int
 	err = db.QueryRow("SELECT current_setting('server_version_num')::integer / 10000").Scan(&version)
 	if err != nil {
 		return util.NewPersistentError("InvalidExternalPgVersion",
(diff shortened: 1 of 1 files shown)


pkg/system/phase1_verifying.go
```diff
@@ -242,17 +242,17 @@ func (r *Reconciler) checkExternalPg(postgresDbURL string) error {
...
 	err = db.QueryRow("SELECT current_setting('server_version_num')::integer / 10000").Scan(&version)
 	if err != nil {
 		return util.NewPersistentError("InvalidExternalPgVersion",
 			fmt.Sprintf("failed getting version of external DB url: %q, error: %s",
 				dbURL, err))
 	}
-	// Check if the version is 15
-	if version != "15" {
+	// Reject PostgreSQL versions older than 15
+	if version < 15 {
```
jackyalbo, line 253: I think we wanted to support only the specific version we worked with? @dannyzaken WDYT? should we just allow 15-16, or 15 and up?

dannyzaken (approved):

PR #2107, opened 2026-09-01
Return reconciler error on endpoint update in CR spec

karthikps97:
### Describe the Problem
When the endpoint field on the CR spec for backingstore or namespace store is edited, the store goes into ready phase but the changes are not actually propagated. We need to block this path and ask user to reset the endpoint field back to the original state.

Note: There is already a new CLI added to support the endpoint changes.

### Explain the changes
1. Log and return an error for both namespace store and backingstore when an endpoint mismatch is detected on spec.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented reconciliation from automatically changing an existing cloud pool or namespace endpoint.
  * Added clear errors when an endpoint change is detected, prompting restoration of the previously configured endpoint.
  * Continued supporting non-endpoint connection updates, such as connection type or identity changes.

### DIFF
--- a/pkg/backingstore/reconciler.go
+++ b/pkg/backingstore/reconciler.go
@@ -93,6 +93,8 @@ type Reconciler struct {
 	CreateHostsPoolParams          *nb.CreateHostsPoolParams
 	UpdateHostsPoolParams          *nb.UpdateHostsPoolParams
 	UpdateExternalConnectionParams *nb.UpdateExternalConnectionParams
+	EndpointChanged                bool
+	OldCoreEndpoint                string
 }
 
 // Own sets the object owner references to the backingstore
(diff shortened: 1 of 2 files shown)


pkg/backingstore/reconciler.go
```diff
@@ -636,6 +636,12 @@ func (r *Reconciler) ReadSystemInfo() error {
 			pool.CloudInfo.Endpoint != conn.Endpoint ||
 			pool.CloudInfo.Identity != string(conn.Identity) {
 			r.Logger.Warnf("using existing pool but connection mismatch %+v pool %+v %+v", conn, pool, pool.CloudInfo)
+
+			if pool.CloudInfo.Endpoint != conn.Endpoint {
```
romayalon, line 640: wouldn't this return an error after the update already happened? 
I would expect this should be changed in 2 places - 
1. When we call updateExternalConnection
2. in the admission webhooks of the backingstore and namespacestores

romayalon (commented):
@karthikps97 
would this keep reconciling forever and show this message?

PR #2109, opened 2026-09-03
Returning PersistentError when create system hangs on INITIALIZING for 5 min

liranmauda:
### Explain the changes
Returning PersistentError when create system hangs on INITIALIZING for 5 min

[DFBUGS-10424](https://redhat.atlassian.net/browse/DFBUGS-10424)

### DIFF
--- a/pkg/system/phase4_configuring.go
+++ b/pkg/system/phase4_configuring.go
@@ -47,6 +47,7 @@ const (
 	ibmLocation                       = "%s-standard"
 	ibmCosBucketCred                  = "ibm-cloud-cos-creds"
 	minutesToWaitForDefaultBSCreation = 10
+	systemInitializingTimeout         = 5 * time.Minute // RPCSendTimeout is 2 min, we will wait 5
 	credentialsKey                    = "credentials"
 	metricsAuthKey                    = "metrics_token"
 	serviceMonitorCAFile              = "/etc/prometheus/configmaps/serving-certs-ca-bundle/service-ca.crt"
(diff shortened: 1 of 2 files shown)


pkg/system/phase4_configuring.go
```diff
@@ -228,6 +229,16 @@ func (r *Reconciler) SetDesiredSecretOp() error {
 			return util.NewPersistentError("SystemCouldNotInitialize",
 				"Something went wrong during system initialization")
 
+		case "INITIALIZING":
+			// create_system is still running (happy path) or core died mid-create.
+			// Use core last_state_change (epoch ms) so the clock survives operator restarts.
+			// Do not sleep here — a TemporaryError requeues reconcile every 3s.
+			if res1.LastStateChange > 0 && time.Since(time.UnixMilli(res1.LastStateChange)) > systemInitializingTimeout {
```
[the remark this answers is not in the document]
liranmauda: It should always be a positive value, and if not, it is a bug in the core.
We should not overprotect this case

pkg/system/phase4_configuring.go
```diff
@@ -228,6 +229,16 @@ func (r *Reconciler) SetDesiredSecretOp() error {
...
+				return util.NewPersistentError("SystemInitializingTimeout",
```
shirady, line 237: You can also add an event with `r.Recorder.Eventf` for debugging, WDYT?

shirady, line 237: I want to ensure that you saw that the suggested message from the reporter is:

> NooBaa system has been stuck in INITIALIZING state — see documentation for reinstall guidance

shirady (approved):
LGTM

pkg/system/phase4_configuring.go
liranmauda: This is an upstream, and we actually do not have any documentation.
So, in this case, they should get an easy error and search for and find the downstream KCS.

liranmauda: I do not think it is necessary. It should just return a simple error that support can look at and find a KCS.
The remedy for this one is to reinstall.
I do not think we want any Kubernetes event here, even if it will make it easier.

PR #2113, opened 2026-09-08
Add 'noobaa backingstore replace' CLI command

kajalpareek-lab:
## Problem
[DFBUGS-6233](https://redhat.atlassian.net/browse/DFBUGS-6233)
When replacing the default backing store on an active system, CLI/UI/S3-created
buckets remain tied to the old pool because BucketClass updates only propagate
to OBC-created buckets. There is no supported CLI command to safely detach the
old backing store from all buckets and accounts.

## Solution

Add a new noobaa backingstore replace CLI command that calls the
pool_api.safe_replace_pool RPC (added in the companion noobaa-core PR).

Usage:

    noobaa backingstore replace <old-backing-store> <new-backing-store> [--migrate]

Workflow:
1. noobaa backingstore replace old new --migrate (start mirroring)
2. Wait for data replication to complete
3. noobaa backingstore replace old new (finalize replacement)
4. oc delete backingstore old (remove the old store)

## Changes

- pkg/nb/types.go: SafeReplacePoolParams and SafeReplacePoolReply structs
- pkg/nb/api.go: SafeReplacePoolAPI method on Client interface and RPCClient
- pkg/backingstore/backingstore.go: CmdReplace and RunReplace functions
- pkg/nb/api_test.go: unit tests for params/reply serialization
- test/cli/test_cli_functions.sh: end-to-end CLI integration test

## Depends On

- noobaa-core: safe_replace_pool RPC (DFBUGS-6233)

## Tests

- 4 unit tests for SafeReplacePoolParams/Reply serialization (all passing)
- 1 CLI integration test (test_backingstore_replace) covering full workflow

Fixes: DFBUGS-6233

## Summary by CodeRabbit

- **New Features**
  - Added a command to safely replace an existing backing store with a ready replacement.
  - Validates backing-store availability and prevents selecting the same store.
  - Supports migration mirroring before finalizing replacement.
  - Updates bucket tiers and account defaults during replacement.
  - Reports replacement results and provides follow-up migration or cleanup guidance.

- **Tests**
  - Added coverage for replacement request and response handling.
  - Added end-to-end coverage for migration, replacement finalization, default-store updates, and cleanup.


pkg/backingstore/backingstore.go
```diff
@@ -352,6 +353,27 @@ func CmdDelete() *cobra.Command {
 	return cmd
 }
 
+// CmdReplace returns a CLI command
+func CmdReplace() *cobra.Command {
+	cmd := &cobra.Command{
```
naveenpaul1, line 358: Is this replace work on all the buckets at once? So if there are 100 buckets and we are doing migration on those buckets will it create any issue?

pkg/backingstore/backingstore.go
```diff
@@ -352,6 +353,27 @@ func CmdDelete() *cobra.Command {
...
+		Use:   "replace <old-backing-store> <new-backing-store>",
+		Short: "Replace one backing store with another across all buckets",
+		Long: `Replace references to one backing store with another across all bucket tiers and account defaults.
```
naveenpaul1, line 361: I think, we should also provide buckets specific replace too,

pkg/backingstore/backingstore.go
```diff
@@ -945,6 +967,90 @@ func RunCreatePVPool(cmd *cobra.Command, args []string) {
...
+	}
+
+	if newBS.Status.Phase != nbv1.BackingStorePhaseReady {
+		log.Fatalf(`❌ BackingStore %q is not Ready (current phase: %s)`, newBSName, newBS.Status.Phase)
+	}
+
+	nbClient := system.GetNBClient()
+
+	if migrate {
+		log.Infof("🔄 Starting migration: mirroring data from %q to %q", oldBSName, newBSName)
```
naveenpaul1, line 1009: Are we really starting the migration here? Is it done by background worker?

pkg/backingstore/backingstore.go
```diff
@@ -945,6 +967,90 @@ func RunCreatePVPool(cmd *cobra.Command, args []string) {
...
+	log.Infof("   Accounts updated: %d", reply.UpdatedAccounts)
+
+	if migrate {
+		log.Infof("")
+		log.Infof("📋 Next steps:")
+		log.Infof("   1. Wait for data replication to complete")
+		log.Infof("      Check progress: noobaa bucket status <bucket-name>")
+		log.Infof("   2. Finalize the replacement:")
+		log.Infof("      noobaa backingstore replace %s %s", oldBSName, newBSName)
+		log.Infof("   3. Prevent operator from recreating the default backing store:")
```
naveenpaul1, line 1041: why are we giving these steps in response?

pkg/backingstore/backingstore.go
```diff
@@ -945,6 +967,90 @@ func RunCreatePVPool(cmd *cobra.Command, args []string) {
...
+		log.Fatalf(`❌ Failed to replace backing store: %s`, err)
+	}
+
+	if reply.ReplacedTiers == 0 && reply.UpdatedAccounts == 0 {
+		log.Infof("ℹ️  No tiers or accounts reference %q — nothing to replace", oldBSName)
+		return
+	}
+
+	log.Infof("✅ %s", reply.Mode)
+	log.Infof("   Tiers updated:    %d", reply.ReplacedTiers)
```
naveenpaul1, line 1031: not sure customer is familier with Tiers, Its better to return bucket name too

pkg/backingstore/backingstore.go
aayushchouhan09: Agreed, may be we can add `--bucket-name` flag (mandatory)

pkg/backingstore/backingstore.go
```diff
@@ -352,6 +353,27 @@ func CmdDelete() *cobra.Command {
...
+Use --migrate to first enable mirroring between the old and new backing stores,
+allowing the system to replicate existing data before completing the switch.
+
+Workflow:
+  1. noobaa backingstore replace <old> <new> --migrate   (start mirroring)
+  2. Wait for data replication to complete
+  3. noobaa backingstore replace <old> <new>              (finalize replacement)
+  4. oc delete backingstore <old>                          (remove the old store)`,
+		Run: RunReplace,
+	}
```
aayushchouhan09, line 372: Suggestion: We can have two more flags `--all-buckets` and `--bucket-name`.

pkg/backingstore/backingstore.go
```diff
@@ -945,6 +967,90 @@ func RunCreatePVPool(cmd *cobra.Command, args []string) {
...
+	log.Infof("   Tiers updated:    %d", reply.ReplacedTiers)
+	log.Infof("   Accounts updated: %d", reply.UpdatedAccounts)
+
+	if migrate {
+		log.Infof("")
+		log.Infof("📋 Next steps:")
+		log.Infof("   1. Wait for data replication to complete")
+		log.Infof("      Check progress: noobaa bucket status <bucket-name>")
+		log.Infof("   2. Finalize the replacement:")
+		log.Infof("      noobaa backingstore replace %s %s", oldBSName, newBSName)
```
aayushchouhan09, line 1040: Can you explain this how `--mirror` works? Do we need to run the replace command twice here?

pkg/backingstore/backingstore.go
```diff
@@ -945,6 +967,90 @@ func RunCreatePVPool(cmd *cobra.Command, args []string) {
...
+		log.Infof("      oc patch noobaa/noobaa -n %s --type json --patch='[{\"op\":\"add\",\"path\":\"/spec/manualDefaultBackingStore\",\"value\":true}]'", options.Namespace)
+		log.Infof("   4. Delete the old backing store:")
+		log.Infof("      oc delete backingstore %s -n %s", oldBSName, options.Namespace)
+	} else {
+		log.Infof("")
+		log.Infof("📋 The old backing store %q has been detached from all tiers.", oldBSName)
+		log.Infof("   You can now delete it:")
+		log.Infof("      oc patch noobaa/noobaa -n %s --type json --patch='[{\"op\":\"add\",\"path\":\"/spec/manualDefaultBackingStore\",\"value\":true}]'", options.Namespace)
+		log.Infof("      oc delete backingstore %s -n %s", oldBSName, options.Namespace)
+	}
```
aayushchouhan09, line 1051: Not sure if we need to print this much info or else we can simply leave a note that we need to manually delete the old backingstore after enabling the `manualDefaultBackingstore`.

pkg/backingstore/backingstore.go
```diff
@@ -945,6 +967,90 @@ func RunCreatePVPool(cmd *cobra.Command, args []string) {
...
+
+	log.Infof("✅ %s", reply.Mode)
+	log.Infof("   Tiers updated:    %d", reply.ReplacedTiers)
+	log.Infof("   Accounts updated: %d", reply.UpdatedAccounts)
+
+	if migrate {
+		log.Infof("")
+		log.Infof("📋 Next steps:")
+		log.Infof("   1. Wait for data replication to complete")
+		log.Infof("      Check progress: noobaa bucket status <bucket-name>")
```
aayushchouhan09, line 1038: Do we have any tracker or something to verify the mirroring status?

pkg/backingstore/backingstore.go
kajalpareek-lab: Yes, it updates all affected tiers in a single system_store.make_changes() call — one atomic bulk DB write. Even with 100 buckets, it's just updating tier documents, not moving data. The heavy lifting (actual data replication in migration mode) is handled by the existing mirror_writer background worker which processes buckets independently. So the replace call itself is fast regardless of bucket count.

pkg/backingstore/backingstore.go
kajalpareek-lab: I thought about this, but I don’t think it works cleanly because tiers can be shared by multiple buckets through tiering policies.
For example, if we replace a pool in a tier for Bucket A, Bucket B could also be affected if both buckets use the same tier. So adding per-bucket filtering could be misleading because users may think they are changing only one bucket when they could actually impact others.
Also, the main reason this bug exists (DFBUGS-6233) is that there was no way to update all non-OBC buckets at once. Making `--bucket-name` mandatory would mean users have to manually provide every bucket, which is exactly the problem we are trying to solve.
If there is a real customer need to filter by bucket, we can add it as an optional flag later if there's a concrete use case. But we should design it carefully because of the shared-tier behavior.

pkg/backingstore/backingstore.go
kajalpareek-lab: We're not starting any new process. The RPC just adds the new pool as a separate mirror group in each tier and sets data_placement: MIRROR. The existing mirror_writer background service (in src/server/bg_services/mirror_writer.js) already runs continuously — it detects MIRROR tiers and copies chunks between mirror groups automatically. So we're just configuring tiers in a way the existing worker picks up.

pkg/backingstore/backingstore.go
kajalpareek-lab: Because the replace workflow has multiple steps — especially in migration mode where the user needs to wait for replication, finalize, set manualDefaultBackingStore, and then delete the old backing store. Printing the exact oc patch and oc delete commands saves the user from looking them up. Without these, the user would need to check external docs to figure out what to do next. It's a one-time print per invocation, not noisy. If you say i will remove them.

pkg/backingstore/backingstore.go
kajalpareek-lab: Not a dedicated one today. The user can check noobaa bucket status <bucket-name> to see the mirror configuration is in place, and the mirror_writer logs progress to the NooBaa server logs.

pkg/backingstore/backingstore.go
kajalpareek-lab: Yes, it's two steps:

noobaa backingstore replace old new --migrate  adds new as a mirror alongside old. The background mirror_writer then copies existing data from old to new automatically.
noobaa backingstore replace old new (without --migrate) removes old and keeps only new.
The two-step approach makes sure data is fully replicated before we cut over. If you don't care about existing data (e.g., empty buckets), you can skip step 1 and just run the direct replace.

pkg/backingstore/backingstore.go
kajalpareek-lab: The detailed output gives copy-paste-ready commands — the oc patch command for manualDefaultBackingStore and the oc delete for the old backing store. I think that's more helpful than a generic "please delete the old backing store" message where the user has to figure out the exact commands. But if you feel its too much, I can trim it down. Let me know.

pkg/backingstore/backingstore.go
kajalpareek-lab: Fair point. Right now the RPC reply only returns tier and account counts — adding bucket names would require traversing tiering policies → buckets in the core RPC, which adds complexity. For this first version, the tier/account counts are enough for an admin to confirm the operation worked. We can enhance the reply to include affected bucket names in a follow-up.

pkg/backingstore/backingstore.go
kajalpareek-lab: The current default behavior is already "all buckets"  adding --all-buckets would just be a no-op confirmation flag that every user has to type for the same result. And --bucket-name has the shared-tier issue I mentioned above.
I think the cleanest approach is: ship the all-buckets behavior now (that's the bug fix), and if per-bucket filtering is needed later, add --bucket-name as an optional flag in a separate PR with proper handling for shared tiers and account defaults.

pkg/backingstore/backingstore.go
aayushchouhan09: Can you please share an example of how bucket status would show mirror status (screenshot may be)?

And now we can move to step 2.

pkg/backingstore/backingstore.go
aayushchouhan09: Why we are not doing it in one step?

Are we manually checking whether the mirroring is done or not. Will step 2 check this part and remove old one after mirroring is done.

pkg/backingstore/backingstore.go
aayushchouhan09: I think you can move outside that final deletion of backingstore from the if-else block as it is common for both.

pkg/backingstore/backingstore.go
kajalpareek-lab: Two steps because the user needs to decide when replication is done we can't automate that judgment. Step 1 sets up mirroring then the mirror_writer starts copying data in the background. The user checks replication progress and decides when to finalize. Step 2 (finalize) just swaps the pools  it doesn't check replication status, it trusts the user has verified it. Making it one step would risk data loss if we cut over before all data is replicated. The two-step approach is the safe path.

pkg/backingstore/backingstore.go
kajalpareek-lab: Addressed.

PR #2114, opened 2026-09-08
Allow store endpoint change using connection CLI

karthikps97:
### Describe the Problem
1. Noobaa admission webhook rejects changes to endpoint for namespace stores. Now with the new connection CLI, we should allow these changes.
2. The validations are currently only for namespace stores. Same should be applicable for backing stores.

### Explain the changes
1. Add validations for endpoint change to backingstores and allow changes to endpoint on presence of noobaa `pause` annotation.
2. Allow changes to namespace store endpoint on presence of noobaa `pause` annotation (true).
3. We are returning error if the old spec is empty

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [x] Tests added

## Summary by CodeRabbit

- **Validation**
  - Backing store endpoint changes are checked before other update validations.
  - Changes to S3-compatible and IBM COS endpoints are restricted to the connection CLI unless reconciliation is paused.
  - Equivalent endpoint formats continue to be accepted; differing endpoints are rejected.
  - Namespace store endpoint changes follow the same pause-reconciliation behavior.
  - AWS S3 endpoint changes remain permitted.
  - Invalid or incomplete endpoint update data is now rejected.

### DIFF
--- a/pkg/backingstore/validation_test.go
+++ b/pkg/backingstore/validation_test.go
@@ -1,0 +1,344 @@
+package backingstore
+
+import (
+	"fmt"
+	"reflect"
+	"testing"
+
+	nbv1 "github.com/noobaa/noobaa-operator/v5/pkg/apis/noobaa/v1alpha1"
+	"github.com/noobaa/noobaa-operator/v5/pkg/constants"
+	"github.com/noobaa/noobaa-operator/v5/pkg/validations"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	defaultEndPointURI = "https://127.0.0.1:6443"
+)
+
+func TestBackingStoreS3Compatible(t *testing.T) {
+
+	// Valid backingstore
+	defaultBS := getDefaultS3CompatibleBsStore()
+	err := validations.ValidateBackingStore(defaultBS)
+	AssertNotError(t, err, "Valid s3-compatible backingstore validation failed")
+
+	// Signature version is empty — allowed
+	defaultBS = getDefaultS3CompatibleBsStore()
+	defaultBS.Spec.S3Compatible.SignatureVersion = ""
+	err = validations.ValidateBackingStore(defaultBS)
+	AssertNotError(t, err, "Empty signature version should be allowed")
+
+	// Valid v2 signature version
+	defaultBS = getDefaultS3CompatibleBsStore()
+	defaultBS.Spec.S3Compatible.SignatureVersion = "v2"
+	err = validations.ValidateBackingStore(defaultBS)
+	AssertNotError(t, err, "Valid signature version v2 should be allowed")
+
+	// Invalid signature version
+	defaultBS = getDefaultS3CompatibleBsStore()
+	defaultBS.Spec.S3Compatible.SignatureVersion = "v5"
+	err = validations.ValidateBackingStore(defaultBS)
+	AssertError(t, err, "Invalid signature version v5 should be denied")
+
+	// Empty secret name
+	defaultBS = getDefaultS3CompatibleBsStore()
+	defaultBS.Spec.S3Compatible.Secret.Name = ""
+	err = validations.ValidateBackingStore(defaultBS)
+	AssertError(t, err, "Empty secret name should be denied")
+
+	// Empty target bucket
+	defaultBS = getDefaultS3CompatibleBsStore()
+	defaultBS.Spec.S3Compatible.TargetBucket = ""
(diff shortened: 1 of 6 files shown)


pkg/validations/namespacestore_validations.go
```diff
@@ -446,6 +447,11 @@ func ValidateNSArchiveSpec(nsStore nbv1.NamespaceStore) error {
 
 // ValidateNSEndpointChange validates the user is not trying to update the namespacestore endpoint
 func ValidateNSEndpointChange(ns nbv1.NamespaceStore, oldNs nbv1.NamespaceStore) error {
+	// skip validation if connection CLI has added the pause annotation
+	if ns.Annotations != nil && ns.Annotations[constants.PauseReconcile] == "true" {
+		return nil
```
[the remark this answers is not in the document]
karthikps97: This is acceptable since the CLI performs the actions on behalf of the user. The annotation is itself validates the claim that the update is from the CLI. We can't validate it using any other approach here.

pkg/validations/backingstore_validations.go
```diff
@@ -62,6 +63,36 @@ func ValidateBackingStore(bs nbv1.BackingStore) error {
...
+// validate changes to endpoint on backingstore
+func ValidateBSEndpointChange(bs, oldBs nbv1.BackingStore) error {
+	// skip validation if connection CLI has added the pause annotation
+	if bs.Annotations != nil && bs.Annotations[constants.PauseReconcile] == "true" {
+		return nil
+	}
+
+	switch bs.Spec.Type {
+	case nbv1.StoreTypeS3Compatible:
+		equal, err := EndpointsEquivalent(bs.Spec.S3Compatible.Endpoint, oldBs.Spec.S3Compatible.Endpoint)
```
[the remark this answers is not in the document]
karthikps97: Isn't this resolved now?

pkg/validations/backingstore_validations.go
```diff
@@ -62,6 +63,36 @@ func ValidateBackingStore(bs nbv1.BackingStore) error {
...
+		if err != nil {
+			return err
+		} else if !equal {
+			return util.ValidationError{
+				Msg: "Changing a Backingstore endpoint is supported only using the connection CLI",
+			}
+		}
+	case nbv1.StoreTypeIBMCos:
+		equal, err := EndpointsEquivalent(bs.Spec.IBMCos.Endpoint, oldBs.Spec.IBMCos.Endpoint)
```
VershaAgrawal: Nil-guard needed for oldBs.Spec.IBMCos / bs.Spec.IBMCos before calling EndpointsEquivalent.

pkg/backingstore/validation_test.go
```diff
@@ -0,0 +1,320 @@
...
+	defaultBS = nbv1.BackingStore{
+		Spec:       nbv1.BackingStoreSpec{Type: nbv1.StoreTypeIBMCos},
+		ObjectMeta: metav1.ObjectMeta{Name: "test"},
+	}
+	err = validations.ValidateBackingStore(defaultBS)
+	AssertError(t, err, "Nil IBMCos spec should be denied")
+}
+
+func TestValidateBSEndpointChange(t *testing.T) {
+
```
VershaAgrawal, line 112: Added cases for :

1. oldBs with nil type-specific spec
2. Endpoint change denied when pause annotation is absent or not "true"

pkg/validations/backingstore_validations.go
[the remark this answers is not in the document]
VershaAgrawal: @karthikps97 Add a test case for this: 
Type transition (e.g. old aws-s3 → new s3-compatible with endpoint change) — should return validation error, not panic

VershaAgrawal: ValidateNSEndpointChange already guards with nil checks — please mirror that here.

VershaAgrawal (changes requested):

pkg/validations/backingstore_validations.go
karthikps97: We would still have a scenario where a change in store type + endpoint would return `nil` eventhough there was change in endpoint.

karthikps97: This can be prevented by adding a validation to prevent change to store type. At least in the scope of this function, we can return an error if the old spec is nil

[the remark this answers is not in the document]
karthikps97: IMO type transition should be handled as a separate validation since this function is specifically for checking endpoint change for same type.

VershaAgrawal:
Add in Explain Changes: We are returning error if the old spec is empty.

VershaAgrawal (approved):
LGTM

PR #2126, opened 2026-09-18
Handle retry and improve rollback for connection CLI

karthikps97:
### Describe the Problem
The error handling strategy is to rollback the changes made to the store. But there is an issue which can cause this to fail due to stale kube object being used during the rollback. Currently there is also no retry being done on any resource version conflicts. 

### Explain the changes
1. Fix issue with rollback due to stale kube object and reference aliasing (Fetch the latest object -> make changes -> issue an update).
2. Use RetryOnConflict strategy with default backoff for persisting updates to stores.

1. 

### Testing Instructions:
1. 

- [ ] Doc added/updated
- [ ] Tests added

## Bug Fixes

- Prevented unnecessary connection updates when endpoints are unchanged.
- Improved reliability by retrying conflicting connection updates automatically.
- Enhanced rollback handling for partially completed updates.
- Improved error reporting during connection setup, completion, and rollback.
- Preserved pause-setting cleanup when another update step fails.
- Improved Kubernetes update failure handling so errors are reported without interrupting normal processing.


pkg/connection/connection.go
```diff
@@ -478,35 +494,60 @@ func setPauseAnnotation(stores []matchedStore, pause bool) (errors []error) {
...
+}
 
-		if !util.KubeUpdate(m.Store) {
-			return fmt.Errorf("failed to patch Store %q", m.name())
+func patchEndpoints(stores []matchedStore, newEndpoint string) ([]matchedStore, error) {
+	patched := []matchedStore{}
+	for i := range stores {
+		m := stores[i]
+		if err := updateWithRetry(m.Store, func(store client.Object) error {
+			mutateEndpoint(store, m.isBackingStore, newEndpoint)
```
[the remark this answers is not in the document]
karthikps97: The CLI is not meant to be run concurrently. The reconcile loop is also paused when the cli runs, so there is no risk of this happening unintentionally.

pkg/connection/connection.go
```diff
@@ -455,21 +466,26 @@ func setPauseAnnotation(stores []matchedStore, pause bool) (errors []error) {
...
+				annotations[constants.PauseReconcile] = "true"
+			} else {
+				delete(annotations, constants.PauseReconcile)
+			}
+			store.SetAnnotations(annotations)
+			return nil
+		})
 
-		if !util.KubeUpdate(store) {
+		if err != nil {
```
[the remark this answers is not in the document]
karthikps97: This is now resolved

pkg/connection/connection.go
```diff
@@ -478,35 +486,60 @@ func setPauseAnnotation(stores []matchedStore, pause bool) (errors []error) {
...
+}
+
+// updateWithRetry updates a Kubernetes object using optimistic concurrency
+// control and retries the operation when a conflict occurs.
+func updateWithRetry(obj client.Object, mutate func(client.Object) error) error {
+	if obj == nil {
+		return fmt.Errorf("function expects a non-nil object")
+	}
+
+	return retry.RetryOnConflict(retry.DefaultBackoff, func() error {
```
aayushchouhan09, line 549: Can you explain the working of this function in short, also may be you can add a short comment

pkg/connection/connection.go
```diff
@@ -562,7 +579,7 @@ func rollback(patched []matchedStore, oldEndpoint string, allStores []matchedSto
 		}
 	}
 
-	if errs := setPauseAnnotation(allStores, false); len(errs) > 0 {
+	if errs := removePauseAnnotations(allStores); len(errs) > 0 {
```
aayushchouhan09: I think we dont need this change, `removePauseAnnotations()` calling the same function `setPauseAnnotations()` internally and logging the same error.
With your changes we will print the similar error log two times, please verify this

pkg/util/util.go
```diff
@@ -586,6 +586,18 @@ func KubeDeleteAllOf(obj client.Object, opts ...client.DeleteAllOfOption) bool {
 	return deleted
 }
 
+// KubeUpdateWithError updates an object and returns the error if any
+func KubeUpdateWithError(obj client.Object) error {
```
aayushchouhan09, line 590: `kubeUpdate()` and `KubeUpdateWithError()` both of them having the same duplicate code except the return, can we reduce the duplication if possible

aayushchouhan09 (commented):
Please add more explaination to the PR description. Thanks

pkg/util/util.go
karthikps97: Unfortunately I need the error as it is, and `kubeUpdate` doesn't return. We could add an additional error return type but it's referenced a lot and we would need to modify all that code.

pkg/connection/connection.go
karthikps97: Removed. Thanks

pkg/connection/connection.go
karthikps97: Yes, sure. Added a line

pkg/connection/connection.go
```diff
@@ -122,14 +125,16 @@ func RunUpdate(cmd *cobra.Command, args []string) {
 
 	// Set pause annotation on all matching stores
 	if errs := setPauseAnnotation(matched, true); len(errs) > 0 {
-		removePauseAnnotations(matched)
-		log.Fatalf("failed to set pause annotation: %s", errs)
+		if errs := setPauseAnnotation(matched, false); len(errs) > 0 {
+			log.Errorf("failed to remove pause annotations during rollback: %s", errs)
+		}
```
aayushchouhan09, line 133: why we did these changes?

aayushchouhan09 (approved):
LGTM

aayushchouhan09 (approved):

liranmauda (changes requested):
Blocking for a few

liranmauda (approved):
Unbloking

PR #2127, opened 2026-09-18
Map alternate key names of secret

karthikps97:
### Describe the Problem
Ceph RGW key names are AccessKey / SecretKey rather than the conventional formats used by AWS, IBM COS. We need to include these as well while fetching the values from the secret.

### Explain the changes
1. Map the alternative key name's also while fetching the values from the secret.

1. 

### Testing Instructions:
1. Create backingstore using Ceph RGW key format (AccessKey / SecretKey). 
2. Now run the CLI for updating the endpoint to the new valuee. It should work as expected.

- [ ] Doc added/updated
- [ ] Tests added

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection setup by recognizing alternate secret-key names when the expected credentials are unavailable.
  * Preserved existing credentials when the expected identity or secret key is present.


VershaAgrawal (approved):
LGTM

